PR #9, opened 2025-08-08
Updates to documentation

catblade:
- Create a hack directory and move examples and kind script to in there
- Update Makefile with some minor bug fixes and add ability to make the kind images
- Update Readme instructions
- Remove no-longer needed file that is now encapsulated within install.yaml

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,12 @@ The driver is deployed as a DaemonSet which contains two core components:
 
 ### Installation
 
-- Create a kind cluster
+- If needed, create a kind cluster. We have one in the repo, if needed, that
+  can be deplayed as follows:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,8 +70,21 @@ build-image: ## build image
...
+		--load
 
 push-image: build-image ## build and push image
 	docker push ${IMAGE}
 	docker push ${IMAGE_LATEST}
+
+kind-cluster:  ## create kind cluster
+	kind create cluster --name ${CLUSTER_NAME} --config hack/kind.yaml
+
+install-cpu-dra: build-image ## install on cluster
```
pravk03, line 83: Image path needs an update

PR #17, opened 2025-08-28
feat: Switch to resource API v1 

pravk03:
- Switch to resource API v1 
- Updating the kind cluster to use `v0.34.0`.
- Enable the `DRAConsumableCapacity` feature gate


johnbelamaric (approved):
/approve
/lgtm

PR #21, opened 2025-09-10
OWNERS: propose ffromani as approver

ffromani:
per offline conversation, proposing myself.


pravk03:
+1. Thanks @ffromani. Your deep expertise in Kubelet resource management gives us a lot of confidence in the changes we make.

Issue #15, opened 2025-08-19
Release guaranteed CPUs as soon as possible back into the shared pool


ffromani:
Following up the conversation on #10, it seems from the code and its TODOs that once a container with exclusive CPUs is removed, the exclusively assigned CPUs are not released back in the shared pool. This will be fixed, but only when the next container is created which can be create undesirable and unnecessary time windows on which the shared pool has less CPU than possible.
xref: https://github.com/kubernetes-sigs/dra-driver-cpu/blob/12c0d1aeb6eb60f147847f05687c9f009f88596a/pkg/driver/nri_hooks.go#L166
xref: https://github.com/kubernetes-sigs/dra-driver-cpu/blob/12c0d1aeb6eb60f147847f05687c9f009f88596a/pkg/driver/nri_hooks.go#L190

ffromani:
/assign

with the caveat that I have partial availability till Sep 1

PR #22, opened 2025-09-11
e2e: test for exclusive CPU assignment

ffromani:
add initial test to automatically validate the exclusive cpu assignment.

Notes for reviewers:
- we intentionally use a bogus image pullSpec to be sure we either use the freshly-built image from current sources or fail loudly
- we fetch the cgroup settings from the test pods using logs, so we let the containers be continuously restarted to refresh the information; see the commit messages for details.

PR #26, opened 2025-09-24
WIP: feat: release guaranteed CPUs as soon as possible back into the shared pool 

ffromani:
WIP

Fixes: #15
Requires: #22


ffromani:
the direction is still promising but I need to do quite some large updates after #20 merged. Will get to it soon.

pkg/driver/pod_store.go
```diff
@@ -120,7 +117,7 @@ func (s *PodConfigStore) GetContainersWithSharedCPUs() []types.UID {
 
 // HasExclusiveCPUAllocation returns true if the container has associated resource claims.
 func (cs *ContainerState) HasExclusiveCPUAllocation() bool {
-	return len(cs.resourceClaimUIDs) > 0
+	return cs != nil && len(cs.resourceClaimUIDs) > 0
```
pravk03: Oh i see, we do a nil check here. I slightly prefer an explicit existence check before calling this function.

pravk03 (commented):
LGTM. Thanks a lot for making this change. Definitely a much cleaner solution than what we had. Added a couple of nits.

pkg/driver/pod_store.go
ffromani: makes sense, I'll change accordingly.

PR #27, opened 2025-09-25
yaml: drop podresources socket mounts

ffromani:
we don't use them and it seems there are no plans to use them, so let's simplify the YAMLs and reduce the plugin surface.


pravk03:
I think this is leftover of one of the approach I was exploring - https://github.com/pravk03/simple-cpu-dra-driver/commit/16d7a260c1581cec8c027a2140fff22b7a3f9524. 

Thanks for removing. 

/lgtm

PR #28, opened 2025-09-26
OWNERS: propose pravk03 as approver

ffromani:
@pravk03 is the main author so I was actually surprised when I discovered he's not approver, let's fix that.


ffromani:
/cc @johnbelamaric 

FYI @pravk03

pravk03:
Thanks @ffromani, @johnbelamaric. Happy to an approver :).

PR #29, opened 2025-10-01
ci: use golangci-lint pre-built binaries

ffromani:
Installing `golangci-lint` from sources is not recommended: see https://golangci-lint.run/docs/welcome/install/#install-from-sources

So, we switch to the recommended way to download pre-built binaries, which also allows us to sidestep https://github.com/golangci/golangci-lint/issues/6017 which we can still hit if we bump e.g. to 2.4.0.

In the CI config, we now reuse the same Makefile targets developers are supposed to use, to make the flow consistent and minimize surprises (e.g. something works locally but not in CI or the other way around).

This PR also adds a way to re-use system tools if existing, to save bandwidth in CI.


pravk03:
I did not realize the bot would merge this without giving a chance to  look at the comments :)

PR #31, opened 2025-10-14
bugfix: cdi: write the CDI spec atomically

ffromani:
The recommended way in unix environments to atomically write a file is to write a temporary file and then use rename(2) which is atomic.
This avoids entirely any possibility of hving a half-written file.

In the future, we may want to look into
```
tags.cncf.io/container-device-interface/pkg/cdi
tags.cncf.io/container-device-interface/pkg/parser
```

to replace out cdi handling code entirely. For the short term, the low hanging fruit is to fix the bug and carry on while the other features mature. We can pivot to the CDI packages after we hit the first milestone

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.24.0
 require (
 	github.com/containerd/nri v0.9.0
 	github.com/go-logr/logr v1.4.2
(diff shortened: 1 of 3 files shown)


pravk03:
/lgtm 

Thanks for fixing this.

Issue #4, opened 2025-08-08
Set appropriate priority, resource request, QOS class for the CPU DRA Driver DaemonSet


pravk03:
The DaemonSet pod (install.yaml) runs as a burstable pod. We need to determine the appropriate CPU and memory resource requirements for the pod.

/kind feature

ffromani:
I see, thanks @pravk03 for elaborating. I think I'll soon post a PR to suggest the changes to the DaemonSet according to the sizing guidelines I'm aware of. It could be a good idea to add some notes in the README, I'll take care in the PR.

PR #42, opened 2026-01-19
daemonset: cleanup and set priority

ffromani:
set pod priority for the dracpu daemonset. This way we decouple the resource requests, which we haven't figured out yet, from the actual pod priority for eviction and preemption.
For more details: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#interactions-of-pod-priority-and-qos

Note 

The collateral question is if we should bump resource requirements and if we should opt in the guaranteed QOS class.
We don't know yet the resource required in normal operations, so we don't change unless we have more data.

Opting in the Guaranteed QOS class would mean set upper hard limit to the resources. This goes against the scalability requirements (if nothing else, bigger clusters means more kubernetes objects to observe) and is even harder to do without data to drive our changes.

For all these reasons, we abstain for now from changing resource requirements till we have real world data.

Fixes: #4

### DIFF
--- a/install.yaml
+++ b/install.yaml
@@ -110,6 +110,10 @@ spec:
           name: etc
         securityContext:
           privileged: true
+        resources:
+          requests:
+            cpu: "100m"
+            memory: "50Mi"
         command:
         - /bin/sh
         - -c


install.yaml
pravk03: Do you think we should copy the resource request from main container to init container. IIUC, zero request would make it more susceptible for eviction/OOM kill before main container runs ?

pravk03:
/lgtm 

Thanks !

PR #55, opened 2026-02-06
Remove NUMA node boundary for ResourceSlice splitting

### LINKED ISSUE #5: Handle ResourceSlice 128-device limit per ResourceSlice
The current implementation has a workaround to create one ResourceSlice per NUMA node if the machine has more than 128 CPUs. This is a obvious limitation and needs a better solution.

/kind feature

pravk03:
Removes the self-imposed restriction of confining each `ResourceSlice` to a single NUMA node when we exceed the 128 device limit (set [here](https://github.com/kubernetes/kubernetes/blob/8e6d788887034b799f6c2a86991a68a080bb0576/pkg/apis/resource/types.go#L245)). Slices will now be created by simply chunking the devices, potentially spanning NUMA nodes.

Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/5

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -19,6 +19,7 @@ package driver
 import (
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 
 	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/cpuinfo"
@@ -182,7 +183,7 @@ func (cp *CPUDriver) createCPUDeviceSlices() [][]resourceapi.Device {
 	})
 
 	devId := 0
-	devicesByNuma := make(map[int64][]resourceapi.Device)
+	var allDevices []resourceapi.Device
 	for _, group := range coreGroups {
 		for _, cpu := range group {
 			numaNode := int64(cpu.NUMANodeID)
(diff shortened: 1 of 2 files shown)


pkg/driver/dra_hooks.go
```diff
@@ -208,42 +208,25 @@ func (cp *CPUDriver) createCPUDeviceSlices() [][]resourceapi.Device {
...
-		if len(numaDevices) > maxDevicesPerResourceSlice {
-			klog.Errorf("number of devices for NUMA node %d (%d) exceeds the slice limit of %d", id, len(numaDevices), maxDevicesPerResourceSlice)
-			return nil
-		}
-		if len(numaDevices) > 0 {
-			allSlices = append(allSlices, numaDevices)
+	if len(allDevices) <= maxDevicesPerResourceSlice {
+		allSlices = append(allSlices, allDevices)
+	} else {
+		// Chunk devices into slices of at most maxDevicesPerResourceSlice
```
ffromani: perhaps https://pkg.go.dev/slices#Chunk can help here?

ffromani (commented):
LGTM with one possible improvement, please see the inline comment

ffromani (commented):
/lgtm

thanks!

PR #65, opened 2026-02-19
dra: attributes: compatibility with dra-driver-sriov

ffromani:
Add compatibility with dra-driver-sriov to enable NUMA alignment. Reviewing https://github.com/k8snetworkplumbingwg/dra-driver-sriov/blob/main/pkg/consts/consts.go this seems the only attribute which makes sense to expose.

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -27,6 +27,7 @@ import (
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/dynamic-resource-allocation/deviceattribute"
 	"k8s.io/dynamic-resource-allocation/kubeletplugin"
 	"k8s.io/dynamic-resource-allocation/resourceslice"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


ffromani:
Exploring sysfs a bit, I think there could be a good way to group cpu cores by pcieRoot affinity. This will call for a new groupBy mode, and this will allow us to expose the standard pcieRoot attribute avoiding the complexity entirely. I'll explore this option next week and hopefully push a PR. 
I don't think we *need* to include this in 0.1.0, worst case we can cut a 0.1.1 with only this addition. Let's see how it goes.

pkg/driver/dra_hooks.go
```diff
@@ -121,6 +122,8 @@ func (cp *CPUDriver) createGroupedCPUDeviceSlices() [][]resourceapi.Device {
 					"dra.cpu/smtEnabled": {BoolValue: &smtEnabled},
 					// TODO(pravk03): Remove. Hack to align with NIC (DRANet). We need some standard attribute to align other resources with CPU.
 					"dra.net/numaNode": {IntValue: &numaID},
+					// dra-driver-sriov (https://github.com/k8snetworkplumbingwg/dra-driver-sriov) compatibility
+					deviceattribute.StandardDeviceAttributePrefix + "numaNode": {IntValue: &numaID},
```
pohly: A big 👎 on this. A driver should never use the reserved namespace for attributes that haven't been defined by Kubernetes.

At this point, `numaNode` is something that has to live inside a driver's own namespace.

ffromani:
thanks @pohly , the feedback is clear. Compatibility needs to be achieved using a different mean, and (ab)using the kubernetes namespace for extra attributes is a bad practice we should try to rectify rather than follow. I think there's a promising avenue to reach compatibility for the CPU (and memory) using the pcieRoot attribute, we will pursue in a upcoming PR.

PR #66, opened 2026-02-19
dra: codeorg: simplify device attributes

ffromani:
remove temporary variables and add a function whose purpose is to add attributes to enable compatibility with other relevant DRA drivers (dranet for now).

We will keep this function until all the attributes we need get standardized in kubernetes core - or any other mechanism the community will pick.

Unfortunately, at the moment the only way to be compatible with nonstandard attribute is crosscheck from both sides (driver <=> driver) creating a dense compatibility matrix.

This is awkward and scales poorly but it's all we have atm.

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/cpuinfo"
 	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/cpumanager"
+	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/device"
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


pkg/driver/dra_hooks.go
```diff
@@ -186,28 +187,23 @@ func (cp *CPUDriver) createCPUDeviceSlices() [][]resourceapi.Device {
 	var allDevices []resourceapi.Device
 	for _, group := range coreGroups {
 		for _, cpu := range group {
-			numaNode := int64(cpu.NUMANodeID)
-			cacheL3ID := int64(cpu.UncoreCacheID)
-			socketID := int64(cpu.SocketID)
-			coreID := int64(cpu.CoreID)
-			cpuID := int64(cpu.CpuID)
-			coreType := cpu.CoreType.String()
+			deviceAttrs := map[resourceapi.QualifiedName]resourceapi.DeviceAttribute{
```
pravk03: nit: maybe we can also add dra.cpu/smtEnabled here ?. Not sure why we missed earlier.

PR #75, opened 2026-03-09
Default to stripping DWARF info and symbol table from binary

fmuyassarov:
DWARF debug info and symbol tables are primarily needed during debugging (e.g. with Delve). Since DWARF was originally designed for C and its utility in Go is minimal, there is little reason to include it in the builds by default.

The diff in terms of image size (before and after)
```
docker images --format "{{.Repository}} {{.Size}}" | grep dra
dra-cpu 137MB
dra-cpu 104MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@
 FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
 ARG TARGETARCH
 ARG GOARCH=${TARGETARCH} CGO_ENABLED=0
+ARG LDFLAGS=-s -w
 
 # cache go modules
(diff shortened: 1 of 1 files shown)


ffromani (commented):
I tend to agree. Our main release artifacts should be as lean as possible. I'd rather add a variant with debug info if needed or just instruct users how to build them.

@pravk03 any concerns or objections?

AutuSnow (commented):
In my opinion, - a and - extldflags=- static are redundant parameters that can be simplified

pravk03:
Since version 0.1.0 is out, this is ready to merge

/lgtm 
/approve

PR #83, opened 2026-03-11
feat: add Helm chart for driver deployment

### LINKED ISSUE #72: Proposal: Helm chart for dra-driver-cpu
The current `install.yaml` is a static manifest that requires manual patching for any customization. A Helm chart would expose driver configuration as proper `values.yaml `parameters, enable environment specific overrides without forking and provide a versioned installation option for users already using Helm.

I guess the bare minimum steps would involve:

1. writing the Helm chart
2. CI integration for linting, templating, packaging, and pushing; 0.0.0-main from main branch, versioned charts from releases
3. hosting the charts, I believe we could use something like `ghcr.io/kubernetes-sigs/dra-driver-cpu/charts/dra-driver-cpu` for the artifacts
4. updating the documentation

If this sounds reasonable and beneficial, I would be happy to work on it.

fmuyassarov:
Add a Helm chart for driver installation. This PR adds:
- Helm chart for driver installation
- Documentation to describe installation and available values
- Linter CI for the charts & schema validation

Follow-up (_TODO_)
- chart packaging and publishing to `ghcr.io/kubernetes-sigs/dra-driver-cpu/charts/dra-driver-cpu`
- versioned releases from tags, `0.0.0-main` from main branch

Note: currently all the templates (DeamonSet, ServiceAccount, etc are based on the what is available in install.yaml).

Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/72

### DIFF
--- a/.github/workflows/helm-e2e.yaml
+++ b/.github/workflows/helm-e2e.yaml
@@ -1,0 +1,91 @@
+name: helm e2e
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'deployment/helm/**'
+      - '.github/workflows/helm-e2e.yaml'
+
+env:
+  TAG: ci-${{ github.sha }}
+  # Matches REGISTRY_CI/IMAGE_NAME in the Makefile
+  IMAGE: dev.kind.local/ci/dra-driver-cpu
+
+jobs:
+  helm-e2e:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: '1.25'
+
+      - name: Install kind
+        run: |
+          curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.27.0/kind-linux-amd64
+          chmod +x ./kind
+          sudo mv ./kind /usr/local/bin/kind
+
+      - name: Install kubectl
+        run: |
+          if ! command -v kubectl &>/dev/null; then
+            curl -LO "https://dl.k8s.io/release/$(curl -Ls https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+            chmod +x kubectl
+            sudo mv kubectl /usr/local/bin/kubectl
+          fi
+
+      - name: Build the image
+        run: make build-image
+
+      - name: Create kind cluster
+        run: |
+          kind delete cluster --name dra-driver-cpu || true
+          kind create cluster --name dra-driver-cpu --config hack/ci/kind-ci.yaml
+          kind get kubeconfig --name dra-driver-cpu > /tmp/kubeconfig-dra-driver-cpu
+          kubectl --kubeconfig /tmp/kubeconfig-dra-driver-cpu \
+            label node dra-driver-cpu-worker node-role.kubernetes.io/worker=''
+
+      - name: Load driver image into kind
+        run: |
+          kind load docker-image --name dra-driver-cpu \
+            "${IMAGE}:${TAG}"
+
+      - name: Install driver via Helm
+        run: |
+          helm install dra-driver-cpu deployment/helm/dra-driver-cpu \
+            --kubeconfig /tmp/kubeconfig-dra-driver-cpu \
+            --namespace kube-system \
+            --set fullnameOverride=dracpu \
+            --set podLabels.app=dracpu \
+            --set image.repository="${IMAGE}" \
+            --set image.tag="${TAG}" \
+            --set image.pullPolicy=Never
+
+      - name: Wait for resourceSlices
+        run: KUBECONFIG=/tmp/kubeconfig-dra-driver-cpu hack/ci/wait-resourcelices.sh
+
+      - name: Check daemonSet pods are Running & Ready
+        run: |
+          kubectl --kubeconfig /tmp/kubeconfig-dra-driver-cpu \
+            rollout status daemonset/dracpu -n kube-system --timeout=120s
+
+      - name: Final check
+        if: failure()
+        continue-on-error: true
+        run: |
+          kubectl --kubeconfig /tmp/kubeconfig-dra-driver-cpu get pods -A -o wide || true
+          kubectl --kubeconfig /tmp/kubeconfig-dra-driver-cpu describe daemonset dracpu -n kube-system || true
+          kubectl --kubeconfig /tmp/kubeconfig-dra-driver-cpu logs -n kube-system -l app=dracpu --tail=100 || true
+          kubectl --kubeconfig /tmp/kubeconfig-dra-driver-cpu get resourceslice -o yaml || true
+
+      - name: Delete kind cluster
+        if: always()
+        run: kind delete cluster --name dra-driver-cpu
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -1,0 +1,18 @@
+name: Helm Lint
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'deployment/helm/**'
+
+jobs:
+  helm-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+
+      - name: Run Helm lint
+        run: make helm-lint
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -25,3 +25,6 @@ jobs:
 
       - name: Check helm-docs is up to date
         run: make helm-docs-check
+
+      - name: Check values.schema.json is up to date
+        run: make helm-schema-check
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -4,6 +4,7 @@ on:
     branches: [ main ]
     paths:
       - 'deployment/helm/**'
+      - '.github/workflows/helm-lint.yaml'
 
 jobs:
   helm-lint:
@@ -11,8 +12,16 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: '1.25'
+
       - name: Install Helm
         uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
 
       - name: Run Helm lint
         run: make helm-lint
+
+      - name: Check helm-docs is up to date
+        run: make helm-docs-check
--- a/deployment/helm/dra-driver-cpu/Chart.yaml
+++ b/deployment/helm/dra-driver-cpu/Chart.yaml
@@ -1,0 +1,44 @@
+# Copyright The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+apiVersion: v2
+name: dra-driver-cpu
+description: Kubernetes DRA driver for managing CPU resources with topology-aware allocation,
+  exclusive CPU assignment, and shared CPU pool management via the Dynamic Resource Allocation
+  framework.
+type: application
+version: 0.1.0
+appVersion: latest
+kubeVersion: ">= 1.34.0-0"
+sources:
+  - https://github.com/kubernetes-sigs/dra-driver-cpu
+home: https://github.com/kubernetes-sigs/dra-driver-cpu
+keywords:
+  - dra
+  - dra-driver
+  - cpu
+  - kubernetes
+  - resource-management
+  - numa
+maintainers:
+  - name: johnbelamaric
+    url: https://github.com/johnbelamaric
+  - name: pohly
+    url: https://github.com/pohly
+  - name: klueska
+    url: https://github.com/klueska
+  - name: ffromani
+    url: https://github.com/ffromani
+  - name: pravk03
+    url: https://github.com/pravk03
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -1,0 +1,60 @@
+# dra-driver-cpu Helm Chart
+
+Deploys the [dra-driver-cpu](https://github.com/kubernetes-sigs/dra-driver-cpu) DaemonSet — a Kubernetes Dynamic Resource Allocation (DRA) driver for managing CPU resources.
+
+## Installation
+
+From a local checkout:
+
+```bash
+helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system
+```
+
+To override values at install time:
+
+```bash
+helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system \
+  --set args.cpuDeviceMode=individual \
+  --set args.reservedCPUs="0-1"
+```
+
+## Configuration
+
+The following table lists the configurable parameters and their default values:
+
+| Parameter | Description | Default |
+|---|---|---|
+| `nameOverride` | Override the chart name | `""` |
+| `fullnameOverride` | Override the full release name | `""` |
+| `image.repository` | Container image repository | `us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu` |
+| `image.tag` | Image tag (falls back to `Chart.AppVersion` if empty) | `latest` |
+| `image.pullPolicy` | Image pull policy | `IfNotPresent` |
+| `imagePullSecrets` | List of image pull secrets | `[]` |
+| `rbac.create` | Create RBAC resources (ClusterRole and ClusterRoleBinding) | `true` |
+| `serviceAccount.annotations` | Annotations to add to the ServiceAccount | `{}` |
+| `podAnnotations` | Annotations to add to pods | `{}` |
+| `podLabels` | Extra labels to add to pods | `{}` |
+| `resources.requests.cpu` | CPU resource request | `100m` |
+| `resources.requests.memory` | Memory resource request | `50Mi` |
+| `resources.limits` | Resource limits (unset by default) | `{}` |
+| `tolerations` | Node tolerations | all NoSchedule taints tolerated |
+| `args.logLevel` | Log verbosity (`--v`) | `4` |
+| `args.cpuDeviceMode` | CPU exposure mode: `grouped` or `individual` | `grouped` |
+| `args.groupBy` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode` or `socket` | `numanode` |
+| `args.reservedCPUs` | CPUs reserved for system/kubelet (e.g. `0-1`). Omitted if empty. | `""` |
+| `args.hostnameOverride` | Override the node name used by the driver. Omitted if empty. | `""` |
+| `healthzPath` | Path for liveness and readiness probes | `/healthz` |
+| `healthzPort` | Port the HTTP server binds to; used for the container port and probes | `8080` |
+
+Parameters can be set at install time using `--set` or a custom values file:
+
+```bash
+helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system --set args.logLevel=4
+helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system -f my-values.yaml
+```
+
+## Uninstallation
+
+```bash
+helm uninstall dra-driver-cpu -n kube-system
+```
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -1,6 +1,6 @@
 # dra-driver-cpu Helm Chart
 
-Deploys the [dra-driver-cpu](https://github.com/kubernetes-sigs/dra-driver-cpu) DaemonSet — a Kubernetes Dynamic Resource Allocation (DRA) driver for managing CPU resources.
+Kubernetes DRA driver for managing CPU resources with topology-aware allocation, exclusive CPU assignment, and shared CPU pool management via the Dynamic Resource Allocation framework.
 
 ## Installation
 
@@ -18,40 +18,38 @@ helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system \
   --set args.reservedCPUs="0-1"
 ```
 
-## Configuration
-
-The following table lists the configurable parameters and their default values:
-
-| Parameter | Description | Default |
-|---|---|---|
-| `nameOverride` | Override the chart name | `""` |
-| `fullnameOverride` | Override the full release name | `""` |
-| `image.repository` | Container image repository | `us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu` |
-| `image.tag` | Image tag (falls back to `Chart.AppVersion` if empty) | `latest` |
-| `image.pullPolicy` | Image pull policy | `IfNotPresent` |
-| `imagePullSecrets` | List of image pull secrets | `[]` |
-| `rbac.create` | Create RBAC resources (ClusterRole and ClusterRoleBinding) | `true` |
-| `serviceAccount.annotations` | Annotations to add to the ServiceAccount | `{}` |
-| `podAnnotations` | Annotations to add to pods | `{}` |
-| `podLabels` | Extra labels to add to pods | `{}` |
-| `resources.requests.cpu` | CPU resource request | `100m` |
-| `resources.requests.memory` | Memory resource request | `50Mi` |
-| `resources.limits` | Resource limits (unset by default) | `{}` |
-| `tolerations` | Node tolerations | all NoSchedule taints tolerated |
-| `args.logLevel` | Log verbosity (`--v`) | `4` |
-| `args.cpuDeviceMode` | CPU exposure mode: `grouped` or `individual` | `grouped` |
-| `args.groupBy` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode` or `socket` | `numanode` |
-| `args.reservedCPUs` | CPUs reserved for system/kubelet (e.g. `0-1`). Omitted if empty. | `""` |
-| `args.hostnameOverride` | Override the node name used by the driver. Omitted if empty. | `""` |
-| `healthzPath` | Path for liveness and readiness probes | `/healthz` |
-| `healthzPort` | Port the HTTP server binds to; used for the container port and probes | `8080` |
-
 Parameters can be set at install time using `--set` or a custom values file:
 
 ```bash
 helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system --set args.logLevel=4
 helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system -f my-values.yaml
 ```
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| args.cpuDeviceMode | string | `"grouped"` | CPU exposure mode: `grouped` (expose NUMA nodes or sockets as devices) or `individual` (expose each CPU as a device) |
+| args.groupBy | string | `"numanode"` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode` or `socket` |
+| args.hostnameOverride | string | `""` | Override the node name the driver registers under; omitted when empty |
+| args.logLevel | int | `4` | Log verbosity level passed as `--v` |
+| args.reservedCPUs | string | `""` | CPUs reserved for the OS and kubelet, excluded from DRA management (e.g. `"0-1"`); omitted when empty |
+| fullnameOverride | string | `""` | Override the full release name |
+| healthzPath | string | `"/healthz"` | Path for liveness and readiness probes |
+| healthzPort | int | `8080` | Port the HTTP server binds to; used for the container port and probes |
+| image.pullPolicy | string | `"IfNotPresent"` | Image pull policy |
+| image.repository | string | `"us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu"` | Container image repository |
+| image.tag | string | `""` | Image tag; defaults to `.Chart.AppVersion` when empty, which is set to the release tag at package time |
+| imagePullSecrets | list | `[]` | List of image pull secrets |
+| nameOverride | string | `""` | Override the chart name |
+| podAnnotations | object | `{}` | Annotations to add to pods |
+| podLabels | object | `{}` | Extra labels to add to pods |
+| rbac.create | bool | `true` | Create RBAC resources (ClusterRole and ClusterRoleBinding) |
+| resources.limits | object | `{}` | Resource limits (unset by default) |
+| resources.requests.cpu | string | `"100m"` | CPU resource request |
+| resources.requests.memory | string | `"50Mi"` | Memory resource request |
+| serviceAccount.annotations | object | `{}` | Annotations to add to the ServiceAccount |
+| tolerations | list | `[{"effect":"NoSchedule","operator":"Exists"}]` | Node tolerations; defaults to tolerating all NoSchedule taints |
 
 ## Uninstallation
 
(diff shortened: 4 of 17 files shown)


fmuyassarov:
> @fmuyassarov Can you add the configurations for livenessProbe and readinessProbe

@AutuSnow added https://github.com/kubernetes-sigs/dra-driver-cpu/pull/84 for the install.yaml and soon will add here as well.

ffromani:
will review again shortly, thanks for the patience

fmuyassarov:
Thanks. But I would ask don't do it yet, because I'm about to add few more improvements in an hour or two. Will ping you once ready.

fmuyassarov:
> Thanks. But I would ask don't do it yet, because I'm about to add few more improvements in an hour or two. Will ping you once ready.

@ffromani 
This should be ready now. Added few more changes I had in mind.

deployment/helm/dra-driver-cpu/Chart.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - name: johnbelamaric
+    url: https://github.com/johnbelamaric
+  - name: pohly
+    url: https://github.com/pohly
+  - name: klueska
+    url: https://github.com/klueska
+  - name: ffromani
+    url: https://github.com/ffromani
+  - name: pravk03
+    url: https://github.com/pravk03
```
fmuyassarov, line 44: I'm sure we need to update this list, but perhaps @pravk03 @ffromani can confirm. This was copy paste of the current OWNERS file.

deployment/helm/dra-driver-cpu/templates/deviceclass.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2026 The Kubernetes Authors.
```
pravk03: nit:  "Copyright The Kubernetes Authors."

same in other files.

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+nameOverride: ""
+fullnameOverride: ""
+
+image:
+  repository: us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu
+  tag: latest
```
pravk03: Should this be a release version ?

deployment/helm/dra-driver-cpu/templates/clusterrole.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - update
+      - delete
+  - apiGroups:
+      - resource.k8s.io
+    resources:
+      - resourceclaims
+      - deviceclasses
+    verbs:
+      - get
+  - apiGroups:
```
pravk03, line 46: We should also introduce the new rules introduced in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/105 here.

pravk03:
Thanks for looking into this! I've left a few minor comments inline. I have very limited experience with Helm, so I will lean on @ffromani and @AutuSnow who are already reviewing this PR for a more thorough review and approval.

deployment/helm/dra-driver-cpu/templates/clusterrole.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - resource.k8s.io
+    resources:
+      - resourceclaims/status
+    verbs:
+      - patch
+      - update
```
AutuSnow, line 52: issing the resourceclaims/driver rule that exists in manifests/base/clusterrole-dracpu.part.yaml. Without it the driver cannot emit associated-node:patch|update on resourceNames: [dra.cpu], which is required for DRA allocation flow on 1.34+. Please add

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+  requests:
+    cpu: "100m"
+    memory: "50Mi"
+  limits: {}
+
+tolerations:
+  - key: "node-role.kubernetes.io/control-plane"
+    operator: "Exists"
+    effect: "NoSchedule"
+
```
AutuSnow, line 61: install.yaml tolerates all NoSchedule taints (operator: Exists, effect: NoSchedule). This default only tolerates node-role.kubernetes.io/control-plane, so the DaemonSet silently stops scheduling on worker nodes that carry other NoSchedule taints. Either keep parity with install.yaml or call out the intentional behavior change in the README migration section.

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
```diff
@@ -0,0 +1,58 @@
...
+app.kubernetes.io/name: {{ include "dra-driver-cpu.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Namespace
+*/}}
+{{- define "dra-driver-cpu.namespace" -}}
+{{- default .Release.Namespace .Values.namespace }}
+{{- end }}
```
AutuSnow: this helper reads .Values.namespace, but (a) values.schema.json has root additionalProperties: false with no namespace key, so --set namespace=foo will fail schema validation; (b) no template actually calls this helper — daemonset/SA hardcode .Release.Namespace. Either drop the helper or wire it through all templates + add namespace to the schema.

deployment/helm/dra-driver-cpu/templates/deviceclass.yaml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+---
+apiVersion: resource.k8s.io/v1
+kind: DeviceClass
+metadata:
+  name: dra.cpu
+spec:
+  selectors:
+    - cel:
+        expression: device.driver == "dra.cpu"
```
AutuSnow, line 22: DeviceClass/dra.cpu is cluster-scoped and shared across any installation. helm uninstall will delete it, breaking other deployments or existing ResourceClaims that reference it.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -0,0 +1,172 @@
...
+  name: {{ include "dra-driver-cpu.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dra-driver-cpu.labels" . | nindent 4 }}
+    tier: node
+    k8s-app: {{ include "dra-driver-cpu.name" . }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "dra-driver-cpu.selectorLabels" . | nindent 6 }}
```
AutuSnow, line 27: install.yaml selector is app: dracpu; this chart uses app.kubernetes.io/{name,instance}. DaemonSet selectors are immutable, so users cannot migrate from install.yaml → Helm in place. Either keep the  legacy app: dracpu selector label for compat, or add an explicit migration note (delete+reinstall) to README.

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+args:
+  logLevel: 4
+  cpuDeviceMode: "grouped"
+  groupBy: "numanode"  # numanode or socket; only used when cpuDeviceMode=grouped
+  reservedCPUs: ""  # e.g. "0-1"
+  hostnameOverride: ""
+
+metricsPath: /healthz
+metricsPort: 8080
```
AutuSnow: Endpoint service/healthz (active/ready status), not Prometheus metrics. Naming it measurement * is misleading. Suggest using healthzPath/healthzPort, as we need to support metric endpoints in version 0.2

deployment/helm/dra-driver-cpu/Chart.yaml
```diff
@@ -0,0 +1,44 @@
...
+---
+apiVersion: v2
+name: dra-driver-cpu
+description: Kubernetes DRA driver for managing CPU resources with topology-aware allocation,
+  exclusive CPU assignment, and shared CPU pool management via the Dynamic Resource Allocation
+  framework.
+type: application
+version: 0.1.0
+appVersion: main
+kubeVersion: ">= 1.34.0"
```
AutuSnow: kubeVersion: ">= 1.34.0" excludes pre-release tags. Prefer ">=1.34.0-0" to match Helm semver convention.

.github/workflows/helm-lint.yaml
```diff
@@ -0,0 +1,18 @@
...
+  helm-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+
+      - name: Run Helm lint
+        run: make helm-lint
```
AutuSnow, line 24: helm lint --strict won't catch rendered-manifest issues (missing required fields, CRD type mismatches). Suggest adding a helm template | kubeconform -strict -summary step.

AutuSnow (commented):
Thank you for your patience and contribution. I left a comment

deployment/helm/dra-driver-cpu/templates/clusterrole.yaml
fmuyassarov: good catch, it was added while this PR was waiting for reviews. But we should add here it too anyways.

.github/workflows/helm-lint.yaml
fmuyassarov: It is true that Helm linter is for templating errors but it doesn't validate rendered manifests against API schemas. Kubeconform might do that but I would like to avoid extra complexity and dependency on yet another binary. I mean we don't even have CRDs here yet and we are already protecting user provided errors anyways with the json scheme.

deployment/helm/dra-driver-cpu/templates/clusterrole.yaml
fmuyassarov: Yup, addressed.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
fmuyassarov: I may be missing something, but I’m not entirely sure why users would need to migrate from install.yaml to Helm. From our perspective, the installation method shouldn’t impact ongoing usage, and we are not expecting/requiring existing users to migrate and I don't see value in doing so.

Those who originally installed the driver using install.yaml can continue running it as is without any changes. For new installations users can choose between install.yaml and Helm, though Helm would generally be the preferred option. Could you help to clarify your question?

deployment/helm/dra-driver-cpu/templates/deviceclass.yaml
fmuyassarov: It it true that uninstalling the chart will delete shared DeviceClass but how is it different from install.yaml?  I mean that it's not a Helm specific problem but rather an inherent characteristic of how this driver works. This is the same behavior as `kubectl delete -f install.yaml`.

deployment/helm/dra-driver-cpu/values.yaml
fmuyassarov: ah true, updated to match the install.yaml

deployment/helm/dra-driver-cpu/values.yaml
fmuyassarov: Fixed.

deployment/helm/dra-driver-cpu/Chart.yaml
fmuyassarov: Updated

deployment/helm/dra-driver-cpu/values.yaml
fmuyassarov: It can be and it doesn't matter much here because eventually we will publish this chart as an OCI artifact same way as container image and during that release process we will have to patch the version to the same as the release version that we are about to cut. In other words, it will be part of the release process PR that I will submit as follow up once the chart lands on the main.

fmuyassarov:
Thanks for the reviews @AutuSnow , @pravk03. I've addressed most of the comments. PTAL.

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
fmuyassarov: thanks fixed

.github/workflows/helm-lint.yaml
AutuSnow: Makes sense

deployment/helm/dra-driver-cpu/templates/deviceclass.yaml
AutuSnow: Agree it mirrors install.yaml behavior. But Helm gives us a tool install.yaml doesn't — helm.sh/resource-policy: keep on the DeviceClass would prevent accidental deletion on helm uninstall while preserving create/upgrade semantics. Worth considering since DeviceClass is cluster-scoped shared state. Not a blocker if you prefer to defer.

fmuyassarov: That's interesting, thanks for sharing. I wasn't aware of `helm.sh/resource-policy: keep` annotation. The only thing I'm unsure about is - once this annotation is set, the resources becomes orphaned (according to this [document](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource)) and in case of a helm upgrade we won't be to replace the resource (when needed).

Are we okay with that?

deployment/helm/dra-driver-cpu/values.yaml
fmuyassarov: or to be exact, with `image.tag` empty values.yaml, the template falls back to `.Chart.AppVersion`, so the DaemonSet will use the same release version image.

.github/workflows/helm-lint.yaml
ffromani: @AutuSnow has a point, but I'd also prefer an incremental addition. +1 to start like this and iterate as needed.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
ffromani: For a 0.1.0 -> 0.2.0 migration path, I would agree we should make reasonnable effort to not hurt users, but OTOH some rough edges should be expected. We need to reserve the option to change things in a non-fully-backward compatible way at this stage, simply because we can't promise otherwise: we are largely still figuring out as we go.

In general, I'd like to deprecate `install.yaml` (cc @pravk03 ) as the helm chart is superior in any way. 
What we can do to bridge the gap from 0.1.0 install yaml to 0.2.0+ helm chart is to prepare a script which makes change (kubectl edit.../kubectl patch...) to deployed manifests such as they become forward compatible. This approach should be additive wrt this PR and reasonnably low effort.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -0,0 +1,172 @@
...
+          name: etc
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            {{- toYaml .Values.resources.requests | nindent 12 }}
+          {{- with .Values.resources.limits }}
+          limits:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
```
ffromani: the PR to remove the init container merged, so we can drop these high privileges (and following code)

deployment/helm/dra-driver-cpu/templates/deviceclass.yaml
ffromani: I for one would need to sit a bit longer and evaluate pros/cons here. Other reviewers are welcome to join.

deployment/helm/dra-driver-cpu/Chart.yaml
ffromani: I'd keep OWNERS for now indeed.

deployment/helm/dra-driver-cpu/README.md
```diff
@@ -0,0 +1,60 @@
...
+| `resources.requests.memory` | Memory resource request | `50Mi` |
+| `resources.limits` | Resource limits (unset by default) | `{}` |
+| `tolerations` | Node tolerations | all NoSchedule taints tolerated |
+| `args.logLevel` | Log verbosity (`--v`) | `4` |
+| `args.cpuDeviceMode` | CPU exposure mode: `grouped` or `individual` | `grouped` |
+| `args.groupBy` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode` or `socket` | `numanode` |
+| `args.reservedCPUs` | CPUs reserved for system/kubelet (e.g. `0-1`). Omitted if empty. | `""` |
+| `args.hostnameOverride` | Override the node name used by the driver. Omitted if empty. | `""` |
+| `healthzPath` | Path for liveness and readiness probes | `/healthz` |
+| `healthzPort` | Port the HTTP server binds to; used for the container port and probes | `8080` |
```
ffromani: is there a way to make sure this list is automatically kept up to date, or do we need to do it manually? Is it feasible (a followup PR is fine) to add some CI checks to verify the options list here match exactly the chart options? Perhaps even a trivial grep can be a good start I reckon.

deployment/helm/dra-driver-cpu/values.schema.json
```diff
@@ -0,0 +1,151 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
ffromani, line 2: likewise, any tooling to keep the schema up to date?

ffromani (commented):
I did only a high-level review. I need to doublecheck the equivalence (or look for and review the differences) between the generated manifests and our current manifests.

The direction LGTM

deployment/helm/dra-driver-cpu/README.md
fmuyassarov: There are options available, but the most commonly used one seems to be https://github.com/norwoodj/helm-docs. If this looks good to you, I can go ahead and add it

deployment/helm/dra-driver-cpu/values.schema.json
fmuyassarov: yes, perhaps this https://github.com/losisin/helm-values-schema-json?

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
fmuyassarov: Correct, updating it now...

deployment/helm/dra-driver-cpu/README.md
fmuyassarov: added https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83/changes/615324a221e7bb99430c713c3494c3c992e60f79

deployment/helm/dra-driver-cpu/values.schema.json
fmuyassarov: added https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83/changes/11161ecb83b3497be8686ae602e2dba6b98115e2

fmuyassarov:
Since we are already making most of the related changes to Helm here (automated docs & schema regeneration), I feel like it makes sense to include the remaining pieces as well. I’m planning to add one final commit to handle packaging and releasing the chart as an OCI artifact. If you’d prefer to keep that separate as a follow up PR, I’m happy to decouple it.

ffromani:
> Since we are already making most of the related changes to Helm here (automated docs & schema regeneration), I feel like it makes sense to include the remaining pieces as well. I’m planning to add one final commit to handle packaging and releasing the chart as an OCI artifact. If you’d prefer to keep that separate as a follow up PR, I’m happy to decouple it.

Thanks for the updates, I will review ASAP
I don't have strong opinions either way so we can proceed like you begun, in order to not waste effort already spent. If and only if the review highlights strong reasons to pivot, we will pivot later, in order to minimize the churn.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
fmuyassarov: Ah now I see what @AutuSnow was after. 

I guess, we currently expect people doing rolling update to migrate to `v0.2.0`.  But how do we migrate between versions and the tooling at the same time without disruption or less disruption is the hard part :). Because the label selectors differ between install.yaml and Helm, we can't do it in-place. So, I think the available options we have to offer are for migration are:

1. delete + reinstall (disruptive). Full cluster-wide disruption during the migration window
    ```
    kubectl delete -f install.yaml
    helm install dra-driver-cpu ... --version v0.2.0
    ```
2. helm adopt the existing (old version) resources. Helm can take ownership of existing resources by adding its labels/annotations without recreating them. But this won't work cleanly because the DaemonSet selector mismatch still can't be patched in-place.

3. Deploy the new Helm managed resources alongside the old ones, drain the old
one, delete it. Avoids disruption but requires some careful coordination. I think Helm will complain if some of the resources with the same naming already exist on the cluster. 

Realistically, first option is the only practical path I think even though quite undesirable. There is a disruption but the window is short (_I hope_),  just the time for pods to restart on all nodes. I think existing workloads will survive the restart since the pinning of CPU resources would be already in place.

fmuyassarov:
> > Since we are already making most of the related changes to Helm here (automated docs & schema regeneration), I feel like it makes sense to include the remaining pieces as well. I’m planning to add one final commit to handle packaging and releasing the chart as an OCI artifact. If you’d prefer to keep that separate as a follow up PR, I’m happy to decouple it.
> 
> Thanks for the updates, I will review ASAP I don't have strong opinions either way so we can proceed like you begun, in order to not waste effort already spent. If and only if the review highlights strong reasons to pivot, we will pivot later, in order to minimize the churn.

Sounds good. Then, I will keep the last bits as a separate/follow up PR.

AutuSnow:
I am wondering if there is any testing information that can be provided for this PR, because if we agree to this PR, it means that all our subsequent installations will rely on the helm file. Therefore, we need to set high standards as much as possible (except for non essential blocking behaviors) to avoid any negative experience or problems for users

fmuyassarov:
> I am wondering if there is any testing information that can be provided for this PR, because if we agree to this PR, it means that all our subsequent installations will rely on the helm file. Therefore, we need to set high standards as much as possible (except for non essential blocking behaviors) to avoid any negative experience or problems for users

Could you please clarify what you mean by testing information?

ffromani:
My initial assessment is if we do our due diligence and if there are no obvious gaps or major/unexplained differences between the new helm chart flow and the existing helm chart flow, then the PR is good enough and can be merged.

As for testing, add more CI lanes to exercise different deployment modes/helm chart parameters is an option I'm very open to pursue, let's just make sure the CI lanes are significantly different, e.g. let's not add say 10 small variations but rather 3 quite different modes (random number just to give some perspective).

fmuyassarov:
I think it’s best to take smaller, incremental steps and iterate as we go. There is definitely room for improvement, but it does not all need to happen at once. I’m happy to keep improving things over time and this approach should also make it easier for reviewers to focus on specific improvements step by step.

ffromani:
> I think it’s best to take smaller, incremental steps and iterate as we go. There is definitely room for improvement, but it does not all need to happen at once. I’m happy to keep improving things over time and this approach should also make it easier for reviewers to focus on specific improvements step by step.

+1 for this approach. I'll review again shortly.

AutuSnow:
I have already run the project to the production environment and it is very important. I do want to make a compromise, but I feel that the above migration plan did not meet expectations. Let it go, that's it

AutuSnow:
LGTM,The final review authority is delegated to@ffromani

fmuyassarov:
> I have already run the project to the production environment and it is very important. I do want to make a compromise, but I feel that the above migration plan did not meet expectations. Let it go, that's it

Thanks for sharing your concerns. If the migration plan didn’t meet your expectations, I’d really appreciate hearing more about what you think could be improved and more importantly how.  Given the current state of the project (still in v0.x.x), we do expect some level of iteration and change. This PR isn’t intended to introduce anything major, but rather to make the driver installation process smoother than it is today.
That said, I’m open to suggestions and happy to improve things further. But as said, let's take smaller steps at a time.

AutuSnow:
> > I have already run the project to the production environment and it is very important. I do want to make a compromise, but I feel that the above migration plan did not meet expectations. Let it go, that's it
> 
> Thanks for sharing your concerns. If the migration plan didn’t meet your expectations, I’d really appreciate hearing more about what you think could be improved and more importantly how. Given the current state of the project (still in v0.x.x), we do expect some level of iteration and change. This PR isn’t intended to introduce anything major, but rather to make the driver installation process smoother than it is today. That said, I’m open to suggestions and happy to improve things further. But as said, let's take smaller steps at a time.

Thank you for your reply. Let me clarify - what I am concerned about is not the Helm chart itself, but the migration path from the existing manifests/install.yaml to Helm.
Gap:
The user currently running VNet application-f manifest/install.yaml does not have a clear Helm upgrade path. PR did not address the following issues:
1. Coexistence: Can Helm managed resources coexist with existing VNet application resources? (The same resource name will result in conflicts)
2. Migration Guide: There is no record of the steps for users to convert from install. yaml → Helm does not need to be shut down
3. Backward compatibility: install.ml may be deprecated, but migration tools
What I am looking forward to:
-How to securely migrate existing deployments, or how to smoothly transition?
kubectl delete -f install.yaml 
Is helm install enough?

fmuyassarov:
The driver is stateless on restart and because of this, I _think_ (might be wrong since didn't test it myself) existing workloads should not be disrupted. Given that, the expected migration step is literally two commands:
```
 kubectl delete -f dist/install.yaml
 helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system
```
I've listed some additional options [here](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83#discussion_r3123028531)  though I haven't validated it it on a production setup.



I don't see a way to avoid a brief downtime window for new allocations during the rollover,  if you know of one,
I would love to hear it and we can document it. But meanwhile, I think we can add a short migration guide from install.yaml to helm the README. Would that address your concern?

fmuyassarov:
What I can do is test the migration and see what breaks in practice. That said, I don't think this should block the PR though - not everyone will be migrating; some users will simply start fresh from the latest release.

Would it make sense to continue the migration discussion on the original [issue](https://github.com/kubernetes-sigs/dra-driver-cpu/issues/72), where it has more context? That way this PR can move forward and we can iterate on the migration story separately?

ffromani:
@fmuyassarov I agree on the general sentiment: we can and should do due diligence to ensure no obvious breakage, but this should not block this PR.

My proposal is to add automated tests, which perhaps we can hook on CI, to ensure that install.yaml and the helm chart deliver both a functioning plugin. TL;DR: variants of the existing lanes.

We don't need to use prow: we can just set up in such a way the tests run against kind or minikube and temporarily hook those in the github actions. This is probably more convenient than a temporary prow setup.

ffromani:
> @fmuyassarov I agree on the general sentiment: we can and should do due diligence to ensure no obvious breakage, but this should not block this PR.
> 
> My proposal is to add automated tests, which perhaps we can hook on CI, to ensure that install.yaml and the helm chart deliver both a functioning plugin. TL;DR: variants of the existing lanes.
> 
> We don't need to use prow: we can just set up in such a way the tests run against kind or minikube and temporarily hook those in the github actions. This is probably more convenient than a temporary prow setup.

@AutuSnow will this proposal + the other ideas @fmuyassarov contributed address your concern?

AutuSnow:
I think I would agree

fmuyassarov:
> @fmuyassarov I agree on the general sentiment: we can and should do due diligence to ensure no obvious breakage, but this should not block this PR.
> 
> My proposal is to add automated tests, which perhaps we can hook on CI, to ensure that install.yaml and the helm chart deliver both a functioning plugin. TL;DR: variants of the existing lanes.
> 
> We don't need to use prow: we can just set up in such a way the tests run against kind or minikube and temporarily hook those in the github actions. This is probably more convenient than a temporary prow setup.

Thanks. I think we can cook up something relatively easy for that purpose. I will prepare it tomorrow.

ffromani:
> > @fmuyassarov I agree on the general sentiment: we can and should do due diligence to ensure no obvious breakage, but this should not block this PR.
> > My proposal is to add automated tests, which perhaps we can hook on CI, to ensure that install.yaml and the helm chart deliver both a functioning plugin. TL;DR: variants of the existing lanes.
> > We don't need to use prow: we can just set up in such a way the tests run against kind or minikube and temporarily hook those in the github actions. This is probably more convenient than a temporary prow setup.
> 
> Thanks. I think we can cook up something relatively easy for that purpose. I will prepare it tomorrow.

SGTM, thanks. Let's run these tests (perhaps for some time in GH actions CI) and we can move forward. No pressure, just recording the next steps.

fmuyassarov:
> What I can do is test the migration and see what breaks in practice. That said, I don't think this should block the PR though - not everyone will be migrating; some users will simply start fresh from the latest release.
> 
> Would it make sense to continue the migration discussion on the original [issue](https://github.com/kubernetes-sigs/dra-driver-cpu/issues/72), where it has more context? That way this PR can move forward and we can iterate on the migration story separately?

Sorry it took me a while to find some time to test this.

I can’t fully simulate a production-like environment, but here’s what I did. The cluster is running with version v0.1.0 of the driver, installed manually using the install.yaml file. I created two workload Pods, each requesting some CPUs via resourceClaims that reference the same deviceClass (`dra.cpu`).

Next I deleted the deviceClass (`dra.cpu`). As expected, this had no effect on the already running Pods. I then created two additional Pods requesting CPUs again referencing the same deviceClass. These Pods remained in the Pending state, since the deviceClass no longer existed.

While the Pods were still pending, I deleted the driver’s daemonSet and installed a new version of the driver with a local Helm chart. The new daemonSet Pods were up and running. The previously pending Pods moved to Running, and the Pods that were already running were not affected.

I know this is not a migration scenario, but it does shows that Pods with already allocated resources are not affected during such changes.

fmuyassarov:
@ffromani I’ve just pushed the https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83/changes/a5782fb182b9d433cd52f1d723f35db3a73cc0ea to add a minimal workflow. It builds the image, loads it into kind, installs the driver via Helm using that image, and then primarily checks for the existence of resourceSlice with the existing script. Here is the example run from this PR: https://github.com/kubernetes-sigs/dra-driver-cpu/actions/runs/25367882428/job/74383380807?pr=83.
So, we have https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83/changes/a5782fb182b9d433cd52f1d723f35db3a73cc0ea and https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83/changes/d7b830d2770ea0dfc0f855fb732b5ab3d6136f07 as a starting point to catch possible bugs for future helm part related patches.

fmuyassarov:
and once this is in place, I’ll do a follow up with another patch so that charts are published alongside the images in the production registry.

ffromani:
thanks for the updates. I'm reviewing but it will take me some more time to fully groc given my very rusty helm knowledge.

.github/workflows/helm-e2e.yaml
```diff
@@ -0,0 +1,91 @@
...
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: '1.25'
+
+      - name: Install kind
+        run: |
+          curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/v0.27.0/kind-linux-amd64
```
pravk03, line 33: nit: consider parameterizing the Kind version.

pravk03 (commented):
Thanks @fmuyassarov. From my limited helm knowledge, everything looked good on this PR.

fmuyassarov:
> Thanks @fmuyassarov. From my limited helm knowledge, everything looked good on this PR.

No worries and thanks for your review @pravk03 .

.github/workflows/helm-e2e.yaml
ffromani: +1. But can be done as followup and/or as sweeping update next time we bump kind version anyway. Thinking about it this would be the way I like most

ffromani: another thought. AFAIR kind (and kubectl?) come preinstalled in the worker images, is that version good/not good?

ffromani: xref: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
ffromani: I tend to agree. Let's proceed considering option 1 as the most likely.

ffromani (commented):
/approve

thanks for pushing this forward @fmuyassarov .

few more followups:
1. use helm as default install, deprecate `install.yaml`
2. document the update path + disruption (https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83/changes/988529f3616b35b734a8fc14168d920e4a228666#r3123028531)
3. for helm support in particular but in general simplify our CI steps leveraging pre-installed software in the runner images: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

ffromani:
/lgtm

from my side and also acknowledging https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83#pullrequestreview-4265949850

ffromani:
> /approve
> 
> thanks for pushing this forward @fmuyassarov .
> 
> few more followups:
> 
>     1. use helm as default install, deprecate `install.yaml`
> 
>     2. document the update path + disruption ([988529f#r3123028531](https://github.com/kubernetes-sigs/dra-driver-cpu/commit/988529f3616b35b734a8fc14168d920e4a228666#r3123028531))
> 
>     3. for helm support in particular but in general simplify our CI steps leveraging pre-installed software in the runner images: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

FYI @fmuyassarov / @pravk03 let's chat about these

fmuyassarov:
Thanks @ffromani. 
I will take care of those points as a follow up.

.github/workflows/helm-e2e.yaml
fmuyassarov: thanks, this is something I was unaware of. We could even drop the Helm.

fmuyassarov:
Filed a follow up in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/144.

PR #84, opened 2026-03-13
chore: add liveness & readiness probes to the container

fmuyassarov:
Add HTTP probes against `/healthz` on port `8080`.
Verified locally:
```
kubectl debug -it dracpu-kg258  -n kube-system --image=curlimages/curl --target=dracpu -- curl -v http://localhost:8080/healthz
```
```
Targeting container "dracpu". If you don't see processes from this container it may be because the container runtime doesn't support this feature.
--profile=legacy is deprecated and will be removed in the future. It is recommended to explicitly specify a profile, for example "--profile=general".
Defaulting debug container name to debugger-97wnz.
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Established connection to localhost (::1 port 8080) from ::1 port 36230
* using HTTP/1.x
> GET /healthz HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.18.0
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
< Date: Fri, 13 Mar 2026 03:16:51 GMT
< Content-Length: 0
<
* Connection #0 to host localhost:8080 left intact
```

### DIFF
--- a/manifests/base/daemonset-dracpu.part.yaml
+++ b/manifests/base/daemonset-dracpu.part.yaml
@@ -84,6 +84,19 @@ spec:
         - --cpu-device-mode=grouped
         image: us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu:latest
         imagePullPolicy: Always
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: http
+          initialDelaySeconds: 10
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: http
+          initialDelaySeconds: 5
+        ports:
+          - name: http
+            containerPort: 8080
         resources:
           requests:
             cpu: "100m"
--- a/test/e2e/healthz_test.go
+++ b/test/e2e/healthz_test.go
@@ -1,0 +1,187 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


AutuSnow:
Since we haven't set up a startupProbe, I'm wondering if there might be scenarios where the discovery of CPU resources on nodes takes a long time, resulting in an insufficient tolerance window for initialDelaySeconds? But unexpectedly, it seems to be working fine

ffromani:
@fmuyassarov thanks for the contributions, looks reasonnable and surely useful.
Please add e2e/integration tests (e2e probably are easier as we have some infra already) which verify the endpoints exist. Just testing the happy path is fine for now. We should likely start a new file in `test/e2e`

fmuyassarov:
@ffromani thanks for the review. Added simple test. Please take look.

fmuyassarov:
@pravk03 would you mind to take a look at this?

pravk03:
Sorry for the delay on this. I see that previous comments from @ffromani  and @AutuSnow have been addressed, and I don't have any additional comments.

One observation - because this is in the main test/e2e suite it will will run in both the e2e-individual-mode and e2e-grouped-mode CI jobs which is wasteful. I don't think it's a blocker for this PR, but we should look into preventing duplicate CI runs for mode agnostic tests in a separate PR.

/lgtm
/approve

fmuyassarov:
> Sorry for the delay on this. I see that previous comments from @ffromani and @AutuSnow have been addressed, and I don't have any additional comments.
> 
> One observation - because this is in the main test/e2e suite it will will run in both the e2e-individual-mode and e2e-grouped-mode CI jobs which is wasteful. I don't think it's a blocker for this PR, but we should look into preventing duplicate CI runs for mode agnostic tests in a separate PR.
> 
> /lgtm /approve

Good point about duplication @pravk03. I can file a patch as a follow up to address that.

fmuyassarov:
Assuming there is no more comments to address, removing the hold
/hold cancel

PR #90, opened 2026-03-17
fix:NRI restarts the loop and calls klog.Fatalf when the context is canceled

AutuSnow:
NRI restart logic: When the context is normally canceled, Run() returns, but the code still enters the loop counting, ultimately calling Fatalf, leading to program crash instead of graceful exit

https://github.com/kubernetes-sigs/dra-driver-cpu/issues/89

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -176,19 +176,9 @@ func Start(ctx context.Context, clientset kubernetes.Interface, config *Config) 
 	plugin.nriPlugin = stub
 
 	go func() {
-		for i := 0; i < maxAttempts; i++ {
-			err = plugin.nriPlugin.Run(ctx)
-			if err != nil {
-				klog.Infof("NRI plugin failed with error %v", err)
-			}
-			select {
-			case <-ctx.Done():
-				return
-			default:
-				klog.Infof("Restarting NRI plugin %d out of %d", i, maxAttempts)
-			}
+		if err := runNRIPluginWithRetry(ctx, plugin.nriPlugin, maxAttempts); err != nil && ctx.Err() == nil {
+			klog.Fatalf("NRI plugin failed for %d times to be restarted", maxAttempts)
 		}
-		klog.Fatalf("NRI plugin failed for %d times to be restarted", maxAttempts)
 	}()
 
 	// publish available resources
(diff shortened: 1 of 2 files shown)


AutuSnow:
/cc @ffromani @pravk03
Please review this PR when you have time

pkg/driver/driver.go
```diff
@@ -207,3 +197,21 @@ func (cp *CPUDriver) Stop() {
 func (cp *CPUDriver) Shutdown(_ context.Context) {
 	klog.Info("Runtime shutting down...")
 }
+
+type nriRunner interface {
+	Run(context.Context) error
+}
+
+func runNRIPluginWithRetry(ctx context.Context, plugin nriRunner, maxAttempts int) error {
+	for i := 0; i < maxAttempts; i++ {
```
pravk03, line 206: Should we introduce some backoff between retries ?

pkg/driver/driver_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	err := runNRIPluginWithRetry(ctx, runner, maxAttempts)
+	require.ErrorIs(t, err, context.Canceled, "should return context.Canceled when context is cancelled")
+	require.Equal(t, int32(1), runner.calls.Load(), "Run should be called exactly once before context cancel")
+}
+
+func TestRunNRIPluginWithRetry_ContextCancelledAfterSeveralRetries(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	var calls atomic.Int32
```
pravk03, line 56: nit: Do we need atomic here ?. Since the for loop in `runNRIPluginWithRetry` runs synchronously ?

pkg/driver/driver.go
AutuSnow: Good question. For the NRI plugin reboot, I don't think we need to roll back because NRI reconnection is usually quick (within seconds) and the maxAttempts limit has prevented infinite loops, and I think adding a rollback would unnecessarily delay recovery

pkg/driver/driver_test.go
AutuSnow: You're right, atoms are not absolutely necessary here, however, I think it's a good thing to keep them as a defensive exercise in case we add concurrent test cases in the future. Of course, I don't mind making changes

pravk03:
/lgtm
/approve 
Thanks!

PR #106, opened 2026-04-01
Update CPU topology discovery to support ARM

pravk03:
* Refactor the CPU discovery mechanism from using ` /proc/cpuinfo` to sysfs based approach (`/sys/devices/system/cpu`)
* Implemented a fallback for L3 cache discovery when cache id file is not present
* Updated SMT detection to correctly handle `notimplemented` status

### DIFF
--- a/pkg/cpuinfo/cpuinfo.go
+++ b/pkg/cpuinfo/cpuinfo.go
@@ -199,7 +199,7 @@ func (s *SystemCPUInfo) IsSMTEnabled() (bool, error) {
 	if status == "on" {
 		return true, nil
 	}
-	if status == "off" || status == "forceoff" || status == "notsupported" {
+	if status == "off" || status == "forceoff" || status == "notsupported" || status == "notimplemented" {
 		return false, nil
 	}
 	return false, fmt.Errorf("unknown SMT status: %s", status)
--- a/pkg/cpuinfo/cpuinfo.go
+++ b/pkg/cpuinfo/cpuinfo.go
@@ -103,11 +103,11 @@ type CPUInfo struct {
 	// NUMANodeID is the NUMA node ID, unique within each SocketID
 	NUMANodeID int `json:"numaNodeID"`
 
-	// NUMA Node Affinity Mask
-	NumaNodeAffinityMask string `json:"numaNodeAffinityMask"`
+	// NUMANodeCPUSet represents the set of CPUs that are in the same NUMA node.
+	NumaNodeCPUSet cpuset.CPUSet `json:"numaNodeCPUSet"`
 
 	// CPU Sibling of the CpuID
-	SiblingCpuID int `json:"sibling"`
+	SiblingCPUID int `json:"sibling"`
 
 	// Core Type (e-core or p-core)
 	CoreType CoreType `json:"coreType,omitempty"`
@@ -237,15 +237,15 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
 	cpuInfos := make([]CPUInfo, 0, onlineCPUs.Size())
 	for _, cpuID := range onlineCPUs.List() {
 		cpuInfo := CPUInfo{
-			CpuID:                cpuID,
-			SocketID:             -1,
-			CoreID:               -1,
-			ClusterID:            -1,
-			NUMANodeID:           -1,
-			NumaNodeAffinityMask: "",
-			UncoreCacheID:        -1,
-			SiblingCpuID:         -1,
-			CoreType:             CoreTypeUndefined,
+			CpuID:          cpuID,
+			SocketID:       -1,
+			CoreID:         -1,
+			ClusterID:      -1,
+			NUMANodeID:     -1,
+			NumaNodeCPUSet: cpuset.New(),
+			UncoreCacheID:  -1,
+			SiblingCPUID:   -1,
+			CoreType:       CoreTypeUndefined,
 		}
 
 		if isHybrid {
@@ -280,20 +280,30 @@ func populateTopologyInfo(cpuInfo *CPUInfo) error {
 	if err != nil {
 		return fmt.Errorf("could not read socket_id for cpu %d from sysfs: %w", cpuID, err)
 	}
-	socketID, _ := strconv.Atoi(strings.TrimSpace(socketStr))
+	socketID, err := strconv.Atoi(strings.TrimSpace(socketStr))
+	if err != nil {
+		return fmt.Errorf("could not parse socket_id %q for cpu %d: %w", socketStr, cpuID, err)
+	}
 	cpuInfo.SocketID = socketID
 
-	// Get Cluster ID from sysfs (for ARM)
+	// Get Cluster ID from sysfs. While this sysfs file is present on most
+	// architectures, on ARM this defines the physical boundary for shared resources (like L2 cache).
 	clusterPath := hostSys(fmt.Sprintf("devices/system/cpu/cpu%d/topology/cluster_id", cpuID))
 	clusterStr, err := ReadFile(clusterPath)
 	if err == nil {
-		clusterID, _ := strconv.Atoi(strings.TrimSpace(clusterStr))
-		// On many x86 systems, the kernel reports 65535 to indicate "No Cluster".
-		// We treat this as -1 to mean unknown/not applicable.
-		if clusterID == 65535 {
+		clusterID, err := strconv.Atoi(strings.TrimSpace(clusterStr))
+		if err != nil {
+			log.Printf("Warning: could not parse cluster_id %q for cpu %d: %v", clusterStr, cpuID, err)
 			cpuInfo.ClusterID = -1
 		} else {
-			cpuInfo.ClusterID = clusterID
+			// The kernel default for an undefined cluster id is -1.
+			// However, due to 16-bit unsigned type casting, sysfs exposes this as 65535.
+			// https://www.kernel.org/doc/Documentation/admin-guide/cputopology.rst
+			if clusterID == 65535 {
+				cpuInfo.ClusterID = -1
+			} else {
+				cpuInfo.ClusterID = clusterID
+			}
 		}
 	} else {
 		cpuInfo.ClusterID = -1 // Default to -1 if not present
@@ -305,7 +315,10 @@ func populateTopologyInfo(cpuInfo *CPUInfo) error {
 	if err != nil {
 		return fmt.Errorf("could not read core_id for cpu %d from sysfs: %w", cpuID, err)
 	}
-	coreID, _ := strconv.Atoi(strings.TrimSpace(coreStr))
+	coreID, err := strconv.Atoi(strings.TrimSpace(coreStr))
+	if err != nil {
+		return fmt.Errorf("could not parse core_id %q for cpu %d: %w", coreStr, cpuID, err)
+	}
 	cpuInfo.CoreID = coreID
 
 	// Get NUMA Node ID from sysfs
@@ -318,21 +331,26 @@ func populateTopologyInfo(cpuInfo *CPUInfo) error {
 	foundNode := false
 	for _, file := range files {
 		if strings.HasPrefix(file.Name(), "node") {
-			nodeID, err := strconv.ParseInt(strings.TrimPrefix(file.Name(), "node"), 10, 64)
+			nodeID, err := strconv.Atoi(strings.TrimPrefix(file.Name(), "node"))
 			if err != nil {
-				continue // Should not happen with a well-formed sysfs
+				log.Printf("Warning: could not parse NUMA node ID from %s: %v", file.Name(), err)
+				continue
 			}
-			cpuInfo.NUMANodeID = int(nodeID)
+			cpuInfo.NUMANodeID = nodeID
 			foundNode = true
 
 			// Get NUMA Affinity Mask
-			maskPath := hostSys(fmt.Sprintf("devices/system/node/node%d/cpumap", nodeID))
-			maskLines, err := ReadLines(maskPath)
-			if err != nil {
+			cpuListPath := hostSys(fmt.Sprintf("devices/system/node/node%d/cpulist", nodeID))
+			cpuListLines, err := ReadLines(cpuListPath)
+			if err != nil || len(cpuListLines) == 0 {
 				log.Printf("Warning: could not read NUMA affinity mask for node %d: %v", nodeID, err)
 				continue
 			}
-			cpuInfo.NumaNodeAffinityMask = formatAffinityMask(maskLines[0])
+			cpuInfo.NumaNodeCPUSet, err = cpuset.Parse(cpuListLines[0])
+			if err != nil {
+				log.Printf("Warning: could not parse NUMA affinity mask for node %d: %v", nodeID, err)
+				continue
+			}
 			break
 		}
 	}
@@ -382,22 +400,22 @@ func populateTopologyInfo(cpuInfo *CPUInfo) error {
 
 		cacheIdPath := filepath.Join(l3CacheDir, "id")
 		idStr, err := ReadFile(cacheIdPath)
-		var id int64
+		var id int
 		if err != nil {
 			// Fallback for ARM systems missing the 'id' file: use the lowest shared CPU ID as the cache ID.
 			if sharedCPUSet.Size() > 0 {
-				id = int64(sharedCPUSet.List()[0])
+				id = sharedCPUSet.List()[0]
 			} else {
-				id = int64(cpuID)
+				id = cpuID
 			}
 		} else {
-			id, err = strconv.ParseInt(strings.TrimSpace(idStr), 10, 64)
+			id, err = strconv.Atoi(strings.TrimSpace(idStr))
 			if err != nil {
 				return fmt.Errorf("could not parse L3 cache id '%s': %w", idStr, err)
 			}
 		}
 
-		cpuInfo.UncoreCacheID = int(id)
+		cpuInfo.UncoreCacheID = id
 		break
 	}
 
@@ -433,27 +451,10 @@ func populateCpuSiblings(cpuInfos []CPUInfo) {
 			cpu1Id, cpu2Id := siblingIds[0], siblingIds[1]
 			cpu1Index, cpu2Index := cpuIndexMap[cpu1Id], cpuIndexMap[cpu2Id]
 
-			cpuInfos[cpu1Index].SiblingCpuID = cpu2Id
-			cpuInfos[cpu2Index].SiblingCpuID = cpu1Id
+			cpuInfos[cpu1Index].SiblingCPUID = cpu2Id
+			cpuInfos[cpu2Index].SiblingCPUID = cpu1Id
 		}
 	}
-}
-
-func formatAffinityMask(mask string) string {
-	if strings.TrimSpace(mask) == "" {
-		return "0x0"
-	}
-	parts := strings.Split(mask, ",")
-	// Reverse the parts to handle the kernel's little-endian word order.
-	for i, j := 0, len(parts)-1; i < j; i, j = i+1, j-1 {
-		parts[i], parts[j] = parts[j], parts[i]
-	}
-	newMask := strings.Join(parts, "")
-	newMask = strings.TrimLeft(newMask, "0")
-	if newMask == "" {
-		return "0x0"
-	}
-	return "0x" + newMask
 }
 
 // ReadFile reads contents from a file.
--- a/pkg/cpuinfo/cpuinfo.go
+++ b/pkg/cpuinfo/cpuinfo.go
@@ -299,15 +299,6 @@ func populateL3CacheIDs(cpuInfos []CPUInfo) error {
 
 			if strings.TrimSpace(levelStr) == "3" {
 				l3CacheDir := filepath.Join(cachePath, entry.Name())
-				cacheIdPath := filepath.Join(l3CacheDir, "id")
-				idStr, err := ReadFile(cacheIdPath)
-				if err != nil {
-					return fmt.Errorf("could not read L3 cache id from %s: %w", cacheIdPath, err)
-				}
-				id, err := strconv.ParseInt(strings.TrimSpace(idStr), 10, 64)
-				if err != nil {
-					return fmt.Errorf("could not parse L3 cache id '%s': %w", idStr, err)
-				}
 
 				sharedCPUListPath := filepath.Join(l3CacheDir, "shared_cpu_list")
 				sharedCPUListStr, err := ReadFile(sharedCPUListPath)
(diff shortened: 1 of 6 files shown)


pravk03:
Verified this locally on the ARM machine I have access to, but we need to ensure the underlying logic covers all diverse ARM cases and we are not missing anything. 

/cc @ffromani @catblade 
/hold

AutuSnow (commented):
@pravk03  I see that in the new code, if there is no difference between CoreID and SocketID, the default value of -1 will still be retained. The old code will make a judgment, and if it is less than 0, it will be skipped. Should we also retain this boundary judgment method?

pravk03:
Good point @AutuSnow. Updated to add the boundary check back. We currently just log errors and exclude the CPU from the ResourceSlice.

AutuSnow:
Thank you for your modification. In my opinion, there are no further issues.
/lgtm

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,183 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
+	cpuInfo.SocketID = socketID
+
+	// Get Cluster ID from sysfs (for ARM)
+	clusterPath := hostSys(fmt.Sprintf("devices/system/cpu/cpu%d/topology/cluster_id", cpuID))
+	clusterStr, err := ReadFile(clusterPath)
+	if err == nil {
+		clusterID, _ := strconv.Atoi(strings.TrimSpace(clusterStr))
+		cpuInfo.ClusterID = clusterID
+	} else {
+		cpuInfo.ClusterID = 0 // Default to 0 if not present
```
AutuSnow: I remember on x86 systems, the ClusterID of all CPUs is set to 0. Will this lead to false sharing or incorrect core counts, and should we use the default value of -1 to indicate unknown

ffromani:
Maybe we discussed this in the past but still: my main concern is to have a reliable testing pipeline in place which exercises the code. Are there good enough (or at all?) ARM machines we can leverage from github actions and/or from prow?

I understand the requirements and I agree is nice to support ARM, but without CI in place the support itself will be brittle and unreliable, not great for users.

fmuyassarov:
> I understand the requirements and I agree is nice to support ARM, but without CI in place the support itself will be brittle and unreliable, not great for users.

One option, is to request CNCF service desk for ARM machines from [Oracle Cloud](https://contribute.cncf.io/blog/2025/11/08/maintainer-summit-update-from-the-project-staff#akamailinode) as they have done some donation to CNCF.
>
>_[Akamai pledged $700K in cloud credits to the CNCF for project use](https://www.cncf.io/blog/2025/11/05/akamai-builds-cloud-native-resilience-cloud-credits-to-power-cncf-projects/). Projects can now request access to onboard Linode. This platform provides another alternative to [Equinix which is due to sunset later this year](https://www.cncf.io/blog/2025/06/04/thank-you-equinix-metal-the-cncf-community-bids-farewell-to-the-bare-metal-cluster/). Oracle Cloud credits also continue to be available, particularly for [AI workloads on ARM and bare metal deployments](https://www.cncf.io/blog/2024/02/02/oracle-oci-credits-are-now-available-to-cncf-projects-here-is-what-you-need-to-know/). Request access to either service via the Service Desk._

fmuyassarov:
and then we could even setup self hosted github runners or things like that

pravk03:
I am thinking about how we handle this in kubelet today and whether we can follow the same approach here. I believe kubelet relies on cAdvisor for all topology information. For the long term, would it make sense for us to adopt the same mechanism here?

pkg/cpuinfo/cpuinfo.go
pravk03: Yeah. -1 is a better default. Updated. 
On x86, the cluster_id file is present but reports 65535 to indicate unknown.  I am also setting that to -1.

ffromani:
Supporting the same ARM flavours kubelet does through CNCF totally makes sense and it's a great start. This would address all my concerns. I'd gate ARM support to having secured availability of these machines for CI.

pravk03:
Once https://github.com/kubernetes/test-infra/pull/36640 merges, I think we can have arm e2e tests on prow ([example](https://github.com/kubernetes/test-infra/blob/686cb0983f17f1cf42f1002346dc12a5d25c4580/config/jobs/etcd/etcd-postsubmits.yaml#L153))

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
+	// Get Socket ID from sysfs
+	socketPath := hostSys(fmt.Sprintf("devices/system/cpu/cpu%d/topology/physical_package_id", cpuID))
+	socketStr, err := ReadFile(socketPath)
+	if err != nil {
+		return fmt.Errorf("could not read socket_id for cpu %d from sysfs: %w", cpuID, err)
+	}
+	socketID, _ := strconv.Atoi(strings.TrimSpace(socketStr))
+	cpuInfo.SocketID = socketID
+
+	// Get Cluster ID from sysfs (for ARM)
```
ffromani: AFAIK the file exists also on x86 (in all the arches supported by linux?) but on x86 the content is not significant as is on ARM. So we may just rephrase as
```
Get ClusterID from sysfs (relevant on ARM)
```

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
+	clusterPath := hostSys(fmt.Sprintf("devices/system/cpu/cpu%d/topology/cluster_id", cpuID))
+	clusterStr, err := ReadFile(clusterPath)
+	if err == nil {
+		clusterID, _ := strconv.Atoi(strings.TrimSpace(clusterStr))
+		// On many x86 systems, the kernel reports 65535 to indicate "No Cluster".
+		// We treat this as -1 to mean unknown/not applicable.
+		if clusterID == 65535 {
+			cpuInfo.ClusterID = -1
```
ffromani: Is this behavior documented somewhere in the kernel docs? if so please add a link.
It would also be nice to move the magic numbers as module-private constants.

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
+		cpuInfo.ClusterID = -1 // Default to -1 if not present
+	}
+
+	// Get Core ID from sysfs
+	corePath := hostSys(fmt.Sprintf("devices/system/cpu/cpu%d/topology/core_id", cpuID))
+	coreStr, err := ReadFile(corePath)
+	if err != nil {
+		return fmt.Errorf("could not read core_id for cpu %d from sysfs: %w", cpuID, err)
 	}
+	coreID, _ := strconv.Atoi(strings.TrimSpace(coreStr))
```
ffromani: we should probably manage the error in the string->int conversion, because an error here is extremely unlikely and therefore would be super hard to debug (here and in the similar code paths)

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
-			}
-		case "physical id":
-			if val, err = strconv.Atoi(value); err != nil {
-				log.Printf("Warning: failed to parse physical ID %q: %v", value, err)
-			} else {
-				cpuInfo.SocketID = val
+	foundNode := false
+	for _, file := range files {
+		if strings.HasPrefix(file.Name(), "node") {
+			nodeID, err := strconv.ParseInt(strings.TrimPrefix(file.Name(), "node"), 10, 64)
```
ffromani: why do we use sometimes Atoi and sometimes ParseInt? do we need a int64 here?

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
+			if err != nil {
+				continue // Should not happen with a well-formed sysfs
```
ffromani: indeed, and therefore it deserves a log

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
-		case "core id":
-			if val, err = strconv.Atoi(value); err != nil {
-				log.Printf("Warning: failed to parse core ID %q: %v", value, err)
-			} else {
-				cpuInfo.CoreID = val
+			cpuInfo.NUMANodeID = int(nodeID)
+			foundNode = true
+
+			// Get NUMA Affinity Mask
+			maskPath := hostSys(fmt.Sprintf("devices/system/node/node%d/cpumap", nodeID))
```
ffromani: do we want to use `cpulist` here?

pkg/cpuinfo/cpuinfo.go
```diff
@@ -223,235 +234,189 @@ func (s *SystemCPUInfo) GetCPUInfos() ([]CPUInfo, error) {
...
-		}
+
+		cpuInfo.UncoreCacheID = int(id)
+		break
 	}
+
 	return nil
 }
 
+// TODO: Handle more complex sibling relationships (e.g. 4-way SMT) if needed in the future. For now we only handle 2-way hyperthreading which is the most common case.
```
ffromani, line 425: I agree: let's note this may happen but let's not implement if we can't test it on real HW

ffromani (commented):
Initial review, will need another pass. Nothing concerning so far and some nice improvements worth merging anyway.
It seems we have a bit of chicken/egg problem: we can't enable ARM CI without this PR, but we can't test this PR without CI _OR_ without developers getting access to ARM boxes, which is unpractical.

But I'm positive about this PR, so we can probably bootstrap merging this code first.

fmuyassarov:
From a quick test run on my Pi, everything looked correct except I noticed that the NUMA id is missing.
```YAML
  ownerReferences:
  - apiVersion: v1
    controller: true
    kind: Node
    name: dra-driver-cpu-control-plane
    uid: 4806a758-aa47-4075-8088-a63328e3d1fb
  resourceVersion: "1421"
  uid: e3d80168-473d-4b01-bcd3-adf7cd910253
spec:
  devices:
  - attributes:
      dra.cpu/cacheL3ID:
        int: -1
      dra.cpu/coreID:
        int: 0
      dra.cpu/coreType:
        string: standard
      dra.cpu/cpuID:
        int: 0
      dra.cpu/numaNodeID:
        int: 0
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 0
      dra.net/numaNode:
        int: 0
    name: cpudev000
  - attributes:
      dra.cpu/cacheL3ID:
        int: -1
      dra.cpu/coreID:
        int: 1
      dra.cpu/coreType:
        string: standard
      dra.cpu/cpuID:
        int: 1
      dra.cpu/numaNodeID:
        int: 0
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 0
      dra.net/numaNode:
        int: 0
    name: cpudev001
  - attributes:
      dra.cpu/cacheL3ID:
        int: -1
      dra.cpu/coreID:
        int: 2
      dra.cpu/coreType:
        string: standard
      dra.cpu/cpuID:
        int: 2
      dra.cpu/numaNodeID:
        int: 0
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 0
      dra.net/numaNode:
        int: 0
    name: cpudev002
  - attributes:
      dra.cpu/cacheL3ID:
        int: -1
      dra.cpu/coreID:
        int: 3
      dra.cpu/coreType:
        string: standard
      dra.cpu/cpuID:
        int: 3
      dra.cpu/numaNodeID:
        int: 0
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 0
      dra.net/numaNode:
        int: 0
    name: cpudev003
  driver: dra.cpu
  nodeName: dra-driver-cpu-control-plane
```



```
$ lscpu
Architecture:                aarch64
  CPU op-mode(s):            32-bit, 64-bit
  Byte Order:                Little Endian
CPU(s):                      4
  On-line CPU(s) list:       0-3
Vendor ID:                   ARM
  Model name:                Cortex-A72
    Model:                   3
    Thread(s) per core:      1
    Core(s) per cluster:     4
    Socket(s):               -
    Cluster(s):              1
    Stepping:                r0p3
    CPU(s) scaling MHz:      100%
    CPU max MHz:             1800.0000
    CPU min MHz:             600.0000
    BogoMIPS:                108.00
    Flags:                   fp asimd evtstrm crc32 cpuid
Caches (sum of all):
  L1d:                       128 KiB (4 instances)
  L1i:                       192 KiB (4 instances)
  L2:                        1 MiB (1 instance)
NUMA:
  NUMA node(s):              2
  NUMA node0 CPU(s):         0-3
  NUMA node1 CPU(s):         0-3
...
```
  
```
$ numactl --hardware && echo "---" && lscpu | grep -E 'Socket|NUMA|Core|Thread'
available: 2 nodes (0-1)
node 0 cpus: 0 1 2 3
node 0 size: 1853 MB
node 0 free: 140 MB
node 1 cpus: 0 1 2 3
node 1 size: 1943 MB
node 1 free: 94 MB
node distances:
node     0    1
   0:   10   10
   1:   10   10
---
Thread(s) per core:                      1
Core(s) per cluster:                     4
Socket(s):                               -
NUMA node(s):                            2
NUMA node0 CPU(s):                       0-3
NUMA node1 CPU(s):                       0-3
```
  

```YAML
Defaulted container "dracpu" out of: dracpu, enable-nri-and-cdi (init)
I0409 13:49:05.294122    2405 app.go:236] dracpu go go1.25.9 build: fa54c505a210940fdbc0290a77e8bb85cef48826 time: 2026-04-09T12:51:01Z
I0409 13:49:05.294362    2405 app.go:118] FLAG: --add_dir_header="false"
I0409 13:49:05.294399    2405 app.go:118] FLAG: --alsologtostderr="false"
I0409 13:49:05.294408    2405 app.go:118] FLAG: --bind-address=":8080"
I0409 13:49:05.294420    2405 app.go:118] FLAG: --cpu-device-mode="individual"
I0409 13:49:05.294433    2405 app.go:118] FLAG: --group-by="numanode"
I0409 13:49:05.294444    2405 app.go:118] FLAG: --hostname-override=""
I0409 13:49:05.294453    2405 app.go:118] FLAG: --kubeconfig=""
I0409 13:49:05.294460    2405 app.go:118] FLAG: --log_backtrace_at=":0"
I0409 13:49:05.294484    2405 app.go:118] FLAG: --log_dir=""
I0409 13:49:05.294493    2405 app.go:118] FLAG: --log_file=""
I0409 13:49:05.294500    2405 app.go:118] FLAG: --log_file_max_size="1800"
I0409 13:49:05.294516    2405 app.go:118] FLAG: --logtostderr="true"
I0409 13:49:05.294523    2405 app.go:118] FLAG: --one_output="false"
I0409 13:49:05.294530    2405 app.go:118] FLAG: --reserved-cpus=""
I0409 13:49:05.294537    2405 app.go:118] FLAG: --skip_headers="false"
I0409 13:49:05.294544    2405 app.go:118] FLAG: --skip_log_headers="false"
I0409 13:49:05.294552    2405 app.go:118] FLAG: --stderrthreshold="2"
I0409 13:49:05.294562    2405 app.go:118] FLAG: --v="4"
I0409 13:49:05.294579    2405 app.go:118] FLAG: --vmodule=""
I0409 13:49:05.296637    2405 envvar.go:172] "Feature gate default state" feature="InOrderInformers" enabled=true
I0409 13:49:05.296717    2405 envvar.go:172] "Feature gate default state" feature="InOrderInformersBatchProcess" enabled=true
I0409 13:49:05.296772    2405 envvar.go:172] "Feature gate default state" feature="InformerResourceVersion" enabled=true
I0409 13:49:05.296793    2405 envvar.go:172] "Feature gate default state" feature="WatchListClient" enabled=true
I0409 13:49:05.296807    2405 envvar.go:172] "Feature gate default state" feature="ClientsAllowCBOR" enabled=false
I0409 13:49:05.296832    2405 envvar.go:172] "Feature gate default state" feature="ClientsPreferCBOR" enabled=false
I0409 13:49:05.302406    2405 draplugin.go:602] "Starting"
I0409 13:49:05.303261    2405 nonblockinggrpcserver.go:90] "GRPC server started" logger="dra" endpoint="/var/lib/kubelet/plugins/dra.cpu/dra.sock"
I0409 13:49:05.303719    2405 nonblockinggrpcserver.go:90] "GRPC server started" logger="registrar" endpoint="/var/lib/kubelet/plugins_registry/dra.cpu-reg.sock"
I0409 13:49:07.304262    2405 cdi.go:76] Initialized CDI file manager for "/var/run/cdi/dra.cpu.json"
time="2026-04-09T13:49:07Z" level=info msg="Created plugin 00-dra.cpu (dracpu, handles CreateContainer,StopContainer,RemoveContainer)"
I0409 13:49:07.304877    2405 app.go:204] driver started
I0409 13:49:07.305027    2405 dra_hooks.go:223] Publishing resources
I0409 13:49:07.305879    2405 resourceslicecontroller.go:538] "Starting ResourceSlice informer and waiting for it to sync" logger="ResourceSlice controller"
I0409 13:49:07.306187    2405 reflector.go:367] "Starting reflector" logger="ResourceSlice controller" type="*v1.ResourceSlice" resyncPeriod="0s" reflector="pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/reflector.go:289"
I0409 13:49:07.306252    2405 reflector.go:411] "Listing and watching" logger="ResourceSlice controller" type="*v1.ResourceSlice" reflector="pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/reflector.go:289"
time="2026-04-09T13:49:07Z" level=info msg="Registering plugin 00-dra.cpu..."
time="2026-04-09T13:49:07Z" level=info msg="Configuring plugin 00-dra.cpu for runtime containerd/v2.2.0..."
time="2026-04-09T13:49:07Z" level=info msg="Started plugin 00-dra.cpu..."
I0409 13:49:07.360822    2405 reflector.go:978] "Exiting watch because received the bookmark that marks the end of initial events stream" logger="ResourceSlice controller" reflector="pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/reflector.go:289" totalItems=1 duration="54.187257ms"
I0409 13:49:07.361135    2405 reflector.go:446] "Caches populated" logger="ResourceSlice controller" type="*v1.ResourceSlice" reflector="pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/reflector.go:289"
I0409 13:49:07.387422    2405 nri_hooks.go:33] Synchronized state with the runtime (10 pods, 10 containers)...
I0409 13:49:07.387791    2405 nri_hooks.go:41] Synchronize pod kube-system/kindnet-gc4wj UID 93c7fefd-9cb7-4688-9457-c13e5671ba19
I0409 13:49:07.387889    2405 nri_hooks.go:41] Synchronize pod kube-system/coredns-7d764666f9-j8ldv UID b3521ec7-baa2-4d34-b630-1fd7ba302d30
I0409 13:49:07.387908    2405 nri_hooks.go:41] Synchronize pod kube-system/coredns-7d764666f9-8grpj UID 5fdb8242-071d-4a7d-84e3-d074fd4be08f
I0409 13:49:07.387923    2405 nri_hooks.go:41] Synchronize pod kube-system/dracpu-2nxh6 UID b8c35ec3-4666-40e6-9fd6-eca33a9038c6
I0409 13:49:07.387936    2405 nri_hooks.go:41] Synchronize pod kube-system/kube-apiserver-dra-driver-cpu-control-plane UID 37ed9f0908e60821052c00fd207c2940
I0409 13:49:07.387950    2405 nri_hooks.go:41] Synchronize pod local-path-storage/local-path-provisioner-67b8995b4b-drswn UID fb74fb34-6064-43a4-ad28-bf74b59ac871
I0409 13:49:07.387961    2405 nri_hooks.go:41] Synchronize pod kube-system/etcd-dra-driver-cpu-control-plane UID 1f74daaa1d7133221e2c745d85d4a2dc
I0409 13:49:07.387972    2405 nri_hooks.go:41] Synchronize pod kube-system/kube-proxy-q8gwb UID a809b7a5-298d-48b7-8bd3-1c1d61dbbf1c
I0409 13:49:07.387984    2405 nri_hooks.go:41] Synchronize pod kube-system/kube-controller-manager-dra-driver-cpu-control-plane UID e6452df7fa0fae1030da6ad72e966c17
I0409 13:49:07.388004    2405 nri_hooks.go:41] Synchronize pod kube-system/kube-scheduler-dra-driver-cpu-control-plane UID 0f01c9d9e97c323e3aa548ef8e0e1a1f
I0409 13:49:08.306950    2405 resourceslicecontroller.go:553] "ResourceSlice informer has synced" logger="ResourceSlice controller"
I0409 13:49:08.307061    2405 resourceslicecontroller.go:184] "Starting" logger="ResourceSlice controller"
```

pravk03:
Thanks for testing the changes locally @fmuyassarov. Taking a look.

fmuyassarov:
> Thanks for testing the changes locally @fmuyassarov. Taking a look.

To add here as well as we discussed with @pravk03 offline. I realized from the output of numactl, that the same CPU belongs to more than one NUMA
```
NUMA node0 CPU(s):   0-3
NUMA node1 CPU(s):   0-3
```
This isn't a real NUMA setup I would say. Normally each CPU belongs to exactly one NUMA node, but here all 4 cores appear in both nodes which shouldn't happen. Although it is weird - it is common for Raspberry Pi 4 (BCM2711 SoC) which exposes two regions in its physical address map, a low memory window and a high memory window.  Basically, it was not the best idea to try out the changes on the Pi :D. Sorry for confusion @pravk03

pkg/cpuinfo/cpuinfo.go
pravk03: Good idea. We would not need custom functions to format the mask.
Updated.

pkg/cpuinfo/cpuinfo.go
pravk03: added

pkg/cpuinfo/cpuinfo.go
pravk03: We can use Atoi everywhere. Updated.

pkg/cpuinfo/cpuinfo.go
pravk03: Updated.

pkg/cpuinfo/cpuinfo.go
pravk03: Updated the comment and added link to kernel documentation

pkg/cpuinfo/cpuinfo.go
pravk03: Updated the comment

pravk03:
The flaky presubmit should be fixed by https://github.com/kubernetes-sigs/dra-driver-cpu/pull/116

pkg/cpuinfo/cpuinfo.go
```diff
@@ -97,11 +97,14 @@ type CPUInfo struct {
...
 
+	// ClusterID is the cluster ID, which sits between Socket and Core on some architectures (e.g. ARM)
+	ClusterID int `json:"clusterID"`
+
 	// NUMANodeID is the NUMA node ID, unique within each SocketID
 	NUMANodeID int `json:"numaNodeID"`
 
 	// NUMA Node Affinity Mask
-	NumaNodeAffinityMask string `json:"numaNodeAffinityMask"`
+	NumaNodeAffinityMask cpuset.CPUSet `json:"numaNodeAffinityMask"`
```
ffromani: A pet peeve of mine is to stick with "cpuset" terminology and not let the "mask" terminology creep in. But this is debatable in itself and, arguably, a pet peeve so I had to write down but I'm not too attached to it. I guess the main argument here is that using the two terms interchangeably leads to some confusion, so I'd rather stick with a single term for the concept (preferably "cpuset" :) )

ffromani (commented):
Besides a terminology thingy which we can fix later, LGTM
This is a nice improvement in how we extract the topology data from the system I'd merge anyway even without arm support.

ffromani:
/approve
/lgtm

@pravk03 your call if you want to address the terminology item here, later (or at all?). Unhold when you prefer.
I'm fine merging this PR first and enable ARM in CI later: I'm fine merging because the x86 improvements, but to claim ARM support we need to have CI, eventually.

pkg/cpuinfo/cpuinfo.go
AutuSnow: +1，It feels strange

pravk03: Makes sense. Renamed.

Issue #108, opened 2026-04-03
Verify by creating hooks in the driver's container as the last line of defense


AutuSnow:
## Problem
When a pod is created with `dra.cpu` resource claims, there's currently no validation to ensure the pod's CPU requests match the claim allocation. This can lead to mismatched configurations that are only caught at container startup.
## Proposal
Add validation in the driver's `NodeStageVolume` callback to check that:
1. For non-PLR mode: sum of container CPU requests == sum of claim CPUs
2. For PLR mode: pod-level CPU requests == sum of claim CPUs
This serves as the final line of defense when:
- The admission webhook is not deployed
- Claims are created asynchronously with pods
- Other validation layers fail
## Key considerations
- **Shared claims**: Handle `allocation.shared` list correctly for pods using sharePolicy: "WhenNotScheduled"
- **PLR compatibility**: Check if `PodLevelResources` feature gate is enabled and adapt validation accordingly
- **Error handling**: Return clear error messages when validation fails
## Related
- PR #61: Admission controller (webhook-based validation)
- KEP-5517: Pod Level Resources

AutuSnow:
current code validation PR was closed because KEP-5517 changed the direction; revisit only after native resource accounting semantics settle.

PR #110, opened 2026-04-06
add CPU request validation in NRI CreateContainer hook

AutuSnow:
Validation rules:
1. If container CPU request is specified, it must exactly match claim allocation
2. Pod-level resources validation (PLR) is a placeholder for future implementation

fixs:https://github.com/kubernetes-sigs/dra-driver-cpu/issues/108

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -77,8 +77,8 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 	return nil, nil
 }
 
-func parseDRAEnvToClaimAllocations(envs []string) (map[types.UID]cpuset.CPUSet, error) {
-	allocations := make(map[types.UID]cpuset.CPUSet)
+func parseDRAEnvToClaimAllocations(envs []string) (ClaimAllocations, error) {
+	allocations := make(ClaimAllocations)
 	for _, env := range envs {
 		if !strings.HasPrefix(env, cdiEnvVarPrefix) {
 			continue
@@ -150,6 +150,12 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
 		klog.Infof("No guaranteed CPUs found in DRA env for pod %s/%s container %s. Using shared CPUs %s", pod.Namespace, pod.Name, ctr.Name, sharedCPUs.String())
 		adjust.SetLinuxCPUSetCPUs(sharedCPUs.String())
 	} else {
+		if err := claimAllocations.Validate(pod, ctr); err != nil {
+			klog.Errorf("CPU request validation failed for pod %s/%s container %s: %v",
+				pod.Namespace, pod.Name, ctr.Name, err)
+			return nil, nil, fmt.Errorf("CPU request validation failed: %w", err)
+		}
+
 		guaranteedCPUs := cpuset.New()
 		claimUIDs := []types.UID{}
 		for uid, cpus := range claimAllocations {
--- a/pkg/driver/validation.go
+++ b/pkg/driver/validation.go
@@ -1,0 +1,78 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	"fmt"
+
+	"github.com/containerd/nri/pkg/api"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/cpuset"
+)
+
+const (
+	// minCPUShares is the Kubernetes minimum CPU shares for best-effort containers.
+	// Kubernetes sets this value for containers without CPU requests.
+	minCPUShares = uint64(2)
+
+	// cpuRequestTolerance allows for minor floating-point precision differences
+	// when comparing CPU requests (in cores) to claim allocations (in integer CPUs).
+	// Set to 0.01 to allow ~10ms difference per CPU core.
+	cpuRequestTolerance = 0.01
+)
+
+// ClaimAllocations represents CPU allocations from DRA claims.
+type ClaimAllocations map[types.UID]cpuset.CPUSet
+
+// TotalCPUs returns the total number of CPUs across all claims.
+func (ca ClaimAllocations) TotalCPUs() int {
+	total := 0
+	for _, cpus := range ca {
+		total += cpus.Size()
+	}
+	return total
+}
+
+// Validate checks that container CPU requests match the claim allocations.
+//
+// Validation behavior:
+//   - When container CPU shares > minCPUShares (2): validates that the request matches claim allocation
+//   - When container CPU shares <= minCPUShares or not set: validation passes (allows best-effort containers
+//     and Pod Level Resources scenarios where container-level requests are not specified)
+//
+// This lightweight validation serves as a final defense against common misconfigurations.
+func (ca ClaimAllocations) Validate(pod *api.PodSandbox, ctr *api.Container) error {
+	totalClaimCPUs := ca.TotalCPUs()
+
+	if ctr.Linux != nil && ctr.Linux.Resources != nil && ctr.Linux.Resources.Cpu != nil {
+		if ctr.Linux.Resources.Cpu.Shares != nil && ctr.Linux.Resources.Cpu.Shares.Value > minCPUShares {
+			containerCPUShares := ctr.Linux.Resources.Cpu.Shares.Value
+			containerCPURequest := float64(containerCPUShares) / 1024.0
+
+			diff := containerCPURequest - float64(totalClaimCPUs)
+			if diff < -cpuRequestTolerance || diff > cpuRequestTolerance {
+				return fmt.Errorf(
+					"container %s CPU request (%.2f cores, %d shares) does not match claim allocation (%d CPUs): "+
+						"when using DRA CPU claims, container CPU requests must exactly equal the claim size",
+					ctr.Name, containerCPURequest, containerCPUShares, totalClaimCPUs,
+				)
+			}
+		}
+	}
+
+	return nil
+}
--- a/pkg/driver/validation_test.go
+++ b/pkg/driver/validation_test.go
@@ -1,0 +1,442 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/containerd/nri/pkg/api"
+	"k8s.io/utils/cpuset"
+)
+
+func TestClaimAllocations_Validate(t *testing.T) {
+	tests := []struct {
+		name          string
+		pod           *api.PodSandbox
+		container     *api.Container
+		claims        ClaimAllocations
+		expectError   bool
+		errorContains string
+	}{
+		{
+			name: "matching container request - 10 CPUs",
+			pod: &api.PodSandbox{
+				Name:      "test-pod",
+				Namespace: "default",
+			},
+			container: &api.Container{
+				Name: "test-container",
+				Linux: &api.LinuxContainer{
+					Resources: &api.LinuxResources{
+						Cpu: &api.LinuxCPU{
+							Shares: &api.OptionalUInt64{Value: 10240}, // 10 CPUs * 1024
+						},
+					},
+				},
+			},
+			claims: ClaimAllocations{
+				"claim-1": cpuset.New(0, 1, 2, 3, 4, 5, 6, 7, 8, 9), // 10 CPUs
+			},
+			expectError: false,
+		},
+		{
+			name: "matching container request - 4 CPUs",
+			pod: &api.PodSandbox{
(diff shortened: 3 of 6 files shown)


pkg/driver/validation.go
```diff
@@ -0,0 +1,75 @@
+/*
+Copyright The Kubernetes Authors.
```
ffromani, line 2: do we want to have a `pkg/validate` or `pkg/driver/validate` (sub)package, and make this API public?
perhaps it is time to start our own `internal` hierarchy?

I'm actually asking, I don't have strong objections keeping this code here, except the fact the validation function is **correctly** private and we should try to not test directly private functions.

pkg/driver/nri_hooks.go
```diff
@@ -150,6 +150,13 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
 		klog.Infof("No guaranteed CPUs found in DRA env for pod %s/%s container %s. Using shared CPUs %s", pod.Namespace, pod.Name, ctr.Name, sharedCPUs.String())
 		adjust.SetLinuxCPUSetCPUs(sharedCPUs.String())
 	} else {
+		// Validate CPU requests match claim allocations
```
ffromani: the comment reiterate in english what the pretty explicit code is down a line below, so I'd remove it or repurpose to explain the "why", if it deserves explanation at all

pkg/driver/validation.go
```diff
@@ -0,0 +1,75 @@
...
+) error {
+	totalClaimCPUs := 0
+	for _, cpus := range claimAllocations {
+		totalClaimCPUs += cpus.Size()
+	}
+
+	if ctr.Linux != nil && ctr.Linux.Resources != nil && ctr.Linux.Resources.Cpu != nil {
+		// minCPUShares is the Kubernetes minimum for best-effort containers (no CPU request).
+		// Shares == 2 means no explicit CPU request was set; skip validation in that case.
+		const minCPUShares = uint64(2)
```
ffromani: this should be a file-level constant, it is important enough.

pkg/driver/validation.go
```diff
@@ -0,0 +1,75 @@
...
+		if ctr.Linux.Resources.Cpu.Shares != nil && ctr.Linux.Resources.Cpu.Shares.Value > minCPUShares {
+			containerCPUShares := ctr.Linux.Resources.Cpu.Shares.Value
+			containerCPURequest := float64(containerCPUShares) / 1024.0
+
+			const tolerance = 0.01
```
ffromani: likewise. And how this value was computed? If it's just our first educated guess, fine, but let's explicitly document it.

pkg/driver/validation.go
```diff
@@ -0,0 +1,75 @@
...
+
+			klog.V(4).InfoS("pod has pod-level CPU request",
+				"namespace", pod.Namespace,
+				"pod", pod.Name,
+				"podLevelCPURequest", podLevelCPURequest,
+				"container", ctr.Name,
+				"claimCPUs", totalClaimCPUs,
+			)
+		}
+	}
```
ffromani: If this just logging? If so, why are we doing inside the validation? Should it be its own little function?

pkg/driver/validation_test.go
```diff
@@ -0,0 +1,286 @@
...
+				},
+			},
+			claims: map[types.UID]cpuset.CPUSet{
+				"claim-1": cpuset.New(0), // 1 CPU
+			},
+			expectError: false,
+		},
+	}
+
+	for _, tc := range tests {
```
ffromani, line 378: the code has tests which exercise significant differences, it would be nice to add tests which exercise the smallest difference and any possible edge cases (can't really think of anything atm, but I haven't tried hard enough yet).

ffromani (commented):
Thanks! I see the point here and I'm supportive of this change. Mostly improvement suggestions inline.

test/e2e/validation_test.go
```diff
@@ -0,0 +1,228 @@
...
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			ginkgo.By("waiting for pod to fail with CreateContainerError")
+			gomega.Eventually(ctx, func(ctx context.Context) (*v1.Pod, error) {
+				return fxt.K8SClientset.CoreV1().Pods(fxt.Namespace.Name).Get(ctx, pod.Name, metav1.GetOptions{})
+			}).
+				WithTimeout(2*time.Minute).
+				WithPolling(5*time.Second).
+				Should(BeFailedToCreate(fxt), "pod should fail to create container")
+		})
```
ffromani, line 117: the test LGTM, but I wonder if we have a documented API (**not** the reason/error message text format) to catch this specific rejection

test/e2e/validation_test.go
```diff
@@ -0,0 +1,228 @@
...
+			pod, err = fxt.K8SClientset.CoreV1().Pods(fxt.Namespace.Name).Create(ctx, pod, metav1.CreateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			ginkgo.By("waiting for pod to fail with CreateContainerError")
+			gomega.Eventually(ctx, func(ctx context.Context) (*v1.Pod, error) {
+				return fxt.K8SClientset.CoreV1().Pods(fxt.Namespace.Name).Get(ctx, pod.Name, metav1.GetOptions{})
+			}).
+				WithTimeout(2*time.Minute).
+				WithPolling(5*time.Second).
+				Should(BeFailedToCreate(fxt), "pod should fail to create container")
```
ffromani, line 143: ditto

test/e2e/validation_test.go
```diff
@@ -0,0 +1,228 @@
...
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			ginkgo.By("creating a Pod without CPU request")
+			pod := makePodWithClaim(fxt.Namespace.Name, "test-pod", claim.Name, nil, nil)
+			pod, err = fxt.K8SClientset.CoreV1().Pods(fxt.Namespace.Name).Create(ctx, pod, metav1.CreateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			ginkgo.By("waiting for pod to be running")
+			err = e2epod.WaitToBeRunning(ctx, fxt.K8SClientset, pod.Namespace, pod.Name)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred(), "pod should be running")
```
ffromani, line 167: I see why you added this but I'm thinking if this should actually fail - changing the current driver behavior

pkg/driver/validation.go
AutuSnow: Actually, I have been thinking about creating a new level of the project because the 0.1 version was not completed during the development phase, so I still added and modified files on the original level

ffromani (commented):
added missing notes about the e2e tests

pkg/driver/validation.go
AutuSnow: I think we should maintain the current structure because this is the first verification function, and it is uncertain how much verification logic will be added in the future. If more verifications (such as memory, device, etc.) are added in the future, they can be refactored as pkg/validation/packages (if shared among multiple packages)
Regarding the issue of testing private functions: I believe that for pure logic validation functions, unit testing private functions is reasonable because the validation logic is complex and requires detailed unit testing coverage. E2E testing already covers integration scenarios, and if only public API testing is used, it will make the test cases too complex

pkg/driver/validation.go
AutuSnow: Good question! The purpose of this code is to:
1. Current status: This is a placeholder used to record the existence of Pod Level Resources, in preparation for future PLR (Pod Level Resources) validation
2. Why is it inside the validation function: Because we have already accessed the resource information of pods and containers here to avoid repeated traversal
3. Why is it just a log: According to the PR description, PLR validation is "placeholder for future implementation"

pkg/driver/validation.go
ffromani: We may end up accepting testing of private function but as general rule I want to try hard _against_ testing private functions directly (yes, this means there's some tech debt to clear over time). Each time should be a documented exception, not a habit we gradually develop.
So let's think a bit harder indeed. Perhaps we can turn the return value of `parseDRAEnvToClaimAllocations` into a proper type and add a `Validate` method to it with the current logic within?

test/e2e/validation_test.go
AutuSnow: Thank you for your reminder. I understand your concerns, but although CreateContainerError is not formally documented, it is not a formally documented constant in the Kubernetes API. It is a string value dynamically set by kubelet at runtime. But in the Kubernetes ecosystem, it is a factual standard, which is the standard reason value used by kubelets when container creation fails. If a more robust solution is needed, specific error text in the Message field can be checked additionally

AutuSnow:
```
if strings.Contains(cntSt.State.Waiting.Message, "CPU request validation failed") {
               xxxxx
}
```

test/e2e/validation_test.go
AutuSnow: Thanks for the question about the "without CPU requests specified" test case.
After discussion with @pravk03   , we decided to keep the validation lightweight in the CreateContainer hook:
- Validate when CPU request is set: Enforce that container.resources.requests.cpu matches the claim allocation
- Skip validation when no CPU request: Allow best-effort containers with claims (shares=2) to pass through
- Rationale: Avoid making CreateContainer too heavy. Full PLR validation will be handled by the scheduler (KEP-5517). The driver serves as a "final line of defense" for the most common misconfiguration
(explicit request mismatch).
This approach balances validation coverage with performance. WDYT?

pkg/driver/validation.go
```diff
@@ -0,0 +1,86 @@
...
+	return total
+}
+
+// Validate checks that container CPU requests match the claim allocations.
+func (ca ClaimAllocations) Validate(pod *api.PodSandbox, ctr *api.Container) error {
+	totalClaimCPUs := ca.TotalCPUs()
+
+	if ctr.Linux != nil && ctr.Linux.Resources != nil && ctr.Linux.Resources.Cpu != nil {
+		if ctr.Linux.Resources.Cpu.Shares != nil && ctr.Linux.Resources.Cpu.Shares.Value > minCPUShares {
```
pravk03, line 62: Lets also document that if shares are not set at container level, validation will pass. This is a valid scenario when pod level resources are set.

pkg/driver/validation.go
```diff
@@ -0,0 +1,86 @@
...
+				return fmt.Errorf(
+					"container %s CPU request (%.2f cores, %d shares) does not match claim allocation (%d CPUs): "+
+						"when using DRA CPU claims, container CPU requests must exactly equal the claim size",
+					ctr.Name, containerCPURequest, containerCPUShares, totalClaimCPUs,
+				)
+			}
+		}
+	}
+
+	if pod.Linux != nil && pod.Linux.PodResources != nil && pod.Linux.PodResources.Cpu != nil {
```
pravk03: I think this is incorrect. From reviewing the [kubelet code](https://github.com/kubernetes/kubernetes/blob/462e759d1995c143fda094cd7f591b10fd8cdee6/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L163 ), this check doesn't strictly confirm that Pod Level Resources (PLR) are enabled. If PLR isn't specified, the kubelet defaults this value to the sum of container requests.

pravk03: I wonder if there is a way to distinguish explicit PLR set in the pod. If we can, we could add an additional validation step - fail the validation if neither a container-level request nor a pod-level resource is specified.

AutuSnow: From the NRI hook, pod.Linux.PodResources is always populated by kubelet — either from explicit PLR or as the sum of container requests (when PLR feature gate is disabled). There's no reliable way to distinguish the two at this layer. The current approach skips validation when container-level shares are not set (shares <= 2), which correctly handles both best-effort containers and PLR scenarios. Full PLR validation is deferred to the  scheduler per KEP-5517.Removed the incorrect PLR detection block accordingly.

pravk03:
I've been thinking a bit more about this. Since KEP-5517 is currently in alpha, I'm concerned that the new validation introduced in this PR might prevent experimentation when the alpha feature gate is enabled.

Initially, I was hoping we could include some validation for Pod Level Resources—specifically checking that the pod-level budget is at least equal to the sum of container-level standard requests + DRA claims. However, during the PR's implementation and review, it has become clear that we can't reliably perform these pod-level validations.

Given this, I'm wondering if we should hold off on adding this validation in code for now, and instead focus on improving our documentation around workload requirements (both with and without KEP-5517)?

@AutuSnow, I know we previously discussed having this validation and I was initially on board with it, but I hadn't fully thought through the KEP-5517 implications at the time. Sorry for the back and forth here!

I would love to hear your thoughts on this @AutuSnow and @ffromani.

/hold

ffromani:
> I've been thinking a bit more about this. Since KEP-5517 is currently in alpha, I'm concerned that the new validation introduced in this PR might prevent experimentation when the alpha feature gate is enabled.
> 
> Initially, I was hoping we could include some validation for Pod Level Resources—specifically checking that the pod-level budget is at least equal to the sum of container-level standard requests + DRA claims. However, during the PR's implementation and review, it has become clear that we can't reliably perform these pod-level validations.
> 
> Given this, I'm wondering if we should hold off on adding this validation in code for now, and instead focus on improving our documentation around workload requirements (both with and without KEP-5517)?
> 
> @AutuSnow, I know we previously discussed having this validation and I was initially on board with it, but I hadn't fully thought through the KEP-5517 implications at the time. Sorry for the back and forth here!
> 
> I would love to hear your thoughts on this @AutuSnow and @ffromani.
> 
> /hold

My very initial thought is that the conflict between validation and KEP-5517 largely depends on the version skew we allow and support. IOW, which versions of the driver is compatible with the kubernetes version?
If this validation would have been merged in time for 0.1.0, the merge process would have been much more straightforward I reckon, because the (implicit) pairing would have been kube 1.35 with the driver 0.1.0.

I think is worth clarifying the interactions we expect. We probably need just a simple table (e.g. kube <= 1.35 -> driver 0.1.0) and/or few versions range.

That said, I'll deep dive in the PR and provide more informed comments.

AutuSnow:
This question has been debated for several days. Can we add the -- enable cpu request validation flag (default true) to be disabled during the KEP-5517 experiment, so as to preserve the safety net of regular scenarios

AutuSnow:
Closing this PR for now.
﻿
After the  discussion, this validation no longer looks like the right direction to merge as-is. In particular, enforcing container CPU requests in the NRI CreateContainer path can conflict with the KEP-5517 / Pod Level Resources direction and may restrict experimentation while that work is still evolving.
﻿
The workload configuration requirements are now documented in README.md, including both Pod Level Resources and container-level resource mirroring. I think we should keep any future runtime-side validation as a separate design discussion, likely tracked from #108, instead of rebasing this older implementation.
﻿
Thanks for the reviews and the discussion here.

/close

PR #113, opened 2026-04-09
fix: set imagePullPolicy to IfNotPresent for local dev

fmuyassarov:
When building manifests with a locally overridden image (OVERRIDE_IMAGE=true), set image pull policy to `IfNotPresent` to avoid unnecessary pull of an image which is already loaded to the kind, otherwise Pods can't run as the image with `v*-dirty` tag doesn't exist on the upstream registry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,6 +173,7 @@ manifests: dist install-yq ## create the install manifest
 	@cd dist && cp -a ../manifests/base/*.part.yaml .
 ifeq ($(OVERRIDE_IMAGE),true)
 	@$(YQ) -i '.spec.template.spec.containers[0].image = "${IMAGE}"' dist/daemonset-dracpu.part.yaml
+	@$(YQ) -i '.spec.template.spec.containers[0].imagePullPolicy = "IfNotPresent"' dist/daemonset-dracpu.part.yaml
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -173,6 +173,7 @@ manifests: dist install-yq ## create the install manifest
 	@cd dist && cp -a ../manifests/base/*.part.yaml .
 ifeq ($(OVERRIDE_IMAGE),true)
 	@$(YQ) -i '.spec.template.spec.containers[0].image = "${IMAGE}"' dist/daemonset-dracpu.part.yaml
+	@$(YQ) -i '.spec.template.spec.containers[0].imagePullPolicy = "IfNotPresent"' dist/daemonset-dracpu.part.yaml
```
ffromani, line 176: can we do this conditionally only if OVERRIDE_IMAGE is true?

This is hacky, but until we overhaul our manifest processing (which we should anyway) I'd rather tread carefully

fmuyassarov: it is conditionally already
`ifeq ($(OVERRIDE_IMAGE),true)`

fmuyassarov: it is just hidden 2 lines above the change

ffromani: Sorry, my fault. I reviewed the change too hastily. LGTM then.

ffromani (commented):
/approve
/lgtm

Makefile
fmuyassarov: No worries :)

Issue #109, opened 2026-04-03
Fix race condition during runtime restarts


pravk03:
This came up during wg-device-management meeting. More details and a potential workaround in https://github.com/containerd/nri/issues/282

SergeyKanzhelev:
The only feasible option for CPU DRA driver based on a requirement to run it for ALL Pods is either to become a global NRI plugin or change NRI to allow NRI plugins to self-register as "global". Does it sound correct or any other options listed in the linked issue is feasible?

pravk03:
@SergeyKanzhelev Seems like that is the available option based on current proposals.

ffromani:
I agree this seems the only available option. In practice, I think the "fix" on our side is to have very loud and clear docs about this requirement.

related: #72 and #94

SergeyKanzhelev:
Is there anything NRI plugin can do when it starts and registers like analyzing currently running containers and validate that settings are correct. if no - have some warning message reported?

Also, does the driver support rolling upgrade so at least one instance of NRI plugin (old or new) is running at all times - and sometimes both are running?

pravk03:
NRI provides the [synchronize](https://github.com/kubernetes-sigs/dra-driver-cpu/blob/59a142f9d0592ec13eb798221266ed875f15d64c/pkg/driver/nri_hooks.go#L32) hook which runs when the plugin starts. We currently use this to recompute our internal state, but it is also possible to issue container updates within this hook. This can only be a recovery mechanism, as any containers created during the downtime would start with the wrong CPU mask.

ffromani:
> NRI provides the [synchronize](https://github.com/kubernetes-sigs/dra-driver-cpu/blob/59a142f9d0592ec13eb798221266ed875f15d64c/pkg/driver/nri_hooks.go#L32) hook which runs when the plugin starts. We currently use this to recompute our internal state, but it is also possible to issue container updates within this hook. This can only be a recovery mechanism, as any containers created during the downtime would start with the wrong CPU mask.

I'd be in favor of adding recovery here. That + logging (and metrics?) is the only thing which we can realistically do. A late recovery is better than keeping the system in broken state.

SergeyKanzhelev:
what about rolling update? Is it supported?

pravk03:
[...]

PR #115, opened 2026-04-10
Reconcile container CPU masks during NRI Synchronize

pravk03:
This is a solution for https://github.com/kubernetes-sigs/dra-driver-cpu/issues/109, although not perfect.

This ensures cgroup is correctly updated if the plugin crash or restarts and is not available during container creation. But while the NRI plugin is down, the containers created will not have their CPU affinity restricted and will have access to all online CPUs on the node (potentially overlapping with exclusive CPUs assigned to other pods with claims)

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -35,7 +35,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cpuAllocationStore := store.NewCPUAllocation(cp.cpuTopology, cp.reservedCPUs)
 	podConfigStore := store.NewPodConfig()
-	var contianerUpdates []*api.ContainerUpdate
+	var containerUpdates []*api.ContainerUpdate
 
 	logger := klog.FromContext(ctx)
 	for _, pod := range pods {
@@ -74,7 +74,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 					ContainerId: container.GetId(),
 				}
 				guaranteedUpdate.SetLinuxCPUSetCPUs(allGuaranteedCPUs.String())
-				contianerUpdates = append(contianerUpdates, guaranteedUpdate)
+				containerUpdates = append(containerUpdates, guaranteedUpdate)
 			}
 			podConfigStore.SetContainerState(types.UID(pod.GetUid()), state)
 		}
@@ -87,8 +87,8 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 	// or restarted and missed updating the cgroup settings.
 	// See: https://github.com/containerd/nri/issues/282
 	sharedContainerUpdates := cp.getSharedContainerUpdates(types.UID(""))
-	contianerUpdates = append(contianerUpdates, sharedContainerUpdates...)
-	return contianerUpdates, nil
+	containerUpdates = append(containerUpdates, sharedContainerUpdates...)
+	return containerUpdates, nil
 }
 
 func parseDRAEnvToClaimAllocations(envs []string) (map[types.UID]cpuset.CPUSet, error) {
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -35,6 +35,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cpuAllocationStore := store.NewCPUAllocation(cp.cpuTopology, cp.reservedCPUs)
 	podConfigStore := store.NewPodConfig()
+	var contianerUpdates []*api.ContainerUpdate
 
 	logger := klog.FromContext(ctx)
 	for _, pod := range pods {
@@ -67,6 +68,13 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				}
 				klog.Infof("Synchronize: Found guaranteed CPUs for pod %s/%s container %s with cpus: %v", pod.Namespace, pod.Name, container.Name, allGuaranteedCPUs.String())
 				state = store.NewContainerState(container.GetName(), containerUID, claimUIDs...)
+
+				// Reconcile guaranteed container CPU mask.
+				guaranteedUpdate := &api.ContainerUpdate{
+					ContainerId: container.GetId(),
+				}
+				guaranteedUpdate.SetLinuxCPUSetCPUs(allGuaranteedCPUs.String())
+				contianerUpdates = append(contianerUpdates, guaranteedUpdate)
 			}
 			podConfigStore.SetContainerState(types.UID(pod.GetUid()), state)
 		}
@@ -74,7 +82,13 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cp.podConfigStore = podConfigStore
 	cp.cpuAllocationStore = cpuAllocationStore
-	return nil, nil
+
+	// Reconcile container CPU masks to handle cases where the NRI plugin might have crashed
+	// or restarted and missed updating the cgroup settings.
+	// See: https://github.com/containerd/nri/issues/282
+	sharedContainerUpdates := cp.getSharedContainerUpdates(types.UID(""))
+	contianerUpdates = append(contianerUpdates, sharedContainerUpdates...)
+	return contianerUpdates, nil
 }
 
 func parseDRAEnvToClaimAllocations(envs []string) (map[types.UID]cpuset.CPUSet, error) {
--- a/test/e2e/nri_reconciliation_test.go
+++ b/test/e2e/nri_reconciliation_test.go
@@ -1,0 +1,262 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/discovery"
+	"github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/fixture"
+	"github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/node"
+	e2epod "github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/pod"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/cpuset"
+)
+
+const (
+	daemonSetNamespaceRule = "kube-system"
+	pollIntervalRule       = 2 * time.Second
+	pollTimeoutRule        = 2 * time.Minute
+)
+
+var _ = ginkgo.Describe("NRI Reconciliation on Restart", ginkgo.Serial, ginkgo.Ordered, ginkgo.ContinueOnFailure, func() {
+	var (
+		rootFxt           *fixture.Fixture
+		targetNode        *v1.Node
+		targetNodeCPUInfo discovery.DRACPUInfo
+		dracpuTesterImage string
+		allocatableCPUs   cpuset.CPUSet
+		reservedCPUs      cpuset.CPUSet
+		cpuDeviceMode     string
+		orgDaemonSet      *appsv1.DaemonSet
+	)
+
+	ginkgo.BeforeAll(func(ctx context.Context) {
+		// early cheap check before to create the Fixture, so we use GinkgoLogr directly
+		dracpuTesterImage = os.Getenv("DRACPU_E2E_TEST_IMAGE")
+		gomega.Expect(dracpuTesterImage).ToNot(gomega.BeEmpty(), "missing environment variable DRACPU_E2E_TEST_IMAGE")
+		ginkgo.GinkgoLogr.Info("discovery image", "pullSpec", dracpuTesterImage)
+
+		var err error
+		if reservedCPUVal := os.Getenv("DRACPU_E2E_RESERVED_CPUS"); len(reservedCPUVal) > 0 {
+			reservedCPUs, err = cpuset.Parse(reservedCPUVal)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			ginkgo.GinkgoLogr.Info("reserved CPUs", "value", reservedCPUs.String())
+		}
+
+		rootFxt, err = fixture.ForGinkgo()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot create fixture")
+		infraFxt := rootFxt.WithPrefix("infra")
+		gomega.Expect(infraFxt.Setup(ctx)).To(gomega.Succeed())
+		ginkgo.DeferCleanup(infraFxt.Teardown)
+
+		ginkgo.By("getting the daemonset configuration")
+		orgDaemonSet, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Get(ctx, "dracpu", metav1.GetOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot get dracpu daemonset")
+		gomega.Expect(orgDaemonSet.Spec.Template.Spec.Containers).ToNot(gomega.BeEmpty(), "no containers in dracpu daemonset")
+
+		if val, ok := findArgInContainer(&orgDaemonSet.Spec.Template.Spec.Containers[0], argCPUDeviceMode); ok {
+			cpuDeviceMode = val
+		}
+
+		// Find target node
+		if targetNodeName := os.Getenv("DRACPU_E2E_TARGET_NODE"); len(targetNodeName) > 0 {
+			targetNode, err = rootFxt.K8SClientset.CoreV1().Nodes().Get(ctx, targetNodeName, metav1.GetOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		} else {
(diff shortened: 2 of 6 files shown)


pkg/driver/nri_hooks.go
```diff
@@ -35,6 +35,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cpuAllocationStore := store.NewCPUAllocation(cp.cpuTopology, cp.reservedCPUs)
 	podConfigStore := store.NewPodConfig()
+	var contianerUpdates []*api.ContainerUpdate
```
ffromani: typo: containerUpdates

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,263 @@
...
+		ginkgo.By("getting the daemonset configuration")
+		orgDaemonSet, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Get(ctx, "dracpu", metav1.GetOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot get dracpu daemonset")
+		gomega.Expect(orgDaemonSet.Spec.Template.Spec.Containers).ToNot(gomega.BeEmpty(), "no containers in dracpu daemonset")
+
+		for _, arg := range orgDaemonSet.Spec.Template.Spec.Containers[0].Args {
+			if val, ok := parseCPUDeviceModeArg(arg); ok {
+				cpuDeviceMode = val
+			}
+		}
```
ffromani: (unrelated to this PR) this is becoming a pattern so we would need an helper soon enough.

pkg/driver/nri_hooks.go
```diff
@@ -35,6 +35,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
...
```
fmuyassarov:
```suggestion
	var containerUpdates []*api.ContainerUpdate
```
nit

pkg/driver/nri_hooks.go
```diff
@@ -67,14 +68,27 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				}
 				klog.Infof("Synchronize: Found guaranteed CPUs for pod %s/%s container %s with cpus: %v", pod.Namespace, pod.Name, container.Name, allGuaranteedCPUs.String())
 				state = store.NewContainerState(container.GetName(), containerUID, claimUIDs...)
+
+				// Reconcile guaranteed container CPU mask.
+				guaranteedUpdate := &api.ContainerUpdate{
+					ContainerId: container.GetId(),
+				}
+				guaranteedUpdate.SetLinuxCPUSetCPUs(allGuaranteedCPUs.String())
+				contianerUpdates = append(contianerUpdates, guaranteedUpdate)
```
fmuyassarov: ditto

ffromani:
arm64 failures are a known issue: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/120

pkg/driver/nri_hooks.go
pravk03: Done. Thanks.

ffromani (commented):
partial review, still need to catch up with the e2e test. Everything else LGTM besides the cpumask/cpuset terminology. Everything else equal, I'd prefer "cpuset" over "cpumask"

pkg/driver/nri_hooks.go
pravk03: fixed

test/e2e/nri_reconciliation_test.go
pravk03: Good point. Added a helper.

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,261 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/cpuset"
+)
+
+const (
+	daemonSetNamespaceRule = "kube-system"
+	daemonSetLabelRule     = "app=dracpu"
```
AutuSnow: daemonSetLabelRule isn't referenced anywhere in this file (listDriverPods in healthz_test.go has its own selector). Please drop it, or use it via a shared helper.

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,261 @@
...
+				return
+			}
+			ds.Spec = orgDaemonSet.Spec
+			_, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Update(ctx, ds, metav1.UpdateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+
+		gomega.Eventually(func(g gomega.Gomega) {
+			// Modify DaemonSet to exclude target node
+			ds := orgDaemonSet.DeepCopy()
```
AutuSnow: `orgDaemonSet` is captured once in BeforeAll, so every retry Updates with the same resourceVersion. On the first conflict the retries are guaranteed to keep failing until the poll timeout. Keep `orgDaemonSet` only as the snapshot used by DeferCleanup to restore the original Spec.

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,261 @@
...
+						MatchExpressions: []v1.NodeSelectorRequirement{req},
+					},
+				}
+			} else {
+				for i := range terms {
+					terms[i].MatchExpressions = append(terms[i].MatchExpressions, req)
+				}
+			}
+			_, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Update(ctx, ds, metav1.UpdateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
AutuSnow: Use the scoped `g.Expect(...)` here (and on the affinity-nil asserts above if you add any). As written, a transient API error or a 409 Conflict fails the whole test instead of letting `Eventually` retry, which defeats the purpose of wrapping the Update in Eventually.

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,261 @@
...
+		alloc2 := getTesterPodCPUAllocation(fxt.K8SClientset, ctx, createdPod2)
+		fxt.Log.Info("Pod 2 CPU allocation (without NRI)", "cpuAssigned", alloc2.CPUAssigned.String())
+		// Since NRI is down, pod2 is not restricted to shared pool CPUs and gets all online CPUs.
+		gomega.Expect(alloc2.CPUAffinity).To(gomega.Equal(allocatableCPUs), "Pod 2 CPU mask not equal to all CPUs")
+
+		ginkgo.By("Bringing up the cpu dra driver on target node")
+		// Restore original DaemonSet
+		ds, err := rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Get(ctx, "dracpu", metav1.GetOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		ds.Spec = orgDaemonSet.Spec
```
AutuSnow, line 224: nit:  This inline restore duplicates the earlier `DeferCleanup`. Since DeferCleanup will run on both success and failure paths, the inline  block is redundant. Suggest removing the inline restore and relying on the DeferCleanup for a single source of truth.

AutuSnow (commented):
Thank you for solving it. It looks like a great idea. I left some comments in the code for unit testing

test/e2e/nri_reconciliation_test.go
pravk03: removed.

test/e2e/nri_reconciliation_test.go
pravk03: Done. Good catch. I updated the loop to fetch the latest copy of the DaemonSet.

test/e2e/nri_reconciliation_test.go
pravk03: Done. Excellent point.

test/e2e/nri_reconciliation_test.go
pravk03: I think we still need the restore here to reinstall the driver ds pod since we run cleanup only at the end.

AutuSnow: Thank you for your rebuttal, which allowed me to reorganize the code and discover that my original idea was wrong

AutuSnow:
/lgtm
/hold
Need @ffromani to evaluate e2e testing

pkg/driver/nri_hooks.go
```diff
@@ -67,14 +68,27 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				}
 				klog.Infof("Synchronize: Found guaranteed CPUs for pod %s/%s container %s with cpus: %v", pod.Namespace, pod.Name, container.Name, allGuaranteedCPUs.String())
 				state = store.NewContainerState(container.GetName(), containerUID, claimUIDs...)
+
+				// Reconcile guaranteed container CPU mask.
```
ffromani, line 72: nit: CPUSet (or CPU set)

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/cpuset"
+)
+
+const (
+	daemonSetNamespaceRule = "kube-system"
+	pollIntervalRule       = 2 * time.Second
```
ffromani, line 41: what does 'Rule' mean here?

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+		infoPod = e2epod.PinToNode(infoPod, targetNode.Name)
+		infoPod, err = e2epod.RunToCompletion(ctx, infraFxt.K8SClientset, infoPod)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		data, err := e2epod.GetLogs(infraFxt.K8SClientset, ctx, infoPod.Namespace, infoPod.Name, infoPod.Spec.Containers[0].Name)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		gomega.Expect(json.Unmarshal([]byte(data), &targetNodeCPUInfo)).To(gomega.Succeed())
+		allocatableCPUs = makeCPUSetFromDiscoveredCPUInfo(targetNodeCPUInfo)
+	})
+
+	ginkgo.It("should recover shared pool mask and preserve exclusive mask after restart", func(ctx context.Context) {
```
ffromani, line 114: nit: cpuset

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+			},
+		}
+		createdClaimTemplate, err := fxt.K8SClientset.ResourceV1().ResourceClaimTemplates(fxt.Namespace.Name).Create(ctx, &claimTemplate, metav1.CreateOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		pod1 := makeTesterPodWithExclusiveCPUClaim(fxt.Namespace.Name, dracpuTesterImage, createdClaimTemplate.Name, 2, targetNode.Name)
+		createdPod1, err := e2epod.CreateSync(ctx, fxt.K8SClientset, pod1)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		ginkgo.By("Verifying Pod 1 has correct exclusive CPU mask")
```
ffromani, line 135: nit: cpuset

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		rootFxt, err = fixture.ForGinkgo()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot create fixture")
+		infraFxt := rootFxt.WithPrefix("infra")
+		gomega.Expect(infraFxt.Setup(ctx)).To(gomega.Succeed())
+		ginkgo.DeferCleanup(infraFxt.Teardown)
+
+		ginkgo.By("getting the daemonset configuration")
+		orgDaemonSet, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Get(ctx, "dracpu", metav1.GetOptions{})
```
ffromani, line 77: nit: `origDaemonSet` is probably more explicit

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+		// Defer restoration of DaemonSet
+		ginkgo.DeferCleanup(func(ctx context.Context) {
+			ginkgo.By("Restoring NRI plugin DaemonSet")
+			ds, err := rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Get(ctx, "dracpu", metav1.GetOptions{})
+			if err != nil {
+				return
+			}
+			ds.Spec = orgDaemonSet.Spec
+			_, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Update(ctx, ds, metav1.UpdateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
ffromani, line 151: this should be done in a retry loop (just an `Eventually` block) to prevent/greatly reduce the chance of flakes

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+					},
+				}
+			} else {
+				for i := range terms {
+					terms[i].MatchExpressions = append(terms[i].MatchExpressions, req)
+				}
+			}
+			_, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Update(ctx, ds, metav1.UpdateOptions{})
+			g.Expect(err).ToNot(gomega.HaveOccurred())
+		}, pollTimeoutRule, pollIntervalRule).Should(gomega.Succeed(), "failed to update DaemonSet affinity")
```
ffromani, line 187: we should use `WithTimeout` and `WithPolling` in our code: same result, more readability

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+			g.Expect(err).NotTo(gomega.HaveOccurred())
+
+			terminated := true
+			for _, p := range pods {
+				if p.Spec.NodeName == targetNode.Name && p.Status.Phase != v1.PodFailed && p.Status.Phase != v1.PodSucceeded {
+					terminated = false
+					break
+				}
+			}
+			g.Expect(terminated).To(gomega.BeTrue(), "Pod on target node is still running")
```
ffromani, line 201: can we expect just the pod to be IsNotFound and wait till it happens?

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+		}, pollTimeoutRule, pollIntervalRule).Should(gomega.Succeed(), "timed out waiting for pod to terminate")
```
ffromani, line 202: ditto about WithPolling/WithTimeout

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+		ginkgo.By("Verifying Pod 1 still has correct exclusive CPU mask while NRI is down")
```
ffromani, line 204: ditto: cpuset

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+		// Since NRI is down, pod2 is not restricted to shared pool CPUs and gets all online CPUs.
+		gomega.Expect(alloc2.CPUAffinity).To(gomega.Equal(allocatableCPUs), "Pod 2 CPU mask not equal to all CPUs")
+
+		ginkgo.By("Bringing up the cpu dra driver on target node")
+		// Restore original DaemonSet
+		ds, err := rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Get(ctx, "dracpu", metav1.GetOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		ds.Spec = orgDaemonSet.Spec
+		_, err = rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespaceRule).Update(ctx, ds, metav1.UpdateOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
ffromani, line 226: I'd also wrap this logic in a retry loop, be it just `Eventually`

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+							ready = true
+							break
+						}
+					}
+					if ready {
+						break
+					}
+				}
+			}
+			g.Expect(ready).To(gomega.BeTrue(), "Pod on target node is not ready")
```
ffromani, line 247: this can be a custom gomega matcher (in another PR)

test/e2e/nri_reconciliation_test.go
```diff
@@ -0,0 +1,262 @@
...
+		alloc1Up := getTesterPodCPUAllocation(fxt.K8SClientset, ctx, createdPod1)
+		gomega.Expect(alloc1Up.CPUAssigned).To(gomega.Equal(exclusiveCPUs), "Pod 1 CPU mask changed after NRI restarted")
+
+		ginkgo.By("Verifying Pod 2 CPU mask IS restricted to shared pool (excludes Pod 1 CPUs)")
+		gomega.Eventually(func(g gomega.Gomega) {
+			alloc2Updated := getTesterPodCPUAllocation(fxt.K8SClientset, ctx, createdPod2)
+			fxt.Log.Info("Pod 2 CPU allocation (after NRI recovery)", "cpuAssigned", alloc2Updated.CPUAssigned.String())
+			// pod2 must NOT contain the exclusive CPUs now
+			g.Expect(alloc2Updated.CPUAssigned.Intersection(exclusiveCPUs).IsEmpty()).To(gomega.BeTrue(), "Pod 2 still has access to exclusive CPUs")
+		}, pollTimeoutRule, pollIntervalRule).Should(gomega.Succeed(), "timed out waiting for Pod 2 CPU mask update")
```
ffromani, line 260: ditto about WhitPolling/WithTimeout

ffromani (commented):
Thanks for the PR. There's a fair amount of e2e improvements we can land, none of them blocking though. It's perhaps better to merge and do sweeping changes to the e2e tests.

ffromani:
/hold cancel

even though there are quite some improvement opportunities, I slightly prefer to make progress. We can take a look to the e2e status with a focused PR and that's probably better.

ffromani:
> /hold cancel
> 
> even though there are quite some improvement opportunities, I slightly prefer to make progress. We can take a look to the e2e status with a focused PR and that's probably better.

https://github.com/kubernetes-sigs/dra-driver-cpu/issues/132

PR #116, opened 2026-04-10
Release exclusive CPUs in NRI StopContainer 

pravk03:
Previously, the driver relied on the DRA `UnprepareResourceClaims` hook to release CPUs back to the shared pool. However, NRI only allows pushing CPU mask updates to other containers during container lifecycle events (like `StopContainer`). If we wait until `UnprepareResourceClaims` to release the CPUs, we miss the opportunity to update the shared pool of existing containers via NRI, leaving them on a restricted pool until a new container event occurs. This PR moves the release of exclusive CPUs into the NRI `StopContainer` hook.

The PR also improves E2E tests by removing dependency on container restarts. Instead, we record the CPU Mask every 5 seconds and read the last log line to get the last mask. Container restarts were resulting in presubmit test timeout as it results in kubelet CrashLoopBackOff.


AutuSnow:
Look Good！Thanks！
/lgtm

PR #117, opened 2026-04-10
fix: a few scattered little fixes

gauravgahlot:


pkg/cpuinfo/cpuinfo_test.go
```diff
@@ -196,7 +198,7 @@ func createFakeCPUTopology(t *testing.T, dir string, topo fakeCPUTopology) {
 			t.Fatal(err)
 		}
 		socketID := i / (coresPerSocket * topo.cpusPerCore)
-		if err := os.WriteFile(filepath.Join(topologyDir, "physical_package_id"), []byte(fmt.Sprintf("%d\n", socketID)), 0600); err != nil {
+		if err := os.WriteFile(filepath.Join(topologyDir, "physical_package_id"), fmt.Appendf(nil, "%d\n", socketID), 0600); err != nil {
```
ffromani: not sure this is actually an improvement though

pkg/driver/dra_hooks_test.go
```diff
@@ -137,7 +137,7 @@ var (
 	mockCPUInfos_DualSocket_120CPUsPerSocket_HT = func() []cpuinfo.CPUInfo {
 		var infos []cpuinfo.CPUInfo
 		numCores := 120
-		for socketID := 0; socketID < 2; socketID++ {
+		for socketID := range 2 {
```
ffromani, line 140: this should be unrolled, perhaps with a helper function

ffromani (commented):
Thanks for the contribution and welcome to the project!

The first commit message is a bit terse. We can use a more expressive commit message for the first commit, and perhaps splitting it further.

pkg/cpuinfo/cpuinfo_test.go
gauravgahlot: The original code first creates a string (`fmt.Sprintf`) and then converts it into a `[]byte`. 

`fmt.Appendf` formats directly into a `[]byte`, and skips the intermediate string entirely.

pkg/driver/dra_hooks_test.go
gauravgahlot: @ffromani Do you mean something like?

```go
...
mockCPUInfos_DualSocket_120CPUsPerSocket_HT = getCPUInfo(120, 2)

func getCPUInfo(numCores int, numSockets int) []cpuinfo.CPUInfo {
	var infos []cpuinfo.CPUInfo

	for socketID := range numSockets {
		for coreID := 0; coreID < numCores/2; coreID++ {
			baseCpuID := socketID * numCores
			infos = append(infos, cpuinfo.CPUInfo{
				CpuID:        baseCpuID + coreID*2,
				CoreID:       coreID,
				SocketID:     socketID,
				NUMANodeID:   socketID,
				CoreType:     cpuinfo.CoreTypePerformance,
				SiblingCpuID: baseCpuID + coreID*2 + 1,
			})

			// create the second logical CPU (thread 1) on the same core
			infos = append(infos, cpuinfo.CPUInfo{
				CpuID:        baseCpuID + coreID*2 + 1,
				CoreID:       coreID,
				SocketID:     socketID,
				NUMANodeID:   socketID,
				CoreType:     cpuinfo.CoreTypePerformance,
				SiblingCpuID: baseCpuID + coreID*2,
			})
		}
	}

	return infos
}
...
```

pkg/cpuinfo/cpuinfo_test.go
pohly: And that is faster? Have you measured it? My expectation is that this is a NOP compile-time type cast.

Last but not least, does it matter at all here?

I'm with @ffromani here: the purpose of the new code is not at all clear, let's not do this.

pkg/driver/dra_hooks_test.go
ffromani: @gauravgahlot yep, this is pretty much what I had in mind

pohly: `for coreID := 0; coreID < numCores/2; coreID++` could be replaced with `for coreID := range numCores/2`, right?

pkg/cpuinfo/cpuinfo_test.go
gauravgahlot: You are right, it does not matter here and it's not clear. I have reverted the change.

pkg/driver/dra_hooks_test.go
```diff
@@ -128,40 +128,13 @@ var (
...
-					SocketID:     socketID,
-					NUMANodeID:   socketID,
-					CoreType:     cpuinfo.CoreTypePerformance,
-					SiblingCPUID: baseCpuID + coreID*2,
-				})
-			}
-		}
-		return infos
-	}()
+	mockCPUInfos_DualSocket_120CPUsPerSocket_HT = getCPUInfo(120, 2)
```
ffromani: mh, not sure this is an improvement. Having plain flat code in tests is a feature, unlike production code on which we want to use helpers and factorization quite aggressively.

A way forward could be to remove this specific change and we can proceed with the rest.

ffromani (commented):
Thanks, it could be a good idea to add a script (or perhaps even a CI action?) to modernize the codebase, perhaps piggybacking on the existing linter code.

ffromani (commented):
/approve
/lgtm

let's start small indeed. Thanks for your first contribution!

PR #121, opened 2026-04-15
Implement HandleError method for DRA plugin error handling

Karthik-K-N:
This PR tries to address the TODO comment. Followed similar approach like other DRA drivers.

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -28,6 +28,7 @@ import (
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/dynamic-resource-allocation/kubeletplugin"
 	"k8s.io/dynamic-resource-allocation/resourceslice"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


fmuyassarov (approved):
/lgtm
Thank you!
/cc @ffromani

pkg/driver/dra_hooks.go
```diff
@@ -448,7 +449,10 @@ func (cp *CPUDriver) unprepareResourceClaim(_ context.Context, claim kubeletplug
...
 
-func (cp *CPUDriver) HandleError(_ context.Context, err error, msg string) {
-	// TODO: Implement this function
-	klog.Error("HandleError error:", err, "msg:", msg)
+// HandleError is called by the kubelet plugin framework when an error occurs in the background,
+// for example while publishing ResourceSlices.
+func (cp *CPUDriver) HandleError(ctx context.Context, err error, msg string) {
+	// For now we just follow the advice documented in the DRAPlugin API docs.
+	// See: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/runtime#HandleErrorWithContext
+	runtime.HandleErrorWithContext(ctx, err, msg)
```
ffromani, line 458: Reading the DRAPlugin API ref, I'd go furter and actually exit if the error is not recoverable. This behavior (abort on unrecoverable errors) can be surprising and can take some iteration to tune, so the earlier we introduce explicit  and loud failure modes, the better.
This compounds to the fact we'are very early in the project maturity, so users will (or should) expect some rough edges.

ffromani (commented):
Thanks for the contribution and welcome to the project!

pkg/driver/dra_hooks.go
Karthik-K-N: Updated as suggested. Please take a look. Thank you.

ffromani (commented):
/approve
/lgtm

thanks!

fmuyassarov:
seems to be a flake of kind...
/test pull-dra-driver-cpu-e2e-device-mode-individual-amd64

Issue #36, opened 2025-12-10
absorb the memory allocation feature


ffromani:
There's growing mindshare in the node community that it's much better, as in more efficient, more scalable and yielding better allocations, to have a single driver for both CPU and memory management. This issue is meant to track this integration work.

- [ ]  merge functionality from https://github.com/ffromani/dra-driver-memory once the latter matures enough
- [ ] integrate CPU and memory resources in a single coherent way
- [ ] rename this project, or possibly fork this project, as `dra-driver-native-resources` or `dra-driver-core-resources` or so. Or maybe keep it named `CPU`? we need to decide

All this work is meant to begin well after 0.1.0 (https://github.com/kubernetes-sigs/dra-driver-cpu/issues/34) is out. We tentatively aim for 0.2.0 or 0.3.0.

PR #126, opened 2026-04-17
chore: drop the initContainer

### LINKED ISSUE #94: Remove the init container used for NRI enablement
NRI has been supported and enabled by default for quite some time (since CRI-O 1.30 and containerd 2.0). Given this, it should be safe to remove the `initContainer` currently used to enable NRI at runtime in all installations. At the moment, the `initContainer` only enables NRI for containerd. This creates a problem for CRI-O based setups, where the init container will fail and does not provide any benefit. Additionally, I tend to believe that enabling NRI is typically the job of admin, rather than something that should be handled within the app running on k8s. Based on these points, I would suggest that we drop the `initContainer` to simplify install.yaml.

fmuyassarov:
It’s been a while since runtimes CRI-O (1.30+) and containerd (2.0+) started enabling NRI by default. Cases where it is not enabled seem rare, so it may not be worth requiring extra config changes for everyone during install. Right now, there is also no way to opt out of this behavior. That said, I added an option in the [Helm chart](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83) to enable NRI for containerd if a user explicitly wants it (although, I would prefer to drop there too). Also, it feels like enabling NRI should be handled by the cluster admin. Runtime config is more of a cluster concern and not something an app should enforce.

Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/94

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,61 @@ This discrepancy is a known issue being addressed by [KEP-5517: Native Resource 
 
 **1-to-1 Claim to Container:** This driver enforces that a specific CPU `ResourceClaim` can only be used by *one* container within or across pods. See [Sharing resource claims](#sharing-resource-claims).
 
+## Prerequisites
+
+The driver relies on [NRI (Node Resource Interface)](https://github.com/containerd/nri) to pin containers to their
+allocated CPUs, and on [CDI (Container Device Interface)](https://github.com/cncf-tags/container-device-interface) to
+inject the allocated cpuset into the container environment.
+
+### Minimum Runtime Requirements
+
+Both NRI and CDI are enabled by default in modern container runtimes:
+
+| Runtime    | NRI enabled by default | CDI enabled by default |
+| ---------- | ---------------------- | ---------------------- |
+| containerd | 2.0+                   | 2.0+                   |
+| CRI-O      | 1.30+                  | always                 |
+
+Both runtimes also ship with the following CDI spec directories configured by default:
+
+```toml
+cdi_spec_dirs = ["/etc/cdi", "/var/run/cdi"]
+```
+
+No manual runtime configuration is needed if you are running one of the versions above or newer.
+
+### Manual Configuration for Older Runtimes
+
+If you are running an older version of containerd (pre-2.0), you need to manually enable CDI and NRI in the containerd
+configuration (typically `/etc/containerd/config.toml`) and restart containerd.
+
+#### Enable CDI
+
+```toml
+[plugins."io.containerd.grpc.v1.cri"]
+  enable_cdi = true
+  cdi_spec_dirs = ["/etc/cdi", "/var/run/cdi"]
+```
+
+#### Enable NRI
+
+```toml
+[plugins."io.containerd.nri.v1.nri"]
+  disable = false
+  disable_connections = false
+  plugin_config_path = "/etc/nri/conf.d"
+  plugin_path = "/opt/nri/plugins"
+  plugin_registration_timeout = "5s"
+  plugin_request_timeout = "5s"
+  socket_path = "/var/run/nri/nri.sock"
+```
+
+After editing the config, restart containerd:
+
+```bash
+systemctl restart containerd
+```
+
 ## Getting Started
 
 ### Installation
--- a/manifests/base/daemonset-dracpu.part.yaml
+++ b/manifests/base/daemonset-dracpu.part.yaml
@@ -24,58 +24,6 @@ spec:
       - operator: Exists
         effect: NoSchedule
       serviceAccountName: dracpu
-      hostPID: true
-      initContainers:
-      - name: enable-nri-and-cdi
-        image: busybox:stable
-        volumeMounts:
-        - mountPath: /etc
-          name: etc
-        securityContext:
-          privileged: true
-        resources:
-          requests:
-            cpu: "100m"
-            memory: "50Mi"
-        command:
-        - /bin/sh
-        - -c
-        - |
-          set -o errexit
-          set -o pipefail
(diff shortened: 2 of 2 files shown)


pravk03:
Thanks for looking into this. I have a question.

> It’s been a while since runtimes CRI-O (1.30+) and containerd (2.0+) started enabling NRI by default.

The init container also enabled CDI and configured `cdi_spec_dirs`.  Is that also enabled out of the box in newer versions ?


Should we also update the README with a "Prerequisites" section to explicitly state the minimum runtime requirements and the manual steps needed if users are on older versions ?. We could do this in a separate PR or with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83.

ffromani:
Generally speaking, I agree system configuration is not an app concern but a system concern.

I would like to keep our reference configuration _and_ an automated way to apply it (whatever it is) OR a configuration check mechanism (dra-driver-cpu option or subcommand, script...) for few releases more as documentation/aid to customers.

I'd be fine to NOT enable this even conditionally in our deployment mechanism, but we should have clear docs about the reference config (e.g. clear section in our readme or so).

I'm positive we can find a good balance during the review.

fmuyassarov:
Thanks for the reviews @pravk03, @ffromani. I will address both of yours reviews here.

Regarding the CDI bits, the init container is currently doing two things - enabling CDI and configuring the CDI spec dirs in the containerd config. For CDI enabling, CDI has been enabled by default starting from containerd v2.0 ([release notes](https://github.com/containerd/containerd/blob/main/docs/containerd-2.0.md#cdi-is-now-enabled-by-default)). For CRI-O I haven’t found a specific release note explicitly mentioning it but I can see that they clearly mention the default directories which gives me a signal that it is on by default.

Regarding CDI spec directories,  both containerd and CRI-O already ship with the same default directories. According to their documentation ([containerd](https://github.com/containerd/containerd/blob/main/docs/cri/config.md), [CRI-O](https://github.com/cri-o/cri-o/blob/main/docs/crio.conf.5.md)), the default configuration is:

```TOML
cdi_spec_dirs = [
  "/etc/cdi",
  "/var/run/cdi",
]
```

Regarding the documentation update here, I fully agree with both of you. I can add a short section in the README and keep this configurable via Helm as an optional feature that requires an explicit opt-in. That way, it can also serve as a complete reference example of the init container setup for anyone who needs it.

fmuyassarov:
Added https://github.com/kubernetes-sigs/dra-driver-cpu/pull/126/changes/d7d83eed73367201263159b8927a1f83f8193a6e for a short note about requirements for NRI, CDI.

ffromani (commented):
In the DRA Memory driver I have a golang-based [setup helper](https://github.com/ffromani/dra-driver-memory/blob/main/tools/setup/containerd/main.go) which is built and shipped in the base image and which, if needed, users can run as high-privilege job to make sure their cluster is configured as needed.

I think we can consider adopting this pattern **as followup PR**,  and I'd be happy to work on that PR if we deem it worthy (cc @pravk03 ).

Thinking about #36, note that differently from the base CDI/NRI support, the hugepage support would require active setting of non-default values especially on containerd, so that pattern would be helpful here.

For this specific instance I think this PR is fine.

ffromani:
/approve
/lgtm

thanks @fmuyassarov I think we can merge this PR and handle any further cleanup/extension (https://github.com/kubernetes-sigs/dra-driver-cpu/pull/126#pullrequestreview-4147729524) as followup.

fmuyassarov:
> In the DRA Memory driver I have a golang-based [setup helper](https://github.com/ffromani/dra-driver-memory/blob/main/tools/setup/containerd/main.go) which is built and shipped in the base image and which, if needed, users can run as high-privilege job to make sure their cluster is configured as needed.
> 
> I think we can consider adopting this pattern **as followup PR**, and I'd be happy to work on that PR if we deem it worthy (cc @pravk03 ).
> 
> Thinking about #36, note that differently from the base CDI/NRI support, the hugepage support would require active setting of non-default values especially on containerd, so that pattern would be helpful here.
> 
> For this specific instance I think this PR is fine.

I don’t recall if I mentioned this earlier, but I also think that people could benefit from a more robust, runtime agnostic way of editing runtimes (CRI-O, containerd) configs, ideally as a static binary. Happy to take care of it as follow up.

ffromani:
> > In the DRA Memory driver I have a golang-based [setup helper](https://github.com/ffromani/dra-driver-memory/blob/main/tools/setup/containerd/main.go) which is built and shipped in the base image and which, if needed, users can run as high-privilege job to make sure their cluster is configured as needed.
> > I think we can consider adopting this pattern **as followup PR**, and I'd be happy to work on that PR if we deem it worthy (cc @pravk03 ).
> > Thinking about #36, note that differently from the base CDI/NRI support, the hugepage support would require active setting of non-default values especially on containerd, so that pattern would be helpful here.
> > For this specific instance I think this PR is fine.
> 
> I don’t recall if I mentioned this earlier, but I also think that people could benefit from a more robust, runtime agnostic way of editing runtimes (CRI-O, containerd) configs, ideally as a static binary. Happy to take care of it as follow up.

I think setup helper above + a template setup kubernetes job fits the bill. I can post a tentative PR so people can have a look and we can make a more informed decision. Most of the work would be trivial copy/paste + adjustment from dra-driver-memory so no big deal.

I think I can post this today or tomorrow.

fmuyassarov:
Thanks @ffromani

ffromani:
Draft WIP, but sufficient (I hope) to grasp what we can add: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/128

PR #123, opened 2026-04-16
fix: pin github actions to commit SHA

fmuyassarov:
Pin `actions/checkout`, `actions/setup-python` and `actions/setup-go` to commit SHAs as. Otherwise, most of the github action jobs are failing as in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/122, with the following:
```YAML
Error: The actions actions/checkout@v4 and actions/setup-python@v5 are not allowed in kubernetes-sigs/dra-driver-cpu because all actions must be pinned to a full-length commit SHA.
```
```sh
curl -s https://api.github.com/repos/actions/checkout/git/ref/tags/v4 | grep sh
    "sha": "34e114876b0b11c390a56381ad16ebd13914f8d5",

curl -s https://api.github.com/repos/actions/setup-go/git/ref/tags/v5 | grep sh
    "sha": "40f1582b2485089dde7abd97c1529aa768e1baff",
    
curl -s https://api.github.com/repos/actions/setup-python/git/ref/tags/v5 | grep sh
    "sha": "a26af69be951a213d495a4c3e4e4022e16d87065",
```
Also use the same version (v5) of setup-go across all workflows.

PR #127, opened 2026-04-17
remove redundant GitHub Actions E2E tests

pravk03:
This PR removes the file `.github/workflows/e2e.yml` which triggers GitHub Actions to run E2E tests. These tests are already fully covered by [prow](https://github.com/kubernetes/test-infra/blob/876d2263c3b84a2eed75291c8603c2e418b1bd32/config/jobs/kubernetes-sigs/dra-driver-cpu/dra-driver-cpu-presubmits-main.yaml#L4).


pravk03:
/assign @ffromani @fmuyassarov. 
/hold

I see that there was a recent change https://github.com/kubernetes-sigs/dra-driver-cpu/pull/123. Does this redundancy add any value ?

fmuyassarov:
> /assign @ffromani @fmuyassarov. /hold
> 
> I see that there was a recent change #123. Does this redundancy add any value ?

#123 was fixing the version pinning across all the GitHub workflows. 
I don't see any reason to have duplicated test runs and I think it is better to do via Prow as PR authors can re-trigger the test if needed, which isn't possible via GitHub action. As such, this LGTM.
/lgtm

pravk03:
Thanks @fmuyassarov. Agree.

/unhold

ffromani:
Agreed is better to have the e2e tests run via prow. We initially had on github actions only for historical reasons.
OTOH, while we had them, #123 was correct to have (no point in running obsolete config)

PR #129, opened 2026-04-22
contextual logging: initial plumbing

ffromani:
Begin plumbing of contextual logging support. In this PR, we set up the initial plumbing:

- port from unstructured logging to structured logging
- get loggers from the context, respecting the call trees. Functions which don't obviously need a context get a logger instance instead; we don't add a context just to carry a logger
- use testr.New (https://github.com/go-logr/logr/tree/master/testr) in unit tests, to reroute the logging uniformly
- if tests use ginkgo, integrate GinkgoLogr instead
- preserve as much as possible the log structure and content, but there's no 1:1 translation from unstructured logging
- except for egregious cases (klog.Error(nil, ...)) never mutate logging verbosity
- don't remove log lines

NOTE: we don't really exploit the "contextual" part just yet. This will come in a followup PR
NOTE: we don't have means to prevent future PRs accidentally reintroduce unstructured logging; we will explore options (borrowing from k/k) in a followup PR

### DIFF
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -218,7 +218,7 @@ func run(logger logr.Logger) error {
 		CpuDeviceMode:    cpuDeviceMode,
 		CPUDeviceGroupBy: groupBy,
 	}
-	dracpu, err := driver.Start(ctx, clientset, driverConfig)
+	dracpu, asyncErr, err := driver.Start(ctx, clientset, driverConfig)
 	if err != nil {
 		return fmt.Errorf("driver failed to start: %w", err)
 	}
@@ -232,6 +232,9 @@ func run(logger logr.Logger) error {
 		cancel()
 	case <-ctx.Done():
 		logger.Info("exiting", "reason", "context cancelled")
+	case err := <-asyncErr:
+		cancel()
+		return fmt.Errorf("fatal NRI driver error: %w", err)
 	}
 
 	// Gracefully shutdown HTTP server
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"net/http"
@@ -226,6 +227,8 @@ func run(logger logr.Logger) error {
 	ready.Store(true)
 	logger.Info("driver started")
 
+	var fatalErr error
+
 	select {
 	case <-signalCh:
 		logger.Info("exiting", "reason", "received signal")
@@ -234,17 +237,16 @@ func run(logger logr.Logger) error {
 		logger.Info("exiting", "reason", "context cancelled")
 	case err := <-asyncErr:
 		cancel()
-		return fmt.Errorf("fatal NRI driver error: %w", err)
+		fatalErr = fmt.Errorf("NRI driver error: %w", err)
 	}
 
 	// Gracefully shutdown HTTP server
 	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer shutdownCancel()
-	if err := server.Shutdown(shutdownCtx); err != nil {
-		// we prefer to carry on
-		logger.Error(err, "HTTP server shutdown failed")
+	if serverErr := server.Shutdown(shutdownCtx); serverErr != nil {
+		fatalErr = errors.Join(fatalErr, fmt.Errorf("HTTP server shutdown error: %w", serverErr))
 	}
-	return nil
+	return fatalErr
 }
 
 func printVersion(logger logr.Logger) {
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -27,6 +27,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/driver"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/sys/unix"
@@ -35,6 +36,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	nodeutil "k8s.io/component-helpers/node/util"
 	"k8s.io/klog/v2"
+	"k8s.io/klog/v2/textlogger"
 	"k8s.io/utils/cpuset"
 )
 
@@ -110,18 +112,33 @@ func init() {
 }
 
 func main() {
-	klog.InitFlags(nil)
+	config := textlogger.NewConfig()
+	config.AddFlags(flag.CommandLine)
 	flag.Parse()
 
-	if err := run(); err != nil {
-		klog.Fatalf("%v", err)
+	logger := textlogger.NewLogger(config)
+	// some key deps still call klog directly, so we need this integration.
+	// TODO: check every time we bump kube libs to a new major version,
+	// as the contextual logging transition to kube libs is still ongoing
+	// k8s.io/client-go
+	// k8s.io/apimachinery
+	// k8s.io/dynamic-resource-allocation
+	// k8s.io/kube-openapi
+	// k8s.io/utils
+	// k8s.io/component-helpers
+	// k8s.io/kubelet
+	klog.SetLoggerWithOptions(logger, klog.ContextualLogger(true))
+
+	if err := run(logger); err != nil {
+		logger.Error(err, "failed to run")
+		os.Exit(1)
 	}
 }
 
-func run() error {
-	printVersion()
+func run(logger logr.Logger) error {
+	printVersion(logger)
 	flag.VisitAll(func(f *flag.Flag) {
-		klog.Infof("FLAG: --%s=%q", f.Name, f.Value)
+		logger.Info("FLAG", "name", f.Name, "value", f.Value.String())
 	})
 
 	reservedCPUSet, err := cpuset.Parse(reservedCPUs)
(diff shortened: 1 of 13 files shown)


pkg/driver/nri_hooks.go.orig
```diff
@@ -0,0 +1,213 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
```
fmuyassarov: I guess this file is accidentally added to the commit?

pkg/driver/nri_hooks.go
```diff
@@ -187,24 +189,25 @@ func (cp *CPUDriver) StopContainer(ctx context.Context, pod *api.PodSandbox, ctr
...
-func (cp *CPUDriver) RemoveContainer(_ context.Context, pod *api.PodSandbox, ctr *api.Container) error {
-	klog.Infof("RemoveContainer Pod:%s/%s PodUID:%s Container:%s ContainerID:%s", pod.Namespace, pod.Name, pod.Uid, ctr.Name, ctr.Id)
+func (cp *CPUDriver) RemoveContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) error {
+	logger := klog.FromContext(ctx)
+	logger.Info("RemoveContainer", "pod", klog.KObj(pod), "podUID", pod.Uid, "container", ctr.Name, "containerID", ctr.Id)
 	claimUIDs := cp.podConfigStore.RemoveContainerState(types.UID(pod.GetUid()), ctr.GetName())
 	if len(claimUIDs) > 0 {
 		// this serves only for debugging purposes. We should never get here
-		klog.Errorf("RemoveContainer spurious updates needed (unexpected, please file a bug): %v", cp.getSharedContainerUpdates(types.UID(ctr.GetId())))
+		logger.Info("RemoveContainer spurious updates needed (unexpected, please file a bug)", "updates", cp.getSharedContainerUpdates(logger, types.UID(ctr.GetId())))
```
fmuyassarov, line 224: Is the downgrade to info level from error intentional here?

pkg/driver/driver.go
```diff
@@ -194,23 +198,25 @@ func (cp *CPUDriver) Stop() {
...
 	Run(context.Context) error
 }
 
 func runNRIPluginWithRetry(ctx context.Context, plugin nriRunner, maxAttempts int) error {
+	logger := klog.FromContext(ctx)
 	for i := 0; i < maxAttempts; i++ {
 		err := plugin.Run(ctx)
 		if ctx.Err() != nil {
-			klog.Infof("NRI plugin stopped: context cancelled")
+			logger.Info("NRI plugin stopped: context cancelled")
```
fmuyassarov: to follow the same style of contextual logging, perhaps we better do ?
```suggestion
			logger.Info("NRI plugin stopped", "reason", "context cancelled")
```

fmuyassarov (commented):
From the quick initial review, it already looks good. Left few comments inline.

pkg/driver/nri_hooks.go.orig
ffromani: yes, this is a relic of fixing a patch application. Removing.

pkg/driver/nri_hooks.go
ffromani: Yes, and reluctantly. The problem here is that the structured logging API requires an `error` which we don't have. We would need to either construct an error on the fly (awkward) or to call `logger.Error(nil, ...)` (ugly). Plus, the error is actually more a warning then else. All of this combined made me change the level from error to info.

I was not happy that was buried in the commit. I think I'll move the change to just this line in its own commit to make it as evident as it can be/

pkg/driver/driver.go
ffromani: good idea, changing

pkg/driver/nri_hooks.go
ffromani: ended up moving this change in its own commit for visibility

fmuyassarov: thanks

pkg/driver/driver.go
```diff
@@ -194,23 +197,25 @@ func (cp *CPUDriver) Stop() {
...
 	for i := 0; i < maxAttempts; i++ {
 		err := plugin.Run(ctx)
 		if ctx.Err() != nil {
-			klog.Infof("NRI plugin stopped: context cancelled")
+			logger.Info("NRI plugin stopped", "reason", "context cancelled")
 			return ctx.Err()
 		}
 		if err != nil {
-			klog.Infof("NRI plugin failed with error %v, restarting %d out of %d", err, i+1, maxAttempts)
+			logger.Info("NRI plugin failed, restarting", "error", err, "attempt", i+1, "maxAttempts", maxAttempts)
```
pravk03: `logger.Error` here ?

cmd/dracpu/app.go
```diff
@@ -232,6 +232,8 @@ func run(logger logr.Logger) error {
 		cancel()
 	case <-ctx.Done():
 		logger.Info("exiting", "reason", "context cancelled")
+	case err := <-asyncErr:
```
pravk03, line 238: we need a `cancel()` here ?

pkg/driver/cdi_test.go
```diff
@@ -80,18 +81,18 @@ func TestAddDevice(t *testing.T) {
 			})
 			cdiSpecDir = t.TempDir()
 
-			mgr, err := NewCdiManager(testDriverName)
+			mgr, err := NewCdiManager(testr.New(t), testDriverName)
```
pravk03: nit: We can instantiate `logger := testr.New(t)` and reuse in all the subsequent calls.

pkg/driver/cdi.go
```diff
@@ -158,11 +158,11 @@ func (c *CdiManager) readSpecFromFile() (*cdiSpec.Spec, error) {
 	if err := json.Unmarshal(data, spec); err != nil {
 		return nil, fmt.Errorf("error unmarshaling CDI spec from %q: %w", c.path, err)
 	}
-	klog.Infof("Read CDI spec from %q spec:%+v", c.path, spec)
+	logger.V(4).Info("read CDI spec", "path", c.path, "spec", fmt.Sprintf("%+v", spec))
```
pravk03: nit: I think we can pass `spec` directly

pravk03 (commented):
Thank you so much for working on this huge PR and the refactoring looks super clean!

I left some nits inline, but nothing blocking. Feel free to unhold and merge.

/lgtm
/hold

fmuyassarov:
I didn't have a chance yet to go through this after the last push and it is on my TODO list for tomorrow

ffromani:
> Thank you so much for working on this huge PR and the refactoring looks super clean!
> 
> I left some nits inline, but nothing blocking. Feel free to unhold and merge.
> 
> /lgtm /hold

thanks for your review @pravk03 , since @fmuyassarov also wants to review I'll have another pass and address your comments.

cmd/dracpu/app.go
fmuyassarov: good catch

cmd/dracpu/app.go
```diff
@@ -195,31 +218,35 @@ func main() {
...
 	select {
 	case <-signalCh:
-		klog.Infof("Exiting: received signal")
+		logger.Info("exiting", "reason", "received signal")
 		cancel()
 	case <-ctx.Done():
-		klog.Infof("Exiting: context cancelled")
+		logger.Info("exiting", "reason", "context cancelled")
+	case err := <-asyncErr:
+		return fmt.Errorf("fatal NRI driver error: %w", err)
```
fmuyassarov: One thing, I could add to this, since we are touching it, maybe we should not return that early in this `case` and instead do it after the `server.Shutdown(shutdownCtx)` so that we ensure that http server is down?

```go
select {
case <-signalCh:
    logger.Info("exiting", "reason", "received signal")
    cancel()
case <-ctx.Done():
    logger.Info("exiting", "reason", "context cancelled")
case err := <-asyncErr:
    asyncFatalErr = err
    cancel()
}

shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
defer shutdownCancel()
if err := server.Shutdown(shutdownCtx); err != nil {
    logger.Error(err, "HTTP server shutdown failed")
}

if asyncFatalErr != nil {
    return fmt.Errorf("fatal NRI driver error: %w", asyncFatalErr)
}
return nil
```

fmuyassarov (commented):
I left a single suggestion but otherwise it is LGTM from my side.

pkg/driver/cdi.go
ffromani: ok, let's pass the spec directly and if we regress on the logger output we'll have a clear case to reintroduce the formatting.

pkg/driver/driver.go
ffromani: nice catch, fixed

cmd/dracpu/app.go
ffromani: good catch indeed, fixed

pkg/driver/cdi_test.go
ffromani: not a nit: cleaner and better. Fixed, thanks

cmd/dracpu/app.go
ffromani: This is an interesting idea. The problem is: which error is more important? probably neither, because we are in the shutdown path. I think I'll follow your approach but using `errors.Join`.

ffromani:
/hold cancel

removing hold because LGTM was removed by the rebase

fmuyassarov (approved):
Thanks @ffromani. This looks good to me. I could approve it now, but I’d like to give @pravk03 a chance to review and confirm he’s good with the final changes. I’ll leave the final lgtm to him.

pravk03:
Thanks @ffromani.
/lgtm

PR #130, opened 2026-04-22
Use CDI Cache API for spec file management

### LINKED ISSUE #58: Evaluate CDI packages to manage cdi files rather than roll out our own code
We need either to highlight the benefits of using our own CDI file manipulation code (which can be real but are not evident) or we can switch
to CDI utility packages: 
```
"tags.cncf.io/container-device-interface/pkg/cdi"
"tags.cncf.io/container-device-interface/pkg/parser"
"tags.cncf.io/container-device-interface/specs-go"
```

Karthik-K-N:
Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/58

Adopt the upstream CDI cache/spec implementation instead of maintaining our own JSON file manager. This adds some dependency weight, but it removes bespoke spec I/O and synchronization logic, aligns behavior with upstream CDI handling, and should be easier to maintain correctly over time. The evaluation here is that the CDI packages are preferable overall to our in-house implementation. This standard is being followed in other dra drivers as well.


ffromani (commented):
Thanks for the submission! #58 includes a part of the evaluation. The tradeoff, for example, is the extra deps we need to pull in. I can infer from this PR the evaluation is that the CDI packages are overall superior to our in-house implementation.

Could you please add the full rationale to the PR cover letter/git message?

Karthik-K-N:
> Thanks for the submission! #58 includes a part of the evaluation. The tradeoff, for example, is the extra deps we need to pull in. I can infer from this PR the evaluation is that the CDI packages are overall superior to our in-house implementation.
> 
> Could you please add the full rationale to the PR cover letter/git message?

I have updated the PR description, Please let me know if you want me to add more details. I took look in to other drivers like from nvidia and intel, I see similar thing being followed.

pkg/driver/cdi.go
```diff
@@ -37,93 +34,120 @@ var (
...
+// AddDevice adds a device to the CDI spec file using the CDI cache.
 func (c *CdiManager) AddDevice(deviceName string, envVar string) error {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
-
-	spec, err := c.readSpecFromFile()
-	if err != nil {
-		return err
-	}
+	// Get the current spec (or create a new one if it doesn't exist)
```
pravk03: I think we still need the mutex here to make the function threadsafe.

pkg/driver/cdi.go
```diff
@@ -37,93 +34,120 @@ var (
...
+// GetSpec returns the current CDI spec.
 func (c *CdiManager) GetSpec() (*cdiSpec.Spec, error) {
-	return c.readSpecFromFile()
+	return c.getOrCreateSpec(), nil
+}
+
+// getOrCreateSpec retrieves the current spec from the cache or creates a new empty one.
+func (c *CdiManager) getOrCreateSpec() *cdiSpec.Spec {
+	// Refresh once at the beginning of the transaction to catch external file changes
+	if err := c.cache.Refresh(); err != nil {
```
pravk03: Is it safe to log and proceed here ? . 

From [cache.go](https://github.com/cncf-tags/container-device-interface/blob/35765bd41b50a86aa3919eb352bc90321e010e68/pkg/cdi/cache.go), looks like it returns an error when it cannot load the file.

pkg/driver/cdi.go
```diff
@@ -37,93 +34,120 @@ var (
...
+		klog.Warningf("Failed to refresh CDI cache: %v", err)
+	}
+
+	// Get all vendor specs and find ours
+	specs := c.cache.GetVendorSpecs(cdiVendor)
+	for _, spec := range specs {
+		if spec.GetVendor() == cdiVendor && spec.GetClass() == cdiClass {
```
pravk03: Does the GetVendorSpecs() called above not filter based on `cdiVendor` ?

pkg/driver/cdi.go
```diff
@@ -37,93 +34,120 @@ var (
...
+	// Only write the spec if there are devices remaining
+	// The CDI spec validation requires at least one device
+	if len(spec.Devices) > 0 {
+		// Write the updated spec using the cache
+		if err := c.cache.WriteSpec(spec, c.specName); err != nil {
+			return fmt.Errorf("failed to write CDI spec: %w", err)
+		}
+	} else {
+		// If no devices remain, remove the spec file entirely
+		if err := c.cache.RemoveSpec(c.specName); err != nil {
```
pravk03, line 154: Since we return an error when `WriteSpec` fails, lets do the same with `RemoveSpec` to keep things consistent

pravk03 (commented):
Thanks for putting this together! This is a partial review for now. I plan to spend a bit more time to understand the API and do a thorough pass soon.

pkg/driver/cdi.go
Karthik-K-N: Yes, it does, Removed the unnecessary check. Thank you.

pkg/driver/cdi.go
Karthik-K-N: Based on this doc, I thought its safe to ignore, But I have changed it to return error now. Thanks

https://github.com/cncf-tags/container-device-interface/blob/98a7d735d9592c1ccca0130e06b0709f4b65ecbf/pkg/cdi/doc.go#L103-L110

pkg/driver/cdi.go
Karthik-K-N: We have a mutext in Cache, so it should be safe right? https://github.com/cncf-tags/container-device-interface/blob/98a7d735d9592c1ccca0130e06b0709f4b65ecbf/pkg/cdi/cache.go#L127

pravk03: I think that internal mutex only protects the cache's own maps. Since Kubelet can call this in parallel , two calls could read and modify the spec overwriting each other without a mutex to protect.

pkg/driver/cdi.go
```diff
@@ -37,93 +34,126 @@ var (
...
 		Name: deviceName,
 		ContainerEdits: cdiSpec.ContainerEdits{
-			Env: []string{
-				envVar,
-			},
+			Env: []string{envVar},
 		},
 	}
-
 	spec.Devices = append(spec.Devices, newDevice)
```
pravk03: `getOrCreateSpec` calls [GetVendorSpecs](https://github.com/cncf-tags/container-device-interface/blob/98a7d735d9592c1ccca0130e06b0709f4b65ecbf/pkg/cdi/cache.go#L127) which returns a pointer to the spec in cache's internal memory. In this function we are modifying that directly which is not recommended because it mutates shared state outside of the cache.

One alternative we might want to consider is maintaining a map `activeDevices map[string]cdiSpec.Device` in CdiManager struct. In AddDevice() and RemoveDevice() we can update this internal map and call `c.cache.WriteSpec` to sync this with the cache.

pkg/driver/cdi.go
pravk03: Alternatively we can explore a lock free model - https://github.com/kubernetes-sigs/dra-driver-cpu/issues/133. But this can be a separate PR.

Karthik-K-N: Got your point, I will update the PR to use the mutex for now and later I will work on the follow up PR

pkg/driver/cdi.go
```diff
@@ -37,167 +35,139 @@ var (
...
+		driverName:    driverName,
+		specName:      fmt.Sprintf("%s.json", driverName),
+		activeDevices: make(map[string]cdiSpec.Device),
+	}
+
+	// Recovery logic incase of restart.
+	if err := c.cache.Refresh(); err != nil {
+		klog.Warningf("Failed to refresh CDI cache during init (starting clean): %v", err)
+	} else {
+		specs := c.cache.GetVendorSpecs(cdiVendor)
```
pravk03: Lets also match with specName before restore

pkg/driver/cdi.go
```diff
@@ -37,167 +35,139 @@ var (
...
-		return fmt.Errorf("failed to write to temporary CDI spec: %w", err)
-	}
+	return nil
+}
 
-	if err := tmpFile.Sync(); err != nil {
-		return fmt.Errorf("failed to sync temporary CDI spec: %w", err)
-	}
+// GetSpec returns a safe, detached copy of the current CDI spec.
+func (c *CdiManager) GetSpec() *cdiSpec.Spec {
```
pravk03: Since we use this only in tests, it would be better to fetch the spec from cache instead of recreating the spec in this function. The tests can actually verify that syncToCache() succeeded in pushing the state.

pkg/driver/cdi.go
```diff
@@ -37,167 +35,139 @@ var (
...
-		if err != nil {
-			_ = os.Remove(tmpFile.Name())
-		}
-	}()
-
-	data, err := json.MarshalIndent(spec, "", "  ")
-	if err != nil {
-		return fmt.Errorf("error marshaling CDI spec: %w", err)
+	// Write the safely constructed spec to disk via the cache
+	if err := c.cache.WriteSpec(spec, c.specName); err != nil {
```
pravk03, line 177: Lets keep tje ordering deterministic by sorting `spec.Devices` before writing to cache.

pkg/driver/cdi_test.go
pravk03, line 1: I think the tests need a revisit to cover the new changes introduced in this PR. It would be great to add coverage for the startup recovery logic and the file deletion behavior. We may need some tests to ensure cache is always synchronized.

pkg/driver/cdi.go
```diff
@@ -37,167 +37,128 @@ var (
...
-			Devices: []cdiSpec.Device{},
+		cache:         cache,
+		cdiKind:       fmt.Sprintf("%s/%s", cdiVendor, cdiClass),
+		driverName:    driverName,
+		specName:      fmt.Sprintf("%s.json", driverName),
+		activeDevices: make(map[string]cdiSpec.Device),
+	}
+
+	// Recovery logic in case of restart.
+	if err := c.cache.Refresh(); err != nil {
```
pravk03, line 71: Looks like this [API](https://github.com/cncf-tags/container-device-interface/blob/98a7d735d9592c1ccca0130e06b0709f4b65ecbf/pkg/cdi/cache.go#L126) returns errors if any file within the CDI dir is malformed. We may just want to log the errors here and still attempt recovery.

pkg/driver/cdi.go
```diff
@@ -37,167 +37,128 @@ var (
...
-	return deviceFound
-}
 
-func (c *CdiManager) readSpecFromFile() (*cdiSpec.Spec, error) {
-	data, err := os.ReadFile(c.path)
-	if err != nil {
-		return nil, fmt.Errorf("error reading CDI spec file %q: %w", c.path, err)
-	}
+	// Remove from internal state
+	delete(c.activeDevices, deviceName)
```
pravk03, line 136: We may introduce a rare bug here. If syncToCache() returns an error, Kubelet retries `UnprepareResourceClaims`. But since its already removed from memory  in `delete(c.activeDevices, deviceName)`, the subsequent call will pass without actually removing from the cache. 

I would suggest rolling back the change on error 

```
	device := c.activeDevices[deviceName]
	delete(c.activeDevices, deviceName)

	if err := c.syncToCache(); err != nil {
		// Revert the state
		c.activeDevices[deviceName] = device 
		return err
	}
	return nil
```

pravk03:
Thanks for continuing to work on this @Karthik-K-N. I left a couple more comments. Lets also wait for @ffromani for his feedback. We may want to incorporate the contextual logging changes after https://github.com/kubernetes-sigs/dra-driver-cpu/pull/129 merges.

Karthik-K-N:
> Thanks for continuing to work on this @Karthik-K-N. I left a couple more comments. Lets also wait for @ffromani for his feedback. We may want to incorporate the contextual logging changes after #129 merges.

Thanks for the continued review and guidance and apologies for taking it long, will make sure to be more precise next time.

pkg/driver/cdi.go
Karthik-K-N: Added recovery logic for both AddDevice and RemoveDevice.

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
-			Kind:    cdiKind,
-			Devices: []cdiSpec.Device{},
-		}
-		if err := c.writeSpecToFile(logger, initialSpec); err != nil {
-			return nil, err
+	// Recovery logic in case of restart.
+	// We log errors but always attempt recovery because another driver's
+	// malformed CDI file should not prevent us from recovering our own valid state.
+	if err := c.cache.Refresh(); err != nil {
+		logger.Info("Encountered errors refreshing CDI cache (attempting recovery anyway)", "error", err)
```
pravk03, line 72: Wondering if we should switch this to an `Error` message, but the downside is that a malformed file in the directory in a different driver leaves an error message in the cpu-dra-driver. IMO, leaving this as Info maybe the safe choice so no action needed here, just leaving this comment for future reference.

pravk03:
> Thanks for the continued review and guidance and apologies for taking it long, will make sure to be more precise next time.

No worries at all, and definitely no need to apologize! Thanks for all the effort in getting familiar with the new API and introducing this important change.

I think the changes look robust now, so I am adding my LGTM. I'll hold off on the `approve` to give @ffromani some time to review in case he has any additional thoughts.

/lgtm

ffromani:
thanks folks, I'll review shortly

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
 	cdiSpecDir = "/var/run/cdi"
 )
 
-// CdiManager manages a single CDI JSON spec file using a mutex for thread safety.
+// CdiManager manages CDI devices using the CDI cache.
+// It provides a simplified interface for adding and removing devices
+// while leveraging the CDI package for spec management.
```
ffromani: how we implement should not be part of the document, so we have freedom to change; IOW the docstring should document the API, the invariants, the purpose, not the implementation- the latter is best done using inline comments

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
+	activeDevices map[string]cdiSpec.Device
 }
 
 // NewCdiManager creates a manager for the driver's CDI spec file.
 func NewCdiManager(logger logr.Logger, driverName string) (*CdiManager, error) {
-	path := filepath.Join(cdiSpecDir, fmt.Sprintf("%s.json", driverName))
-
-	if err := os.MkdirAll(cdiSpecDir, 0755); err != nil {
-		return nil, fmt.Errorf("error creating CDI spec directory %q: %w", cdiSpecDir, err)
+	// Create a CDI cache with the spec directory
```
ffromani: the comment seems redundant

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
+	cache, err := cdiapi.NewCache(
+		cdiapi.WithSpecDirs(cdiSpecDir),
+		cdiapi.WithAutoRefresh(false),
```
ffromani: OTOH it would be worthy a one-liner to record why we don't want autorefresh

pkg/driver/cdi.go
ffromani: I agree. Error() seems best suited for fatal errors. Info() seems better suited for recoverable errors. In this case these errors are expected to be non-fatal but rare, so V(0) seems the right level

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
+	// malformed CDI file should not prevent us from recovering our own valid state.
+	if err := c.cache.Refresh(); err != nil {
+		logger.Info("Encountered errors refreshing CDI cache (attempting recovery anyway)", "error", err)
+	}
+
+	// Always check the cache, even if Refresh threw a warning
+	specs := c.cache.GetVendorSpecs(cdiVendor)
+
+	for _, spec := range specs {
+		// Match both the CDI class and the exact target filename
```
ffromani: this seems obvious from the code though

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
-		if err := c.writeSpecToFile(logger, initialSpec); err != nil {
-			return nil, err
+	// Recovery logic in case of restart.
+	// We log errors but always attempt recovery because another driver's
+	// malformed CDI file should not prevent us from recovering our own valid state.
+	if err := c.cache.Refresh(); err != nil {
+		logger.Info("Encountered errors refreshing CDI cache (attempting recovery anyway)", "error", err)
+	}
+
+	// Always check the cache, even if Refresh threw a warning
```
ffromani: "..even if Refresh() returned errors"?

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
+	specs := c.cache.GetVendorSpecs(cdiVendor)
+
+	for _, spec := range specs {
+		// Match both the CDI class and the exact target filename
+		if spec.GetClass() == cdiClass && filepath.Base(spec.GetPath()) == c.specName {
+			for _, device := range spec.Devices {
+				c.activeDevices[device.Name] = device
+			}
+
+			logger.Info("Successfully recovered state", "devices", len(c.activeDevices))
```
ffromani, line 84: likewise this is best indeed to be logged as V(0) because recovery is both important and infrequent

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
+			break
 		}
-	} else if err != nil {
-		return nil, fmt.Errorf("error accessing CDI spec file %q: %w", path, err)
 	}
 
-	logger.Info("initialized CDI file manager", "path", path)
+	logger.Info("Initialized CDI manager with cache", "specName", c.specName)
```
ffromani, line 89: why did we remove "path" again?

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
 	}
 
-	spec.Devices = append(spec.Devices, newDevice)
-	return c.writeSpecToFile(logger, spec)
-}
-
-// RemoveDevice removes a device from the CDI spec file.
-func (c *CdiManager) RemoveDevice(logger logr.Logger, deviceName string) error {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
```
ffromani, line 108: in the future we may want to have an explicit Rollback() function and perhaps trigger with a named return, but as starting point this seems fine

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
 		}
-		return err
-	}
-
-	if removeDeviceFromSpec(spec, deviceName) {
-		return c.writeSpecToFile(logger, spec)
+		return fmt.Errorf("failed to sync cache during addition (state rolled back): %w", err)
 	}
 
+	logger.Info("Added CDI device", "deviceName", deviceName, "env", envVar)
```
ffromani: this seems V(2)? Perhaps even V(4)? it's normal operation and part of the standard flow

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
+		return fmt.Errorf("failed to sync cache during removal (state rolled back): %w", err)
 	}
 
-	spec := &cdiSpec.Spec{}
-	if err := json.Unmarshal(data, spec); err != nil {
-		return nil, fmt.Errorf("error unmarshaling CDI spec from %q: %w", c.path, err)
-	}
-	logger.V(4).Info("read CDI spec", "path", c.path, "spec", spec)
-	return spec, nil
+	logger.Info("Removed CDI device", "deviceName", deviceName)
```
ffromani: likewise, V(2) or V(4)?

pkg/driver/cdi.go
```diff
@@ -37,167 +37,151 @@ var (
...
-		return fmt.Errorf("failed to close temporary CDI spec: %w", err)
-	}
+	// Sort devices by name to ensure deterministic order
+	sort.Slice(spec.Devices, func(i, j int) bool {
+		return spec.Devices[i].Name < spec.Devices[j].Name
+	})
 
-	if err := os.Rename(tmpFile.Name(), c.path); err != nil {
-		return fmt.Errorf("failed to rename temporary CDI spec: %w", err)
+	// Write the safely constructed spec to disk via the cache
```
ffromani: the comments seems redundant here

pkg/driver/cdi_test.go
```diff
@@ -177,11 +196,50 @@ func TestRemoveDevice(t *testing.T) {
...
+			got := getSpecFromCache(mgr)
 			if diff := cmp.Diff(got, tcase.expectedSpec); diff != "" {
-				t.Errorf("unexpected spec from empty: %v", diff)
+				t.Errorf("unexpected spec: %v", diff)
 			}
 		})
 	}
 }
+
+func TestStateRecovery(t *testing.T) {
```
ffromani, line 207: nice addition. Are there known test gaps? do we see the test coverage as sufficient to exercise the changes?

ffromani (commented):
Thanks for the contribution. Mostly questions and minor improvements inside. Do we see testing gaps we want to fill?

Also, please make sure to record the rationale and the benefit of this change in the commit message.

pkg/driver/cdi.go
Karthik-K-N: Isn't the default level is V(0)?

pkg/driver/cdi.go
Karthik-K-N: We have removed the path field.

pkg/driver/cdi.go
Karthik-K-N: Changed to V(4).

ffromani:
/lgtm

thanks! we may add more tests but this can be done as followup.

Issue #63, opened 2026-02-17
Issue reporting/troubleshooting template and tooling


ffromani:
Once we have the first release, we need a step-by-step guide to help users report issues efficiently. Which data should they make sure to include? What's the data we are sure we always need to have to troubleshoot user issue (e.g. NUMA topology, machine listing). Can we create helper tools to gather the data we know we need, so the user can just run a simple helper and add their report on top?

ffromani:
more design thoughts continuing from https://github.com/kubernetes-sigs/dra-driver-cpu/pull/143#issuecomment-4438813242

## naming

`debuginfo` is overloaded. In Openshift there is [must-gather](https://github.com/openshift/must-gather). This tool does something similar in spirit, so perhaps `dracpu-gatherinfo` ? or even `dracpu-diagnose`? but diagnose seems more about live diagnosis. Maybe `dracpu-collect-info`? Honestly out of habit I'd go for `dracpu-gatherinfo`

AutuSnow:
@pravk03  Do you have any other suggestions or better ideas to draw on?

pravk03:
Sorry for the delay. I'll get to this soon, likely tomorrow, as I'm out of the office today.

pravk03:
Thanks for driving this work @AutuSnow. Regarding naming, I agree with @ffromani . `dracpu-gatherinfo` is intuitive and fits well.

Since the primary goal of this tool is to help us debug user issues effectively, I wanted to share my mental model for the data we would ideally want to have. I don't expect we need everything in the initial PR—we can build this iteratively once we agree on the core details we want to include and structure.


Here is what I think would be most valuable to capture:

1. Hierarchical Topology
    - Node: Total CPUs, Total Cores, SMT (on/off), Total L3 caches
    - Socket 0
       - NUMA 0: Num CPUs, Num L3 caches, CPUMask
          - L3 0: Num CPUs, CPUMask
          - L3 1: Num CPUs, CPUMask

2. Driver Configuration
     - startup parameters, flags, etc.

3. Resource Stats
    - Node Stats 
       * CPU:  Available, Reserved, Exclusive, and Shared CPUMasks
    - Claim Stats 
      * Active Claims -  Claim name, resource request, constraints, assigned CPUMask


[...]

PR #131, opened 2026-04-24
refactor: migrate CLI to Cobra framework

AutuSnow:
Benefits:
- Subcommand support (foundation for #63 debuginfo tool)
- Better help formatting
- Standard pattern 
- All klog flags now visible in --help

Related: #63


ffromani:
Thanks for filing this PR

> Benefits:
> 
>     * Subcommand support (foundation for [Issue reporting/troubleshooting template and tooling #63](https://github.com/kubernetes-sigs/dra-driver-cpu/issues/63) debuginfo tool)

I need to review the conversation here because I don't remember what we told here. Another cheaper option could be add symlinks to the main executable and to make logic on `os.Args[0]`, which is a old common unix pattern.
IOW, I'm not yet convinced we _need_ cobra for this

>     * Better help formatting
> 
>     * Standard pattern

True benefits. Can we run a quick survey to see how cobra is actually popular in the kube ecosystem? it's like overwhelming popular? popular? mixed adoption?

>     * All klog flags now visible in --help

This is again a true benefit but we don't _need_ cobra to achieve this.

Basically, my main concern here is if the extra dep chain pulled by cobra is worth the benefits.

pohly:
Cobra is fairly standard for Go command line apps. However, it's maintenance was or is problematic. For a binary that hardly ever gets invoked multiple times with different parameters by normal users it seems overkill to me.

AutuSnow:
Thank you for your rebuttal, which made me feel that introducing Cobra is not the best framework. I will use the symbolic linking method mentioned above to complete the function of bebuginfo

PR #135, opened 2026-05-01
Skip CPU allocation when available CPU pool is empty

### LINKED ISSUE #134: prepareGroupedResourceClaim fails when kubelet CPU manager owns all CPUs
#### What happened?

When `cpuManagerPolicy: static` is configured, the kubelet CPU manager owns all non-reserved CPUs. The DRA CPU driver's shared pool (`GetSharedCPUs()`) is empty because no CPUs are available for DRA allocation. `TakeByTopologyNUMAPacked` is called with an empty available set and fails with "not enough cpus available to satisfy request", causing `NodePrepareResources` to fail.

This happens when a pod has DRA CPU claims but the kubelet CPU manager is responsible for CPU pinning (option B in the DRA topology-aware placement design). The DRA CPU device is used only as a topology marker for `matchAttribute: resource.kubernetes.io/numaNode` alignment — the actual CPU pinning is handled by the kubelet.

#### What did you expect to happen?

When the available CPU pool is empty, the driver should skip allocation and return a `PrepareResult` with the device entries but no CDI cpuset injection. The NRI `CreateContainer` hook already handles this case — it falls through to the shared CPU pool when no `DRA_CPUSET_*` env var is set.

#### How to reproduce it?

1. Configure kubelet with `cpuManagerPolicy: static` and `reservedSystemCPUs: "0-3"`
2. Deploy the DRA CPU driver
3. Create a ResourceClaim requesting a CPU device
4. Create a Guaranteed QoS pod referencing the claim
5. `NodePrepareResources` fails with "not enough cpus available to satisfy request"

#### Environment

- Kubernetes v1.36+
- DRA CPU driver with `--cpu-device-mode=grouped` or `--cpu-device-mode=individual`
- `cpuManagerPolicy: static`

/kind bug

johnahull:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When `cpuManagerPolicy: static` is active, the kubelet CPU manager owns all non-reserved CPUs. The DRA CPU driver's shared pool is empty, causing `TakeByTopologyNUMAPacked` to fail with "not enough cpus available to satisfy request."

This fix:
1. Skips the allocation attempt when `availableCPUsForDevice` is empty
2. When no CPUs are assigned for the entire claim, returns a `PrepareResult` with Device entries (pool, device, request) but no CDI cpuset injection

The NRI `CreateContainer` hook already handles missing DRA cpuset env vars by falling through to the shared CPU pool. The kubelet CPU manager's cgroup cpuset takes precedence for dedicated pods.

This allows the DRA CPU driver to coexist with `cpuManagerPolicy: static` without failing prepare for pods that have DRA CPU claims used purely as topology markers for `matchAttribute` alignment.

#### Which issue(s) this PR is related to:

Fixes https://github.com/kubernetes-sigs/dra-driver-cpu/issues/134

#### Does this PR introduce a user-facing change?

```release-note
Fix DRA CPU driver to skip allocation when kubelet CPU manager owns all CPUs (cpuManagerPolicy: static).
```

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -327,6 +327,11 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 			logger.Info("NUMA node CPU availability", "numaNodeID", numaNodeID, "numaCPUs", numaCPUs.String(), "availableCPUs", availableCPUsForDevice.String())
 		}
 
+		if availableCPUsForDevice.Size() == 0 {
+			logger.V(2).Info("No available CPUs for device, skipping allocation", "device", alloc.Device)
+			continue
+		}
+
 		cur, err := cpumanager.TakeByTopologyNUMAPacked(logger, topo, availableCPUsForDevice, int(claimCPUCount), cpumanager.CPUSortingStrategyPacked, true)
 		if err != nil {
 			return kubeletplugin.PrepareResult{Err: err}
@@ -336,8 +341,19 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 	}
 
 	if cpuAssignment.Size() == 0 {
-		logger.V(5).Info("claim has no CPU allocations for this driver", "claim", klog.KObj(claim))
-		return kubeletplugin.PrepareResult{}
+		logger.V(2).Info("No CPUs assigned for claim, returning devices without cpuset", "claim", klog.KObj(claim))
+		var preparedDevices []kubeletplugin.Device
+		for _, allocResult := range claim.Status.Allocation.Devices.Results {
+			if allocResult.Driver != cp.driverName {
+				continue
+			}
+			preparedDevices = append(preparedDevices, kubeletplugin.Device{
+				PoolName:   allocResult.Pool,
+				DeviceName: allocResult.Device,
+				Requests:   []string{allocResult.Request},
+			})
+		}
+		return kubeletplugin.PrepareResult{Devices: preparedDevices}
 	}
 
 	cp.cpuAllocationStore.AddResourceClaimAllocation(logger, claim.UID, cpuAssignment)


pravk03:
Thanks @johnahull for filing the issue.

> When cpuManagerPolicy: static is active, the kubelet CPU manager owns all non-reserved CPUs. The DRA CPU driver's shared pool is empty, causing TakeByTopologyNUMAPacked to fail with "not enough cpus available to satisfy request."

I am a little confused by this part. The Kubelet and the DRA driver scan the machine topology independently and I am not clear as to how enabling static CPU policy is making the DRA driver's pool empty? Are you passing a specific `--reserved-cpus` setting to the driver ?

Right now, the DRA driver and the Kubelet's static CPU policy do not work well together. There is not way to prevent Kubelet (with static policy) allocate the CPUs already allocated to the claim by the driver. The current recommendation is to disable the static CPU policy in Kubelet.

When DRA driver is used with Kubelet static policy disabled 
1. It pins CPUs to containers with claims
2. It maintains a shared pool and assigns them to container without claims.

> DRA CPU claims used purely as topology markers for matchAttribute alignment.

Does this mean you want to use the DRA driver just to guide the scheduler (i'e ensure the numa domain / socket in `matchAttribute` has enough capacity), without the driver actually pinning the CPUs?. If yes, why are we enabling Kubelet static CPU policy ?. I am not sure if I understand the use case here, could you please share more details.

ffromani:
> Right now, the DRA driver and the Kubelet's static CPU policy do not work well together. There is not way to prevent Kubelet (with static policy) allocate the CPUs already allocated to the claim by the driver. The current recommendation is to disable the static CPU policy in Kubelet.

I think we somehow missed to make this explicitly and loud in the docs. People already heavily invested and/or working in this area took that for granted and we always worked with this assumption in mind but it was so obvious to us that we failed to record. I think this is the real true bug here. The contributed fix seems to have true merit (didn't review it yet had a quite crazy past few days), but the doc gaps seems to be the real thing IMO.

I'll file a PR to fix the docs and the README.md

pravk03:
> I think we somehow missed to make this explicitly and loud in the docs.

Woah. I thought this was already covered in the docs. For some reason, I had it in my head that this limitation was documented since it's been discussed several times. Definitely agree - let's update the docs to make this clear.

johnahull:
@pravk03 Thanks for the feedback.                                                          
                                                                                                                                        
  ▎ I am a little confused by this part. Are you passing a specific --reserved-cpus setting to the driver?
                                                                                                                                        
  No custom --reserved-cpus. The pool is empty because cpuManagerPolicy: static takes ownership of all non-reserved CPUs.               
                                                                                                                                        
  ▎ Does this mean you want to use the DRA driver just to guide the scheduler... If yes, why are we enabling Kubelet static CPU policy? 
                                                                                             
  I had the DRA CPU driver still deployed from earlier testing when I switched to cpuManagerPolicy: static to test a different approach. In that approach, the kubelet CPU manager handles pinning and a kubelet patch (#138732) reads numaNode from GPU/NIC ResourceSlices to align CPU pinning with DRA devices. The DRA CPU driver isn't needed for that — but it was still running and crashed on the empty pool.                                                                                      
                                               
  The driver shouldn't crash if it happens to be deployed alongside cpuManagerPolicy: static. I think a skip with a warning log is better than a hard failure. But if you'd prefer to just document the incompatibility and close this, I have no problem.

ffromani:
> > I think we somehow missed to make this explicitly and loud in the docs.
> 
> Woah. I thought this was already covered in the docs. For some reason, I had it in my head that this limitation was documented since it's been discussed several times. Definitely agree - let's update the docs to make this clear.

Here: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/136

@johnahull the doc bug was clear and major. I'm open to improve the code and I'll review your PR carefully ASAP.

pkg/driver/dra_hooks.go
```diff
@@ -324,6 +324,11 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(ctx context.Context, claim *res
 			klog.Infof("NUMA node %d CPUs:%s available CPUs: %s", numaNodeID, numaCPUs.String(), availableCPUsForDevice.String())
 		}
 
+		if availableCPUsForDevice.Size() == 0 {
```
ffromani, line 330: I think there's a merit in adding this safety guard, it's probably the last line of defence we can add to detect a misconfigured cluster. But we should hard fail in this path rather than carry on.
@pravk03 thoughts?

ffromani: Another thought. Since we are moving towards a [golang-based setup helper](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/128), which users can run when they wish to (e.g. as kubernetes job) and not necessarily as init container, would it make sense to validate the kubelet config?

pravk03: I am ok with adding this check, but I wonder if this check would ever get triggered in practice.

From my understanding, the only reason it may be triggered is when a pod with scheduled on the node, but we do not have enough CPUs available on the node (or socket/numa if we use some form of filtering in the claim).  But this should not happen in the first place, because the scheduler should not select the node.

Even with a misconfiguration (i.e., enabling the static CPU policy while using the driver), kubelet might have assigned the CPUs to a different pod. However, from the DRA driver's point of view, if a CPU is not assigned to another claim, it should still be considered available. The DRA driver does not know what is assigned by kubelet, and hence we would not trigger this check.

pkg/driver/dra_hooks.go
```diff
@@ -334,8 +339,19 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(ctx context.Context, claim *res
 	}
 
 	if cpuAssignment.Size() == 0 {
```
pravk03, line 343: 1. We can consider broadening the scope of this check to include any mismatch - `cpuAssignment.Size() != totalClaimCPUCount`
2. How about returning an error and failing early ? 
`return kubeletplugin.PrepareResult{Err: "allocated CPU count does not match requested count"}`

pravk03: > We can consider broadening the scope of this check to include any mismatch - cpuAssignment.Size() != totalClaimCPUCount

Upon checking, this cannot happen has we would throw an error in TakeByTopologyNUMAPacked().

pkg/driver/dra_hooks.go
ffromani: I think you have a point here. The main issue was the documentation bug (first) and lack of automated validation helpers (second). The doc bug was addressed, automated validation helpers require more design and effort.

ffromani:
Per thread ending at https://github.com/kubernetes-sigs/dra-driver-cpu/pull/135#discussion_r3396108709 not much else we can do here. If we have a more detailed use case or a new bug, please reopen/file a new bug so we can evaluate new evidence

PR #137, opened 2026-05-05
implement contextual logging

ffromani:
Implement the actual contextual logging for all our major flows, which map to driver startup/management and DRA and NRI callbacks.
At the topmost level of each flow, we augment the logger with key-value pairs relevant to the flow which will be automatically logged in every function down the call chain.
Therefore, every message downstream will always get these KV pairs, making easier to track the flow in the logs.
    
To ensure consistency, we intentionally override the context to include the augmented logger, so functions fetching the logger from the context get a consistent logger.

Some DRA flows, most notably PrepareResourceClaim and UnprepareResourceClaim, are designed to work on a batch of claims.
We added enough contextual log to be able to follow the flow of each claim, but we lack a way to easily correlate claims within a batch.
Therefore we introduce a custom short ID to group the claims pertaining to a batch. 
Another option could have been to use a UUID, but the UUID identifiers are long and add noise to the logs and we don't really need yet such uniqueness guiaranties.
Therefore, we start with a custom hexadecimal short id.

Turns out this `batchID` concept can be useful in general, so we generalize it into a opID and use it everywhere in out main operation (aka DRA and NRI callback implementations).
This allows to trivially demux the interleaved logs into in high concurrency scenarios at a very cheap cost.
When (if?) we integrate OTEL we will replace this with OTEL trace_id/span_id, perhaps with a transitional period.

We mark begin and end of each callback so it's possible to  de-entangle massively concurrent logs.
We intentionally use minimal delimiters without extra data like elapsed time, result, and so forth because the intent is to make easy to parse log to understand the component behavior aka the "why", not to replace telemetry.
The fact we add pseudo-span delimiters is an artifact of how log are emitted - an interleaved streams rather than an attempt to reimplement e.g. OTEL spans poorly.
Likewise the `opID` concept, this is another concept we borrow from telemetry because makes sense in the logs; it's lightweight, useful and improves across a different dimension.
Pending further review, when (if?) we integrate OTEL this concept should remain.

Finally, we add e2e tests to ensure basic coverage of the log properties introduced here.
We use ReportAfterSuite, which runs after all specs in the suite have completed.
The alternative would be using a dedicated spec that must be ordered last, which seems more fragile and cumbersome. We may re-evaluate in the future depending on how the e2e suite evolves.

Builds on top of #129

### DIFF
--- a/docs/dev/logging.md
+++ b/docs/dev/logging.md
@@ -3,7 +3,7 @@
 ## picking a verbosity level
 
 V(0): high-signal operational logs and errors always visible at default verbosity
-V(2): normal operation, important decisions, summary
+V(2): normal operation, important decisions, summary, container lifecycle
 V(4): allocation internals, fine details about decisions
 V(6): debug messages, very fine details, mostly for developers
 
--- a/docs/dev/logging.md
+++ b/docs/dev/logging.md
@@ -1,0 +1,46 @@
+# logging guidelines
+
+## picking a verbosity level
+
+V(0): high-signal operational logs and errors always visible at default verbosity
+V(2): normal operation, important decisions, summary
+V(4): allocation internals, fine details about decisions
+V(6): debug messages, very fine details, mostly for developers
+
+V(7) or more is reserved for future usage.
+
+## the opID tag
+
+`opID` is a cheap trace identifier similar in spirit to OTEL's
+(OTEL=OpenTelemetry) `trace_id`. Same concept, applied to logs:
+a unique identifier constantly
+logged in all the entries pertaining a flow, which trivially
+enable to extract all the logs with `grep`.
+
+Other examples sit in `test/e2e/contextual_logging_test.go`
+and in `test/pkg/logcheck/`.
+
+We reimplement because the concept is super cheap and super useful
+for logs, even without telemetry implementation.
+
+If in the future we integrate with OTEL, we will add OTEL's
+identifier in addition to the current opID for a transitional
+period, then we will remove our homegrown solution.
+
+### sizing opID
+
+Since we use it everywhere, we need to minimize the log noise,
+so we set the value of `opIDLen` to "only" 8 hex char.
+We evaluate this value to still provide plenty of
+randomness for a *local only* (not distributed) identifier.
+
+Here's the back-of-envelope math:
+
+- 12 hex chars = 48 bits → ~281 trillion values
+- 8 hex chars = 32 bits → ~4.3 billion values
+
+The probability of at least one collision among n operations is roughly n\*\*2/(2N).
+With 8 chars:
+
+- 1000 operations: ~0.01% collision probability.
+- 10000 operations: ~1.2% collision probability.
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -57,7 +57,7 @@ const (
 
 // createGroupedCPUDeviceSlices creates Device objects based on the CPU topology, grouped by a specific criteria.
 func (cp *CPUDriver) createGroupedCPUDeviceSlices(logger logr.Logger) [][]resourceapi.Device {
-	logger.Info("creating grouped CPU devices", "groupBy", cp.cpuDeviceGroupBy)
+	logger.Info("creating grouped CPU devices")
 	var devices []resourceapi.Device
 
 	topo := cp.cpuTopology
@@ -224,7 +224,8 @@ func (cp *CPUDriver) createCPUDeviceSlices() [][]resourceapi.Device {
 
 // PublishResources publishes ResourceSlice for CPU resources.
 func (cp *CPUDriver) PublishResources(ctx context.Context) {
-	logger := klog.FromContext(ctx)
+	logger := klog.FromContext(ctx).WithValues("deviceMode", cp.cpuDeviceMode, "groupBy", cp.cpuDeviceGroupBy)
+	ctx = klog.NewContext(ctx, logger)
 	logger.Info("publishing resources")
 
 	var deviceChunks [][]resourceapi.Device
@@ -269,12 +270,13 @@ func (cp *CPUDriver) PrepareResourceClaims(ctx context.Context, claims []*resour
 	}
 
 	for _, claim := range claims {
+		cLogger := logger.WithValues("claim", klog.KObj(claim), "claimUID", claim.UID)
 		if cp.cpuDeviceMode == CPU_DEVICE_MODE_GROUPED {
-			logger.Info("claim is for a grouped resource", "claim", klog.KObj(claim))
-			result[claim.UID] = cp.prepareGroupedResourceClaim(logger, claim)
+			cLogger.Info("claim is for a grouped resource")
+			result[claim.UID] = cp.prepareGroupedResourceClaim(cLogger, claim)
 		} else {
-			logger.Info("claim is for an individual resource", "claim", klog.KObj(claim))
-			result[claim.UID] = cp.prepareResourceClaim(logger, claim)
+			cLogger.Info("claim is for an individual resource")
+			result[claim.UID] = cp.prepareResourceClaim(cLogger, claim)
 		}
 	}
 	return result, nil
@@ -285,7 +287,7 @@ func getCDIDeviceName(uid types.UID) string {
 }
 
 func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *resourceapi.ResourceClaim) kubeletplugin.PrepareResult {
-	logger.Info("preparing grouped resource claim", "claim", klog.KObj(claim))
+	logger.Info("preparing grouped resource claim")
 
 	if claim.Status.Allocation == nil {
 		return kubeletplugin.PrepareResult{
@@ -303,7 +305,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		if quantity, ok := alloc.ConsumedCapacity[cpuResourceQualifiedName]; ok {
 			count := quantity.Value()
 			claimCPUCount = count
-			logger.Info("found CPU request", "numCPUs", count, "device", alloc.Device, "claim", klog.KObj(claim))
+			logger.Info("found CPU request", "numCPUs", count, "device", alloc.Device)
 		}
 
 		topo := cp.cpuTopology
@@ -336,7 +338,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 	}
 
 	if cpuAssignment.Size() == 0 {
-		logger.V(5).Info("claim has no CPU allocations for this driver", "claim", klog.KObj(claim))
+		logger.V(5).Info("claim has no CPU allocations for this driver")
 		return kubeletplugin.PrepareResult{}
 	}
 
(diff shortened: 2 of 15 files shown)


ffromani:
un-drafting to let CI run so I can fine-tune the e2e test

ffromani:
*update* we need another mechanical followup because reviewing our usage of logr, some trivial klog boilerplate calls (NewContext/FromContext/KObj) are unnecessary and still create avoidable coupling *in the application layer*. We still need to integrate properly with libraries using klog directly, but we can simplify our code even more.

This is going to be a mechanical change. I'll post once this work is done.

fmuyassarov:
@ffromani I guess this is ready now for review?

ffromani:
> @ffromani I guess this is ready now for review?

yes folks sorry for being a bit terse. The PR is completed and ready for review

fmuyassarov:
> > @ffromani I guess this is ready now for review?
> 
> yes folks sorry for being a bit terse. The PR is completed and ready for review

all good :), I will go through it today

pkg/driver/dra_hooks.go
```diff
@@ -260,7 +261,10 @@ func (cp *CPUDriver) PublishResources(ctx context.Context) {
 
 // PrepareResourceClaims is called by the kubelet to prepare a resource claim.
 func (cp *CPUDriver) PrepareResourceClaims(ctx context.Context, claims []*resourceapi.ResourceClaim) (map[types.UID]kubeletplugin.PrepareResult, error) {
-	logger := klog.FromContext(ctx)
+	logger := klog.FromContext(ctx).WithValues("batchID", generateShortID(batchIDLen))
+	// unnecessary now, but easy to miss in the future. Reset the context to a known-good state to be forward compatible
+	_ = klog.NewContext(ctx, logger)
```
swatisehgal: What is the reason derived context here but then immediately discarding it? For a future-proof contextual propagation, shouldn't we retain the derived context?

pkg/driver/dra_hooks.go
```diff
@@ -438,7 +442,10 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 
 // UnprepareResourceClaims is called by the kubelet to unprepare the resources for a claim.
 func (cp *CPUDriver) UnprepareResourceClaims(ctx context.Context, claims []kubeletplugin.NamespacedObject) (map[types.UID]error, error) {
-	logger := klog.FromContext(ctx)
+	logger := klog.FromContext(ctx).WithValues("batchID", generateShortID(batchIDLen))
+	// unnecessary now, but easy to miss in the future. Reset the context to a known-good state to be forward compatible
+	_ = klog.NewContext(ctx, logger)
+
```
swatisehgal: Same as above.

pkg/driver/nri_hooks.go
```diff
@@ -152,7 +154,8 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
 	// unnecessary now, but easy to miss in the future. Reset the context to a known-good state to be forward compatible
 	_ = klog.NewContext(ctx, logger)
 
-	logger.Info("CreateContainer")
+	logger.V(4).Info("begin: CreateContainer")
+	defer logger.V(4).Info("end: CreateContainer")
```
swatisehgal: Container lifecycle transitions like (CreateContainer, StopContainer, RemoveContainer etc) are operationally important. Should we bump the verbosity level to (atleast) 2? I understand if this is cosnidered outside the scope of this PR but wanted to point out regardless.

swatisehgal: Ah! I see in the next commit we are updating the verbosity level of the logs so perhaps we can consider making updates to the container lifecycle logs as well.

docs/dev/logging.md
```diff
@@ -0,0 +1,46 @@
+# logging guidelines
+
+## picking a verbosity level
+
+V(0): must-see logs, errors
+V(2): normal operation, important decisions, summary
+V(4): allocation internals, fine details about decisions
+V(6): debug messages, very fine details, mostly for developers
```
swatisehgal: I would assign V(2) log level to container lifecycle logs.

docs/dev/logging.md
```diff
@@ -0,0 +1,46 @@
+# logging guidelines
+
+## picking a verbosity level
+
+V(0): must-see logs, errors
```
swatisehgal:
```suggestion
V(0): high-signal operational logs and errors always visible at default verbosity
```

pkg/driver/dra_hooks.go
```diff
@@ -224,7 +224,7 @@ func (cp *CPUDriver) createCPUDeviceSlices() [][]resourceapi.Device {
 
 // PublishResources publishes ResourceSlice for CPU resources.
 func (cp *CPUDriver) PublishResources(ctx context.Context) {
-	logger := klog.FromContext(ctx).WithValues("deviceMode", cp.cpuDeviceMode, "groupBy", cp.cpuDeviceGroupBy)
+	logger := klog.FromContext(ctx).WithValues("opID", generateShortID(opIDLen), "deviceMode", cp.cpuDeviceMode, "groupBy", cp.cpuDeviceGroupBy)
```
swatisehgal, line 227: Perhaps this is because I am still getting familiar with the repo and logging flow, but I am not fully grasping the opID concept yet. Is the intention for this to act as a correlation/tracing identifier for an operation flow across related log entries?

swatisehgal (commented):
I am still getting familiar with this repo and its logging patterns, so some of my comments may be missing project-specific context. That said, I do have some experience working in the contextual logging/observability area in Kubernetes, so I mainly focused my review on logger propagation, verbosity usage, and flow correlation patterns.

pkg/driver/dra_hooks.go
ffromani: the problem is the footgun here, which I tried to explain in the comment above. If we miss now to reinject the logger, it can get missed in the future when we actually need to consume the logger here.
I'm torn what's the best move here: adding log-specific e2e tests doesn't seem trivial, ~adding~ enriching the context now is actually unnecessary.

docs/dev/logging.md
ffromani: They should follow in the "normal operation" because these are part of the core value proposition of this driver

docs/dev/logging.md
ffromani: reads better, will change accordingly

pkg/driver/dra_hooks.go
ffromani: yes, this is the intent. Happy to clarify further the initial doc I sketched and/or to commit message

pkg/driver/dra_hooks.go
ffromani: I guess that all in all we can remove both lines, they seem to create more problem that they want to solve. hopefully future reviews and tests will disarm this trap.

ffromani: Or, *maybe*, *maybe* I can fix it differently, we need some more changes in this area anyway. I can remove from here and try to reintroduce cleanly in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/137#issuecomment-4438118647

swatisehgal: Thanks, that makes sense. I understand the concern around future consumers calling klog.FromContext(ctx) and silently missing the enriched logger if the propagation step is forgotten later. My confusion mainly came from creating the derived context and immediately discarding it, since it read a bit like dead code. 

The comment clarifies the rationale well enough for me, so I am okay leaving it as is.

ffromani: sure! I'll try to improve nevertheless in the next (and last) contextual logging PR.

fmuyassarov:
> 
> all good :), I will go through it today

Sorry, but I couldn’t finalize my review yesterday. Hopefully I’ll complete it today.

pravk03:
Sorry for the delay. I plan to review this tomorrow.

ffromani:
NP folks, not urgent :)

pkg/driver/dra_hooks.go
pravk03: I had the same question until i read this conversation. 

nit: Consider adding a TODO in the comment to make the intent more obvious 

Maybe something like - 

```
// TODO: We are not yet passing this context to any downstream calls.
// When we add calls that take 'ctx', change this to `ctx = klog.NewContext(ctx, logger)`.
```

Please  feel free to resolve this if you disagree.

test/e2e/contextual_logging_test.go
```diff
@@ -0,0 +1,120 @@
...
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	ctxlogFxt.Log.Info("using worker node", "nodeName", targetNode.Name)
+
+	dracpuPod, err := e2epod.GetDRACPUPod(ctx, ctxlogFxt.K8SClientset, targetNode.Name)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+	logs, err := e2epod.GetLogs(ctx, ctxlogFxt.K8SClientset, dracpuPod)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot get logs from the dracpu pod: %v", err)
+
+	if envVal, ok := os.LookupEnv("DRACPU_E2E_DUMP_RAW_LOGS"); ok {
```
pravk03, line 65: nit: can be documented in `test/e2e/README.md`

pravk03 (commented):
Thanks @ffromani. This looks super clean! Introducing opID tracing in the logs is a great addition that will make debugging so much easier.

/lgtm
/hold

fmuyassarov (commented):
Thanks @ffromani 
I was stumbled upon same thing as in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/137#discussion_r3234262305 but I got the intention eventually from the discussion. Otherwise, pretty good IMO.
/lgtm

test/e2e/contextual_logging_test.go
ffromani: absolutely, will add to the README.md!

pkg/driver/dra_hooks.go
ffromani: I can't ignore the fact that 2/2 reviewers found this confusing. We need a third PR anyway to do some final cleanup and remove some `klog` calls we can remove for better clarity and proper application code isolation. Therefore I will *remove* these changes from this PR and either
1. reintroduce with the recommended docs in the last PR, or
2. find a cleaner way to resolve the same issue with some (minimal) refactoring/extra code.

ffromani:
thanks folks. Even if I got LGTM, the fact each reviewer had to stop and pause on the same point, even if that point was explained, deserves more attention. I need to fix a couple areas anyway, so I'll resubmit and implement https://github.com/kubernetes-sigs/dra-driver-cpu/pull/137#discussion_r3253230131

ffromani:
/hold cancel

resubmitted as promised, so LGTM is gone, no longer need to hold

swatisehgal:
Thanks for addressing the review comment, LGTM. The only (non-blocking) pending item from my side is the one about updating the log level for container lifecycle events. Just want to make sure that we are intentionally deferring that and not inadertantly missing the update.

ffromani:
> Thanks for addressing the review comment, LGTM. The only (non-blocking) pending item from my side is the one about updating the log level for container lifecycle events. Just want to make sure that we are intentionally deferring that and not inadertantly missing the update.

I was actually wrestling with this idea and can't make my mind. I guess that considering we're around 0.2.0 we can bump these log levels indeed.

swatisehgal (commented):
/lgtm

ffromani:
I think I have addressed all the points now. If I missed something I'd rather address them in the last (and final, hopefully) contextual logging PR.
The main goal of that PR is remove `klog.FromContext/NewContext` - logr has alternatives. The idea is that klog should be consumed only at the plumbing layer; the application code should use `logr`.

PR #138, opened 2026-05-06
e2e: extract node selection helper

ffromani:
We are starting to have boilerplate proliferation in three calls sites, so it's time to evaluate code extraction.

Production code wants to be DRY, and same applies for test *infrastructure*, not test *logic*. This commit wants to consolidate the setup plumbing, which should be shared and ginkgo agnostic.

OTOH, test assertions or test scenarios still should actually prefer duplication for readability and independence. Coupling test logic through abstractions and DRYness often turns out to be painful in the long term.

Bonus: fix a TODO in the old node selection logic: the worker node pick is actually random now, even though the random choice is not crypto strong by design, which makes little sense in this test context.

### DIFF
--- a/test/pkg/node/node.go
+++ b/test/pkg/node/node.go
@@ -18,10 +18,15 @@ package node
 
 import (
 	"context"
+	"math/rand/v2"
+	"os"
+	"time"
 
+	"github.com/go-logr/logr"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	k8swait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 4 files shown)


test/pkg/node/node.go
```diff
@@ -50,3 +55,29 @@ func IsReady(node *v1.Node) bool {
...
+	var node *v1.Node
+	immediate := false // shortcut to make the call more readable
+	err := k8swait.PollUntilContextTimeout(ctx, interval, timeout, immediate, func(ctx context.Context) (bool, error) {
+		workerNodes, err := FindWorkers(ctx, cs)
+		if err != nil {
+			return false, err
+		}
+		if len(workerNodes) == 0 {
+			return false, fmt.Errorf("no worker nodes detected") // should never happen
+		}
```
fmuyassarov, line 82: Looks good and totally makes sense. The only minor thing I have in mind is, `FindWorkers()` that is called returns only ready nodes. I'm wondering if there is a chance that during some short window node is still in notReady state which means we get nil from `FindWorkers()` and thus fail here.

fmuyassarov (commented):
Looks pretty good. Left a single question inline.

test/pkg/node/node.go
ffromani: Thanks @fmuyassarov , you raise a good point. I was wrestling with the linter here and managed to convince myself the linter was right and the error should bubble up. But the more I think about it, the more I believe I made the wrong call. Your point is correct, not the linter's. I'll flip this `if` because it has a real chance to increase flakes unnecessarily.

ffromani:
/hold cancel

took another pass at `PickWorker` in its entirety, good to review again (not urgent)

fmuyassarov (commented):
Thanks @ffromani .
/lgtm

PR #139, opened 2026-05-06
Check for overlapping device allocation in individual mode

### LINKED ISSUE #44: Evaluate impact of https://github.com/kubernetes/kubernetes/pull/136269 on this driver
Per recommendations emerged in wg-device-management meeting on Jan 20, 2026

pravk03:
Adds a check to ensure that CPUs requested in individual mode are not already allocated to an existing claim

Fixes: #44

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -399,6 +399,14 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 	}
 
 	claimCPUSet := cpuset.New(claimCPUIDs...)
+	// All the CPUs allocated to a claim should currently be in the shared pool.
+	sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
+	if !claimCPUSet.IsSubsetOf(sharedCPUs) {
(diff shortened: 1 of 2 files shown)


pkg/driver/dra_hooks.go
```diff
@@ -399,6 +399,14 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 	}
 
 	claimCPUSet := cpuset.New(claimCPUIDs...)
+	// All the CPUs allocated to a claim should currently be in the shared pool.
+	sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
+	if claimCPUSet.Difference(sharedCPUs).Size() > 0 {
```
ffromani: can we use `IsSubsetOf` here? using that API I think the guard would become super-obvious

ffromani (commented):
thanks, almost LGTM! just one possible improvement inside

pkg/driver/dra_hooks.go
pravk03: Yes, this is better. Updated.

ffromani:
/lgtm

thanks for the quick update!

PR #144, opened 2026-05-12
chore: follow up improvments to helm

fmuyassarov:
This is a follow up patch to the Helm PR (https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83) and addresses https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83#pullrequestreview-4273569446.


ffromani (commented):
/approve
/lgtm

thanks for the quick update

PR #68, opened 2026-02-23
WIP: consume and expose the PCIERoot attribute

ffromani:
TODO: before unlift the WIP:
- try to gather a real more complex topology using the bundled `hack/pciescan`. If I can get that, build more complex tests at least
- can we exercise the PCIe paths on CI? if not that's a quite big underwater stone anyway
- should we emphasize more the experimental nature of the attributes? (but if so, are we implicitly committing the OTHER attributes to be stable? Not sure that was ever the intention. What's the actual DRA best practice here?)

---

handle the pcieroot attribute.
The mindset shift handling this attribute and the questions it raises call for a more elaborate discussion. This PR includes in-deep documents which trace the behavior of the linux kernel, how it exposes topology data, how we can leverage and how it gets in the way. This give the foundation for the decisions made in this PR.

As for the implementation, we do that in stages:
- add machinery to discover CPUs local to a pcieroot. There's no forward mapping CPU->devices, so we do that backwards: enumerate PCI devices, find host bridges, query the kernel for CPUs local to those
- add new *experimental* grouping mode: group cpus by their PCIe root affinity, like we do for NUMA nodes and sockets
- expose the PCIe root attribute in individual mode.

disclosure: this PR sneaks in a couple of QoL devex fixes. Can split them off if needed.

there are still open points (not in priority order):

- list attributes: use a custom attribute (current choice) or just override the standard one with a list value? what's the kube-recommended flow?
- resource modelization: do we need shared counters (KEP 5941)?
- resource modelization: not too happy with the current approach, but in grouped mode we should probably rethink how we expose L3/LLC domains in addition to PCIe root domains. IOW, we probably need to rethink the model entirely, which is a bigger endeavour.
We're at release 0.2.0 and this is a concern for the cpumem combined driver, mostly.

Overall, I think this PR brings value to the project. The road ahead is not clear, but this very project is the right place for this kind of experimentation. The feature composability, the intrinsic opt-in nature of attributes and the declared maturity stage of the project provide enough safety to enable such experiments, which are vital to let a good solution eventually emerge.

PR #146, opened 2026-05-13
Rebase to Kubernetes v1.36

pravk03:
Rebasing on top of Kubernetes 1.36 to get the new DRA changes. This unblocks experimentation with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/68 and [KEP-5517](https://github.com/kubernetes/enhancements/pull/5755).

Currently, we rely on pulling precompiled kind node image from docker hub. In this PR, we switch to building the node image locally. This decouples our test infrastructure from Docker Hub release delays (like we are having with 1.36)  and enables early testing against new Kubernetes releases.

### DIFF
--- a/.github/workflows/helm-e2e.yaml
+++ b/.github/workflows/helm-e2e.yaml
@@ -28,10 +28,14 @@ jobs:
       - name: Build the image
         run: make build-image
 
+      # Try to parse KIND_K8S_VERSION from the Makefile and use it to build the node image and create the cluster.
       - name: Create kind cluster
         run: |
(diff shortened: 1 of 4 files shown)


pravk03:
cc @ffromani @fmuyassarov

ffromani:
/lgtm

this is a pragmatic choice. We will likely need to keep for a good while till the kubernetes API side stabilizes.
My only concern was the build time in CI but glancing at logs it seems much faster than I anticipated.
thanks @pravk03 !

PR #148, opened 2026-05-19
finalize contextual logging support

### LINKED ISSUE #118: Switch to contextual and structured logging
Following the process core kubernetes has begun, we should switch from unstructured klog.Infof logging to structured logging and to contextual logging. Contextual and structured logging are independent from each other but together the greatly improve the troubleshooting experience.

We have abundance of tooling and lessons learned from core kubernetes transition, and the bulk of the porting process is mechanical and easy (and safe) to automate, also using LLMs.

ffromani:
Finalizes the port to contextual logging.
Port `cpuinfo` to contextual logging, which missed the previous update train; update the logging guidelines.
Isolate `klog` calls cleanly - our application code should reason in terms of `logr`, and `klog` should only used in plumbing, or where the few and rare exception (documented) warrant.

Finally, address the concern raised in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/137#issuecomment-4470689688 about the footgun of missing the enriched logger.

Builds on top of #129 #137 
Closes: #118

### DIFF
--- a/docs/dev/logging.md
+++ b/docs/dev/logging.md
@@ -9,6 +9,32 @@ V(6): debug messages, very fine details, mostly for developers
 
 V(7) or more is reserved for future usage.
 
+## when to use Error()
+
+Summary: use Error() if we need to log an unrecoverable error which leads
+to a failed state, not to a degraded-but-operational state.
+
+Description: sometimes a code flow returns an error we want to log.
+How to log it depends on what the error means.
+
+If the error is *recoverable*, and the flow can continue,
+the log should represent a non-fatal, handled situation and we prefer to use
+something like
+
+```
+logger.Info("descriptive message", "key", value, "err", error)
+```
+
+note the error is always the last pair.
+
+If the error is *unrecoverable* and represents an actionable failure, then
+first and foremost the flow must fail, then the error should guide
+the log-reader towards resolution. We prefer to use something like
+
+```
+logger.Error(err, "descriptive message")
+```
+
 ## the opID tag
 
 `opID` is a cheap trace identifier similar in spirit to OTEL's
--- a/internal/ctxlog/ctxlog.go
+++ b/internal/ctxlog/ctxlog.go
@@ -1,0 +1,116 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ctxlog
+
+import (
+	"context"
+	"flag"
+	"log"
+	"runtime/debug"
+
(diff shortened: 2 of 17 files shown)


pkg/cpumanager/cpu_assignment_test.go
```diff
@@ -28,6 +28,8 @@ import (
 
 // NOTE: This file is a copy of https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/cpumanager/cpu_assignment_test.go
 // as of commit https://github.com/kubernetes/kubernetes/commit/fd5b2efa76e44c5ef523cd0711f5ed23eb7e6b1a with minor modifications.
+// NOTE: about contextual logging: different from the other parts of the codebase, we *intentionally* keep klog references
+// to minimize the changes wrt the kubernetes codebase till (if) we decide to fully fork it.
 
 func TestCPUAccumulatorFreeSockets(t *testing.T) {
 	logger := klog.Background()
```
AutuSnow, line 35: Should we use testr. New (t) instead of klog. Background() in this place to maintain consistency

ffromani: Thanks for the comment! There's a tension here between project standards (prefer `testr.New`) and the fact we're borrowing code from kubelet. I tried to explain the reasoning in the comment just above. This file is a near-copy of upstream k8s.io/kubernetes code, and we intentionally keep klog here to minimize the diff against upstream. If we ever decide to fully fork the cpumanager library code we use, then yes, we should move to `testr.New`. Makes sense?

AutuSnow: Thank you for your answer. I don't think there's any problem now

AutuSnow:
/lgtm
Thanks！

internal/ctxlog/ctxlog.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// MarshalLog implements logr.Marshaler.
+func (ref ObjectRef) MarshalLog() interface{} {
+	return ref
+}
+
+// KObj returns an ObjectRef for the given object, for use as a log value.
+func KObj(obj KMetadata) ObjectRef {
+	return ObjectRef{
```
pravk03, line 118: nit: should we have a nil check here ?

docs/dev/logging.md
```diff
@@ -9,6 +9,32 @@ V(6): debug messages, very fine details, mostly for developers
 
 V(7) or more is reserved for future usage.
 
+## when to use Error()
+
+Summary: use Error() if we need to log an unrecoverable error which leads
```
pravk03, line 14: nit: I just grepped for Error() and noticed a couple of places in nri_hooks.go where we are using it for recoverable errors. We can change those to Info() based on this definition.

pravk03 (commented):
This PR looks great !!. Thanks for standardizing the logs and decoupling from klog. Left a couple of minor suggestions, but they are optional. Since this is already merged, we can easily include these comments in a future change if you agree with these.

internal/ctxlog/ctxlog.go
ffromani: Interesting point. Thing is: it's a glorified copy-paste of klog code. I think being 1:1 equal has the highest ROI, but I'm open to talk here.

docs/dev/logging.md
ffromani: good catch, somehow missed that. I'll try to sneak into a clean up. Or can be a good first issue!

Issue #103, opened 2026-03-31
Support mixed shared and isolated CPU allocation with CPU DRA


SchSeba:
## Summary
Support a mixed CPU allocation mode where a Pod can request:
- isolated CPUs through CPU DRA
- additional fractional CPU from the regular shared CPU pool
A concrete example is:
- Pod CPU request/limit: `5.5`
- CPU DRA request: `5`
Expected result:
- `5` isolated CPUs are allocated through CPU DRA
- the remaining `0.5` CPU stays available for non-critical application threads
## Motivation
High-performance workloads such as DPDK applications often need a fixed set of isolated CPUs for dataplane threads, but still need a small amount of non-dedicated CPU for helper, control-plane, logging, or housekeeping threads.
Today, the common workaround is to over-request dedicated CPUs and manage thread placement manually. That wastes isolated capacity and makes the deployment harder to express and automate.
## Requested behavior
When a Pod combines a fractional CPU request with an integer CPU DRA request, the implementation should:
- allocate the integer DRA portion as isolated CPUs
- keep the fractional remainder outside the dedicated CPU set
- avoid forcing all workload threads onto the isolated CPUs
- preserve the dedicated CPUs for the latency-sensitive part of the workload
Using the example above, a Pod that requests `5.5` CPUs and `5` CPUs through CPU DRA should end up with:
- `5` isolated CPUs for the dataplane / dedicated threads
- `0.5` CPU worth of regular shared runtime for non-critical threads
## Application-facing API / discoverability
The implementation should also expose the allocated isolated CPU set to the Pod using the existing DRA application-facing mechanism, so user space applications can discover which CPUs are dedicated without relying on out-of-band configuration.
That data should be usable by applications such as DPDK to:
- pin dataplane threads to the dedicated CPU set
- keep helper / control threads off the dedicated CPUs
- document a recommended way to launch an application so non-critical threads do not consume the isolated cores
## Why this matters
This would make it possible to express a very common real-world pattern:
- dedicated CPUs for the hot path
- a small shared CPU budget for everything else
That improves utilization of isolated CPUs and reduces the need for manual cpuset handling in the application.
## Acceptance criteria
- a Pod requesting `5.5` CPUs and `5` CPUs through CPU DRA gets exactly `5` isolated CPUs plus `0.5` CPU from the shared pool
- the isolated CPU set is exposed to the application through the existing DRA application-facing path
[...]

PR #149, opened 2026-05-20
[Do Not Merge] POC for Mixed CPU Allocation

pravk03:
POC for https://github.com/kubernetes-sigs/dra-driver-cpu/issues/103

This PR introduces Mixed Allocation Mode to the CPU DRA driver, enabling workloads to request both exclusive and shared CPUs. 

Key Principles: 
* Any request that goes in the standard container spec is considered as a request for shared CPUs.
* Any claim request is considered as a request for guaranteed exclusive CPUs.
* Relies on [KEP-5517](https://github.com/kubernetes/enhancements/issues/5517) to allow for correct scheduler accounting.

Flags:
* Added `--mixed-allocation-mode` flag to enable or disable mixed allocation mode.
* Added `--disable-node-allocatable-mapping` flag to toggle native KEP-5517 scheduler accounting.

DRA Hook:
* Adds a native NodeAllocatable resource mapping to published ResourceSlices. This allows kube-scheduler to track and subtract DRA based CPU requests directly from node allocatable capacity.

NRI Hook:
* Mixed Allocation Mode Implementation:
  - Configured the NRI plugin to inspect container CPU shares at runtime.
  - Containers with shares > 2 are automatically identified as running in mixed mode. Kubelet sets container shares to exactly 2 when standard CPU requests are omitted, as defined in [helpers_linux.go](https://github.com/kubernetes/kubernetes/blob/d0ab3fc75768ae8081bff58d7050be045eb7ec2e/pkg/kubelet/cm/helpers_linux.go#L44).
  - Configures the CPU affinity mask of mixed containers to the union of their dedicated exclusive CPUs and the node's shared CPU pool.
  - Shrinks the shared CPU pool of other standard containers by subtracting exclusive CPUs.


ffromani:
/cc

I'm aware this it is a POC and about all the caveats, still very interested in this work! thanks for posting

pravk03:
I will continue exploring this, but the progress is likely going to be slow as the v1.37 deadlines approach. Additionally, this work depends on [KEP-5517](https://github.com/kubernetes/enhancements/pull/5755) reaching Alpha 2 (hopefully in 1.37) so that Kubelet can set safe pod-level cgroup boundaries and container level limits to include DRA.

PR #150, opened 2026-05-20
use helm in CI flows

ffromani:
We prefer helm to handle the manifests, **but we will still keep the legacy manifests flow for 0.2.0**.
To streamline the flows, exercise the helm flow more and to gain more confidence in the helm flow, we should
use helm in our prow CI environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,10 +147,13 @@ kind-load-image: build-image  ## load the current container image into kind
 	kind load docker-image ${IMAGE} ${IMAGE_LATEST} --name ${CLUSTER_NAME}
 
 kind-uninstall-cpu-dra: ## remove cpu dra from kind cluster
-	kubectl delete -f dist/install.yaml || true
+	helm uninstall dra-driver-cpu --namespace kube-system || true
 
-kind-install-cpu-dra: kind-uninstall-cpu-dra build-image kind-load-image ## install on cluster
-	kubectl apply -f dist/install.yaml
+kind-install-cpu-dra: kind-uninstall-cpu-dra build-image kind-load-image ## install on cluster mimicking production settings
+	helm install dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+		--set image.repository=${REGISTRY}/${STAGING_REPO_NAME}/${IMAGE_NAME} \
+		--set image.tag=${TAG} \
+		--set image.pullPolicy=IfNotPresent
 
 delete-kind-cluster: ## delete kind cluster
 	kind delete cluster --name ${CLUSTER_NAME}
--- a/Makefile
+++ b/Makefile
@@ -73,13 +73,13 @@ $(OUT_DIR):  ## creates the output directory (used internally)
 
 # get image name from directory we're building
 CLUSTER_NAME=dra-driver-cpu
-STAGING_REPO_NAME=dra-driver-cpu
 IMAGE_NAME=dra-driver-cpu
 # docker image registry, default to upstream
 REGISTRY ?= us-central1-docker.pkg.dev/k8s-staging-images
 # this is an intentionally non-existent registry to be used only by local CI using the local image loading
 REGISTRY_CI := dev.kind.local/ci
-STAGING_IMAGE_NAME := ${REGISTRY}/${STAGING_REPO_NAME}/${IMAGE_NAME}
+IMAGE_REPO := ${REGISTRY}/${IMAGE_NAME}/${IMAGE_NAME}
+IMAGE_REPO_CI := ${REGISTRY_CI}/${IMAGE_NAME}
 # tag based on date-sha
 GIT_VERSION := $(shell date +v%Y%m%d)-$(shell git rev-parse --short HEAD)
 ifneq ($(shell git status --porcelain),)
@@ -87,10 +87,10 @@ ifneq ($(shell git status --porcelain),)
 endif
 TAG ?= $(GIT_VERSION)
 # the full image tag
-IMAGE_LATEST?=$(STAGING_IMAGE_NAME):latest
-IMAGE := ${STAGING_IMAGE_NAME}:${TAG}
-IMAGE_CI := ${REGISTRY_CI}/${IMAGE_NAME}:${TAG}
-IMAGE_TEST := ${REGISTRY_CI}/${IMAGE_NAME}-test:${TAG}
+IMAGE_LATEST?=$(IMAGE_REPO):latest
+IMAGE := ${IMAGE_REPO}:${TAG}
+IMAGE_CI := ${IMAGE_REPO_CI}:${TAG}
+IMAGE_TEST := ${IMAGE_REPO_CI}-test:${TAG}
 # target platform(s)
 PLATFORMS?=linux/amd64,linux/arm64
 LOCAL_PLATFORM?=linux/$(ARCH)
@@ -111,7 +111,7 @@ HELM_COMMON_ARGS := \
 # required to enable buildx
 export DOCKER_CLI_EXPERIMENTAL=enabled
 image: ## docker build load
-	docker build . -t ${STAGING_IMAGE_NAME} --load
+	docker build . -t ${IMAGE_REPO} --load
 
 build-image: ## build image
 	docker buildx build . \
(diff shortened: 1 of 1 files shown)


ffromani:
/cc @pravk03 @AutuSnow 

this actually started because I wanted -v=6 in CI to finalize my own #68

Makefile
```diff
@@ -199,34 +176,31 @@ endif
...
-	$(call kind_setup,$(CI_MANIFEST_FILE))
+	kind create cluster --name ${CLUSTER_NAME} --image=kindest/node:$(KIND_K8S_VERSION) --config hack/ci/kind-ci.yaml
+	kubectl label node ${CLUSTER_NAME}-worker node-role.kubernetes.io/worker=''
+	kind load docker-image --name ${CLUSTER_NAME} ${IMAGE_CI} ${IMAGE_TEST}
+	helm install dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+		--set image.repository=${REGISTRY_CI}/${IMAGE_NAME} \
+		--set image.tag=${TAG} \
+		--set image.pullPolicy=IfNotPresent \
+		--set args.cpuDeviceMode=${DRACPU_E2E_CPU_DEVICE_MODE} \
+		--set-string args.reservedCPUs=${DRACPU_E2E_RESERVED_CPUS}
```
ffromani, line 206: TIL about `--set` vs `--set-string`

ffromani:
CI failures seem unrelated, looking into them

Makefile
```diff
@@ -197,17 +174,34 @@ endif
...
+ifeq ($(OVERRIDE_IMAGE),true)
+	helm template dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+		--set podLabels.build=${GIT_VERSION} \
+		--set image.repository=${STAGING_IMAGE_NAME} \
+		--set image.tag=${TAG} \
+		--set image.pullPolicy=IfNotPresent \
+		> dist/install.yaml
+else
+	helm template dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+		> dist/install.yaml
```
fmuyassarov, line 190: nit, we could drop the dist target because helm template with `--output-dir` flag will create the directory if that doesn't exist already

ffromani:
CI failures are caused by the `ctxlog` tests which do a too naive parsing of logs, a fix will be uploaded in the next few hours.

fmuyassarov (commented):
Thanks @ffromani. I think this is the right move, and thanks for taking care of it.
I have one question though. According to `kind-install-cpu-dra`, we are using Helm only for templating here, not for the actual installation, and this is intentional, right?
I can see the point, especially since this is mainly for CI. There is probably no need for Helm state tracking, upgrades, etc. If we had tests that needed to verify upgrade scenarios, then it would make sense to use Helm for the driver installation as well. But for this use case, it seems safe not to rely on it.

fmuyassarov:
Forgot to mention, one ask  from me would be, could you please update https://github.com/kubernetes-sigs/dra-driver-cpu/blob/ee2570069f5a124f66ded0867688145194cd124e/Makefile#L247 and https://github.com/kubernetes-sigs/dra-driver-cpu/blob/ee2570069f5a124f66ded0867688145194cd124e/Makefile#L259 to use the same variable for the chart path?

ffromani:
> CI failures are caused by the `ctxlog` tests which do a too naive parsing of logs, a fix will be uploaded in the next few hours.

fix: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/151

fmuyassarov:
> > CI failures are caused by the `ctxlog` tests which do a too naive parsing of logs, a fix will be uploaded in the next few hours.
> 
> fix: #151

that was quick :)

Makefile
```diff
@@ -197,17 +174,34 @@ endif
...
+	kubectl label node ${CLUSTER_NAME}-worker node-role.kubernetes.io/worker=''
+	kind load docker-image --name ${CLUSTER_NAME} ${IMAGE_CI} ${IMAGE_TEST}
+	helm install dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+		--set image.repository=${REGISTRY_CI}/${IMAGE_NAME} \
+		--set image.tag=${TAG} \
+		--set image.pullPolicy=IfNotPresent \
+		--set args.logLevel=6 \
+		--set args.cpuDeviceMode=${DRACPU_E2E_CPU_DEVICE_MODE} \
+		--set-string args.reservedCPUs=${DRACPU_E2E_RESERVED_CPUS}
+	hack/ci/wait-resourcelices.sh
```
AutuSnow, line 207: I am considering whether Helm can be used in this place with built-in -- wait -- timeout=xxx

ffromani: I'd be open to that. In my #68 I'm adding a bit of debug logs if the wait fails though, and I'd like to keep that. I think using the `wait-resourcelices.sh` helper is a bit simpler - perhaps renaming `ensure-resourceslices.sh`. For this reason I'd defer the switch to helm --wait for a bit more.

ffromani:
> Thanks @ffromani. I think this is the right move, and thanks for taking care of it. I have one question though. According to `kind-install-cpu-dra`, we are using Helm only for templating here, not for the actual installation, and this is intentional, right?

Thanks for catching! that was actually an oversight. It may indeed be the right call, but right now it was accidental. Let me reason about it. I'll address your other comments meanwhile.

Makefile
AutuSnow: I agree, this issue is not a blocking problem, it can only be used as a hint

Makefile
ffromani: Good idea! But there's a tension here. We need a single concatenated YAML for practicality (and to avoid unnecessary backward compat breakages), and IIUC `--output-dir` creates split files, while using stdout creates the concatenation we need. 
So if I use `--output-dir` I'd need to concatenate the files later - not great. OTOH if I use a trivial redirect I still need to ensure `dist/` exists before we run the command. Overall I'm sticking with keeping the `dist` prerequisite.

fmuyassarov:
> > Thanks @ffromani. I think this is the right move, and thanks for taking care of it. I have one question though. According to `kind-install-cpu-dra`, we are using Helm only for templating here, not for the actual installation, and this is intentional, right?
> 
> Thanks for catching! that was actually an oversight. It may indeed be the right call, but right now it was accidental. Let me reason about it. I'll address your other comments meanwhile.

It would be also good to rebase so that we get the fix for the CI too.

Makefile
fmuyassarov: You are right, for some reason I thought `--output-dir` will do it, but apparently it creates just a directory. So, please ignore my comment and better to keep as it is.

ffromani:
updated, all the changes are now intentional and with a rationale in the commit message(s).

AutuSnow:
For me, I have already reached LGTM

Makefile
```diff
@@ -145,43 +147,22 @@ kind-load-image: build-image  ## load the current container image into kind
...
+	helm uninstall dra-driver-cpu --namespace kube-system || true
 
-kind-install-cpu-dra: kind-uninstall-cpu-dra build-image kind-load-image ## install on cluster
-	kubectl apply -f dist/install.yaml
+kind-install-cpu-dra: kind-uninstall-cpu-dra build-image kind-load-image ## install on cluster mimicking production settings
+	helm install dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+		--set image.repository=${IMAGE_REPO} \
+		--set image.tag=${TAG} \
+		--set image.pullPolicy=IfNotPresent \
+		--set args.logLevel=2
```
fmuyassarov: Can I ask why do we want to reduce the log level from 4 (default) to 2 ?

Makefile
```diff
@@ -197,17 +178,34 @@ endif
 		> dist/install.yaml
 	@rm dist/*.part.yaml
 
-ci-manifests: install-yq ## create the CI install manifests
-ifneq ($(DRACPU_E2E_VERBOSE),)
-	@echo "setting up manifests for mode=$(DRACPU_E2E_CPU_DEVICE_MODE)"
+manifests: dist ensure-helm ## create the install manifest
```
fmuyassarov, line 180: I understand we’re keeping the legacy `manifests-legacy` target during the transition, but since the new installation method is now just a one liner Helm command, do we still need to keep this target?

fmuyassarov (commented):
Looks great. But I have two questions to ask.

Makefile
ffromani: actually no real strong reason. Let's use the defaults coherently with grouped-by and device-mode.

Makefile
ffromani: I'm inclined to say yes at least for 0.2.0 as transitional bridge and as reference about how to run helm - because we came from a hand-rolled scenario. I'm not too attached to this decision though.

fmuyassarov: Okay, I don't think it is a blocker nor will it hurt by any means. Let's keep it then for some time.

fmuyassarov (commented):
Thank you @ffromani . 
/lgtm

PR #151, opened 2026-05-20
ctxlog: e2e: handle nested quotes

ffromani:
the `ctxlog` tests want to do basic sanity test about our contextual logging.
Lacking better libraries to do the parsing, the test infra reimplements a minimal parser to extract just enough to do our validation.

The internal KV splitting logic is based on a simple regex which broke with nested quoting, creating false positive. Rather than doubling down on regexes or create a more complex parser (overkill, still), we implement a filter based on golang 1.22+ stdlib.

The key reason is we don't care about nested quotes anyway, so it's pointless to try to parse them.

**VERY IMPORTANT NOTE**
why did it fail on #150 and not before? because the test order is not deterministic, and the ctxlog tests only run predictably at the end. Both facts are intentional and arguably a feature.
Some tests will cause the `dracpu` driver to restart (notably: "NRI Reconciliation on Restart"), which may lead some logs to be lost (in the sense of not validated), this is why a false positive can take some time to surface

### DIFF
--- a/test/pkg/logcheck/ctxlog.go
+++ b/test/pkg/logcheck/ctxlog.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -67,7 +68,12 @@ func logSeverityFromRawLine(rawLine string) string {
 }
 
 type ParsedLog struct {
-	lines []parsedLogLine
+	// OpaqueKeys tell the log parser which keys it should avoid to process.
+	// The reason for not processing keys is that we want to implement a minimal
+	// log parser for validation purposes. A general purpose structured log parser
+	// is a much larger endeavour which has no real usecase yet.
+	OpaqueKeys []string
+	lines      []parsedLogLine
 }
 
 // parseTextloggerPayload splits a payload line in message and kvPairs.
@@ -136,9 +142,11 @@ func (pl ParsedLog) KVPairsAt(idx int) [][]string {
 	if idx < 0 || idx >= len(pl.lines) {
 		return nil
 	}
-	// differently from other *At() method, we do lazily because
-	// this is a relatively expensive operation
-	return kvPairRE.FindAllStringSubmatch(pl.lines[idx].kvPairs, -1)
+	kvPairs := pl.lines[idx].kvPairs
+	for _, sk := range pl.OpaqueKeys {
+		kvPairs = RedactQuotedValue(kvPairs, sk)
+	}
+	return kvPairRE.FindAllStringSubmatch(kvPairs, -1)
 }
 
 type FlowEntries struct {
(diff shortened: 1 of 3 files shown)


ffromani:
/cc @fmuyassarov

since we talked about this fix in #150

test/pkg/logcheck/ctxlog.go
```diff
@@ -286,3 +299,16 @@ func FormatDuplicateKeyResult(parsedLog ParsedLog, dupKeys ...DuplicateKeyResult
...
+	before, after, found := strings.Cut(s, key+`="`)
+	if !found {
+		return s
+	}
+	quoted, err := strconv.QuotedPrefix(`"` + after)
+	if err != nil {
+		return s
+	}
+	return before + after[len(quoted)-1:]
+}
```
AutuSnow, line 316: Strconv.QuotedRefix returns a complete quoted string (such as "value") containing two quotation marks in length. To skip the entire quoted string, use len (quoted) instead of len (quoted) -1.

test/pkg/logcheck/ctxlog.go
```diff
@@ -286,3 +299,16 @@ func FormatDuplicateKeyResult(parsedLog ParsedLog, dupKeys ...DuplicateKeyResult
 	}
 	return sb.String()
 }
+
+// StripQuotedValue removes a key="..." pair from s, handling escaped quotes inside the value.
+func StripQuotedValue(s, key string) string {
+	before, after, found := strings.Cut(s, key+`="`)
```
AutuSnow: Strings.Cut only handles the first occurrence of the key. If the behavior is inconsistent and the format is incorrect multiple times, it will silently return the original string, which can lead to hidden real problems

fmuyassarov (commented):
Thanks @ffromani . LGTM from my side, but adding hold so that you have chance to take a lot at other comments above.
/hold
/lgtm

test/pkg/logcheck/ctxlog.go
ffromani: Thanks for raising the point. I think I was too clever here, or shipped too early (or both). I'm adding and removing quotation marks (\") trying to make the stripping code a bit more robust, but this made also harder to follow. The +1/-1 tricks are unnecessary and obfuscate the intent. I just removed them and simplified the code.

test/pkg/logcheck/ctxlog.go
ffromani: Yes, this is more correct. I was too focused on the occurrence at hand, but we should filter all the keys to let the downward duplication detection code actually work. We need to iterate over all the KV pairs and preprocess them. I also took the chance to redesign the code a bit to make it more intent-revealing. Thanks!

ffromani:
/hold cancel

because LGTM is gone

AutuSnow:
I feel that your modifications are no longer problematic in my opinion, and I agree.
Thank you for your modification
/lgtm

PR #155, opened 2026-05-22
chore: install helm to local bin directory

### LINKED ISSUE #145: install dependencies locally
Our flows depend on non-critical binaries which may not be in the base system. While we fail hard if golang is missing (correctly), we install automatically e.g. helm if not found in the system (helm is just an example. golangci-lint also comes to mind).

This is good and we should keep it. But we should be a bit more respectful towards the user environment:
1. install soft-dep binaries we need locally in `bin/`
2. allow user to force download of local deps, e.g. if we require a more modern version of a system tool (e.g. golangci-lint more recent than the system provides).

we will still depend on the system to provide the core hard deps like `golang` and `make`. and shell. These are too fundamental to override.

fmuyassarov:
Avoid installing binaries outside of the driver tree when developers call Helm related build targets and instead keep it local to the repository within bin directory. Note that, we aren't touching the golangci-lint, because it already ends up within the local `/bin` directory.
Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/145

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ KIND_K8S_VERSION ?= v1.36.0
 # paths
 YQ = $(OUT_DIR)/yq
 GOLANGCI_LINT = $(OUT_DIR)/golangci-lint
+HELM = $(OUT_DIR)/helm
 
 # disable CGO by default for static binaries
 CGO_ENABLED=0
@@ -147,10 +148,10 @@ kind-load-image: build-image  ## load the current container image into kind
 	kind load docker-image ${IMAGE} ${IMAGE_LATEST} --name ${CLUSTER_NAME}
 
 kind-uninstall-cpu-dra: ## remove cpu dra from kind cluster
-	helm uninstall dra-driver-cpu --namespace kube-system || true
+	$(HELM) uninstall dra-driver-cpu --namespace kube-system || true
 
 kind-install-cpu-dra: kind-uninstall-cpu-dra build-image kind-load-image ## install on cluster mimicking production settings
-	helm install dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
+	$(HELM) install dra-driver-cpu ${HELM_CHART} ${HELM_COMMON_ARGS} \
 		--set image.repository=${IMAGE_REPO} \
 		--set image.tag=${TAG} \
 		--set image.pullPolicy=IfNotPresent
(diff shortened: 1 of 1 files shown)


fmuyassarov:
/hold
seems I missed other dependencies on ensure-helm. Let me figure it out first.

fmuyassarov:
I think it is ready now. 
/hold cancel
It is quite simple but I guess it does addresses https://github.com/kubernetes-sigs/dra-driver-cpu/issues/145. Happy to improve other bits as well if there is a room for that.

/cc @ffromani @pravk03

ffromani:
/lgtm

thanks!
deferring approve to @pravk03 so he has a chance to review.

Makefile
```diff
@@ -147,10 +148,10 @@ kind-load-image: build-image  ## load the current container image into kind
 	kind load docker-image ${IMAGE} ${IMAGE_LATEST} --name ${CLUSTER_NAME}
 
 kind-uninstall-cpu-dra: ## remove cpu dra from kind cluster
-	helm uninstall dra-driver-cpu --namespace kube-system || true
+	$(HELM) uninstall dra-driver-cpu --namespace kube-system || true
 
 kind-install-cpu-dra: kind-uninstall-cpu-dra build-image kind-load-image ## install on cluster mimicking production settings
```
pravk03, line 153: Do we need `install-helm` as a dependency on this (and a few other make targets) they depend on helm ?

Makefile
```diff
@@ -267,20 +268,20 @@ helm-schema-check: helm-schema ## verify values.schema.json is up to date; fails
...
 		echo "Helm not found, installing helm@$(HELM_VERSION_SHA) ..."; \
-		go install helm.sh/helm/v3/cmd/helm@$(HELM_VERSION_SHA); \
+		GOBIN=$(OUT_DIR) go install helm.sh/helm/v3/cmd/helm@$(HELM_VERSION_SHA); \
 	fi
 
 .PHONY: helm-package
-helm-package: ensure-helm ## package helm chart for release
-	helm package ${HELM_CHART} \
+helm-package: install-helm ## package helm chart for release
+	$(HELM) package deployment/helm/dra-driver-cpu \
```
pravk03: Is it intentional to replace {HELM_CHART} here ?

fmuyassarov: nope, that is wrong. thanks for the catch

Makefile
fmuyassarov: For uninstall specifically, I’d say no, if the driver chart is already present, that’s a clear signal that the Helm binary exists. Targets related to packaging already depend on `install-helm`, since we need it for cloud build and CI. The only target that doesn’t use install-helm is `kind-install-cpu-dra`, and I’m a bit hesitant to add Helm install as a dependency there for the same reason, we don’t want to silently install binaries without explicit intent. If Helm isn’t available, the target should simply fail, which acts as a clear signal that the user needs to install it.
That said, CI is a different case there, the environment is ephemeral and doesn’t include Helm by default, so we handle it differently and do install it before running Helm commands.

pravk03:
/lgtm
/approve

Thanks !

PR #156, opened 2026-05-25
expose the pcieroot device attribute finding the roots from devices

ffromani:
Expose the pcie root attribute for devices in both grouped and individual mode, leveraging the cpu locality of PCIe devices reported by the linux kernel.
The core idea: cpus don't expose their proximity to PCIe roots, but top-level PCIe devices (like host bridges) do expose the CPUs local to them, so we can build the reverse map and figure out the mapping.

The reality turned out to be quirkier, and while the approach still works, the current representation tend to collapse towards the NUMA representation, even though on paper the pcie root locality should be finer-grained.
However, the net result is that the reported data is coarser than actually true at silicon level, but never wrong.

As for the implementation, we do that in stages:
- add machinery to discover CPUs local to a pcieroot. There's no forward mapping CPU->devices, so we do that backwards: enumerate PCI devices, find host bridges, query the kernel for CPUs local to those
- expose the PCIe root attribute in grouped and individual mode.

**special notes for the reviewers**

1. discussion is still pending around the standard `pcieRoot` attribute and how to handle the scalar/list situation. Until we reach agreement, we can only expose a custom `dra.cpu/pcieRoots` attribute. Either way, we now depend on the `DRAListTypeAttributes` FG being set to true.
1. *ARM64 support is intentionally deferred* till we gain more experience and more hw support
1. ~if we merge the overlay fs support, a followup is in order to make `pkg/cpuinfo` consume the `fs.FS` interface so it gets up to speed compared to `pkg/device` and can use the overlay package.~ we agreed to remove the overlayfs from this PR and to redesign it
1. ~first 3 commits and the last commit can be split in their own related PRs, especially the first 3. Waiting on reviewer feedback before to split~ DONE: #158 and #159

Replaces: #68 
Fixes: #114 #56

PR #158, opened 2026-05-27
CI cleanups

ffromani:
CI QoL improvements and cleanups: align and simplify the kind configuration we ship, and make sure to dump dracpu logs should the resourceslice fail to be come available when doing setup.
Spun from the review of #156


pravk03:
/lgtm 

Thanks!

PR #161, opened 2026-06-02
Allow for installation without cloning the repo.

pravk03:
Currently we need clone the entire repository to install the driver, which can be inconvenient.

This PR  publishes a mutable `0.0.0-latest` tag for helm chart alongside the commit-specific version and updates the instructions in README to use this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,13 +436,29 @@ can be deployed as follows:
 make kind-cluster
 ```
 
-The recommended way to install the driver is via the provided Helm chart:
+The recommended way to install the driver is via the provided Helm chart.
+
+If you have the repository cloned locally:
 
 ```bash
 helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system
 ```
 
 See the [Helm chart README](deployment/helm/dra-driver-cpu/README.md) for the full list of configuration options.
+
+If you want to install it directly without cloning the repository (using the OCI registry):
+
+```bash
+# Deploy the latest staging chart using the mutable 0.0.0-latest tag
+VERSION=0.0.0-latest
+helm install dra-driver-cpu oci://us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/charts/dra-driver-cpu --version ${VERSION} -n kube-system
+```
+
+If you need to pin to a specific commit instead (e.g. for reproducible developer testing), you can discover version tags (formatted as `YYYYMMDD-gitsha`) by running `skopeo` or checking [Google Cloud Artifact Registry Console](https://console.cloud.google.com/artifacts/docker/k8s-staging-images/us-central1/dra-driver-cpu/charts%2Fdra-driver-cpu).
+
+```bash
+skopeo list-tags docker://us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/charts/dra-driver-cpu
+```
 
 #### Installation via install.yaml (deprecated)
 
@@ -452,6 +468,13 @@ See the [Helm chart README](deployment/helm/dra-driver-cpu/README.md) for the fu
 ```bash
 make manifests
 kubectl apply -f dist/install.yaml
+```
+
+If you want to install the driver directly without cloning the repository, you can fetch the compiled manifest from the GitHub Releases page. Specify the desired version tag in the `VERSION` environment variable (e.g., `v0.1.0`):
+
+```bash
+VERSION=v0.1.0
+kubectl apply -f https://github.com/kubernetes-sigs/dra-driver-cpu/releases/download/${VERSION}/install.yaml
 ```
 
 ### Migrating from install.yaml to Helm
--- a/scripts/build-and-publish-image.sh
+++ b/scripts/build-and-publish-image.sh
@@ -45,7 +45,14 @@ make push-image \
 	TAG="${IMG_TAG}" \
 	PLATFORMS="linux/amd64,linux/arm64"
 
+# Push the commit-specific versioned chart tag
 make helm-push \
 	CHART_REGISTRY="${IMG_PREFIX}/charts" \
 	CHART_VERSION="${CHART_VERSION}" \
+	TAG="${IMG_TAG}"
+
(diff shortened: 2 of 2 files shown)


pravk03:
@ffromani @fmuyassarov This is an optional QOL improvement, please feel free to push back if this is not necessary.

AutuSnow:
I like the direction of this PR. Allowing users to install the driver without cloning the repository is a real usability improvement, especially for quick testing and staging deployments.
﻿
One thing I think we should be careful about is how we present the mutable `0.0.0-latest` chart tag. Since it is mutable, I would prefer to describe it explicitly as a quickstart/staging convenience, not as the default reproducible installation path. For anything used in tests, debugging, or shared instructions, the pinned `YYYYMMDD-gitsha` chart versions seem safer.
﻿
Could we also confirm that the staging Artifact Registry allows overwriting/re-pushing the same Helm chart tag? Some OCI registry setups reject reusing an existing tag, and CI may not catch that until the publish job runs.
﻿
Minor doc point: the GitHub Releases `install.yaml` link is useful, but because `install.yaml` is already marked deprecated, maybe we should keep that section clearly framed as legacy-only so it does not compete with the Helm recommendation.

README.md
```diff
@@ -287,14 +287,30 @@ can be deployed as follows:
...
 ```
 
 See the [Helm chart README](deployment/helm/dra-driver-cpu/README.md) for the full list of configuration options.
 
+If you want to install it directly without cloning the repository (using the OCI registry):
+
+```bash
+# Deploy the latest staging chart using the mutable 0.0.0-latest tag
+VERSION=0.0.0-latest
+helm install dra-driver-cpu oci://us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/charts/dra-driver-cpu --version ${VERSION} -n kube-system
```
fmuyassarov, line 454: Since this is from the HEAD, I guess it deserves a little note that it is not for official releases. Because, now the emphasize is on the avoiding the cloning but it may give an impression that this is for the official releases too, which we don't have yet + for the official releases we should use the production registry as I mentioned above.

scripts/build-and-publish-image.sh
```diff
@@ -45,7 +45,14 @@ make push-image \
...
 make helm-push \
 	CHART_REGISTRY="${IMG_PREFIX}/charts" \
 	CHART_VERSION="${CHART_VERSION}" \
 	TAG="${IMG_TAG}"
+
+# Push a mutable 0.0.0-latest versioned chart tag for easier staging deployments without cloning the repo
+make helm-push \
+	CHART_REGISTRY="${IMG_PREFIX}/charts" \
+	CHART_VERSION="0.0.0-latest" \
+	TAG="${IMG_TAG}"
```
fmuyassarov, line 58: can't we control the `CHART_VERSION` value from the test-infra to avoid duplication here?

fmuyassarov (commented):
I think this makes sense and is needed. For developers like us, the local chart will still be useful, but having a chart version that can be pulled directly would likely be faster and more convenient.

One thing not to forget (for myself as well), for the official release charts, I think it would make more sense to use the production repository rather than staging. What do you think? In other words, the local chart can remain the default for development and testing, while the production repository would be the source for charts used in official releases.

```
helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driver-cpu --version ${VERSION} -n kube-system
```
based on this

```
skopeo list-tags docker://registry.k8s.io/dra-driver-cpu/dra-driver-cpu
{
    "Repository": "registry.k8s.io/dra-driver-cpu/dra-driver-cpu",
    "Tags": [
        "sha256-0ab4277b74a88655b3a344e1939551f815d9729f971459dafef94a4642da4c71.sig",
        "sha256-2c3eedbe4c9eed789d15c48cb92ab0b81969e30582098927bc0f8bac6310a535.sig",
        "sha256-2f81cbd98adbb24e0f9cb19fef432a80eeacde839cb35af1d41c279d8530182d.sig",
        "sha256-3b1ae241c65ed73381e1518e19ef996596086ae77f201d81e8bc1d35a6caae29.sig",
        "sha256-a74bef2b40dff4e0aeabe247bd6513152d013932749f8d32c0e39e5e8fb810af.att",
        "sha256-a74bef2b40dff4e0aeabe247bd6513152d013932749f8d32c0e39e5e8fb810af.sig",
        "v0.1.0"
    ]
}
```

pravk03:
Thanks for the comments @AutuSnow @fmuyassarov. 

Re:
> One thing not to forget (for myself as well), for the official release charts, I think it would make more sense to use the production repository rather than staging

>One thing I think we should be careful about is how we present the mutable 0.0.0-latest chart tag. Since it is mutable, I would prefer to describe it explicitly as a quickstart/staging convenience, not as the default reproducible installation path

Maybe it makes sense to wait for 0.2.0 (since we are close) and then add instructions to pull the chart directly from the production repos. Maybe we can hold off on this PR. 

For local development, the current local chart instructions work fine. And for remote/clone-free installations, using official release versions from the production registry is probably the best path anyway. We can also avoid the mutable tag. WDYT ?

fmuyassarov:
> Thanks for the comments @AutuSnow @fmuyassarov.
> 
> Re:
> 
> > One thing not to forget (for myself as well), for the official release charts, I think it would make more sense to use the production repository rather than staging
> 
> > One thing I think we should be careful about is how we present the mutable 0.0.0-latest chart tag. Since it is mutable, I would prefer to describe it explicitly as a quickstart/staging convenience, not as the default reproducible installation path
> 
> Maybe it makes sense to wait for 0.2.0 (since we are close) and then add instructions to pull the chart directly from the production repos. Maybe we can hold off on this PR.
> 
+1 from my side

> For local development, the current local chart instructions work fine. And for remote/clone-free installations, using official release versions from the production registry is probably the best path anyway. We can also avoid the mutable tag. WDYT ?

+1 from my side

pravk03:
Closing this since this is already addressed by https://github.com/kubernetes-sigs/dra-driver-cpu/pull/198

Issue #114, opened 2026-04-09
NIC / CPU alignment (by supporting "resource.kubernetes.io/pcieRoot" list attribute)


everpeace:
ref: https://github.com/google/dranet/issues/92

Kubernetes v1.36 will release [KEP-5491: DRA: List Types for Attributes](https://github.com/kubernetes/enhancements/issues/5491) with alpha. Thus, supporting "resource.kubernetes.io/pcieRoot" list attribute can achieve CPU/NIC alignment as below:

```yaml
#
# NOTE: DRAListTypeAttributes feature gate is required
# 
apiVersion: resource.k8s.io/v1
kind: ResourceSlice
metadata:
  name: cpu-slice
spec:
  driver: dra.cpu
  nodeName: my-node
  devices:
  - name: cpu0
    attributes:
      # Keep grouping cpu cores by NUMA nodes
      numaNodeId:
        int: 0
      # Additional attribute
      # for pcieRoot (multiple) that each core is affine to.
      # Note: CPU core could be affine to multiple PCIe Roots.
      resource.kubernetes.io/pcieRoot:
        strings: ["pci0000:00", "pci0000:80"]
      ..and other attrs...    
---
apiVersion: resource.k8s.io/v1
kind: ResourceSlice
metadata:
  name: nic-slice
spec:
  driver: dranet.net
  nodeName: my-node
  devices:
  - name: nic0
    attributes:
      # DRANET already supports this standard attribute
      resource.kubernetes.io/pcieRoot:
        string: "pci0000:80"
      ..and other attrs...
---
apiVersion: resource.k8s.io/v1
kind: ResourceClaim
metadata:
  name: cpu-nic-alignment
spec:
  devices:
    requests:
    - name: cpu
      deviceClassName: dra.cpu
    - name: nic
      deviceClassName: dranet.net
    constraints:
    - requestNames: ["cpu", "nic"]
      # DRAListTypeAttributes feature gate can select
      # cpu and nic that have a common PCIe root address
      # (scalar value will be converted to a singleton implicitly)
      matchAttribute: "resource.kubernetes.io/pcieRoot"
```

cc/ @aojea @pravk03

ffromani:
xref: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/68#issuecomment-4074324506

not sure we can really solve without [shared consumable capacity](https://github.com/kubernetes/enhancements/pull/5942)

[...]

PR #163, opened 2026-06-03
expose the pcieroot device attribute by scanning the pci buses

ffromani:
Alternative take to #156 implementing the algo sketched in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/156#discussion_r3333727046

Expose the pcie root attribute for devices in both grouped and individual mode, leveraging the cpu locality of PCIe devices reported by the linux kernel.
The core idea: cpus don't expose their proximity to PCIe roots, but PCIe buses do expose the CPUs local to them, so we can build the reverse mapping and from there feed into attributes.

The reality turned out to be quirkier, and while the approach still works, the current representation tend to collapse towards the NUMA representation, even though on paper the pcie root locality should be finer-grained.
However, the net result is that the reported data is coarser than actually true at silicon level, but never wrong.

**special notes for the reviewers**

1. reviewers familiar with #156 will notice almost all the changes are confined in `pkg/device` and in its support code (`tools/gen-pcie-testdata/...`)
1. discussion is still pending around the standard `pcieRoot` attribute and how to handle the scalar/list situation. Until we reach agreement, we can only expose a custom `dra.cpu/pcieRoots` attribute. Either way, we now depend on the `DRAListTypeAttributes` FG being set to true.
1. *ARM64 support is intentionally deferred* till we gain more experience and more hw support


ffromani:
@fmuyassarov @pravk03 @AutuSnow 

as promised. I must say: turned even simpler than I anticipated

fmuyassarov:
Thanks @ffromani, this is on my priority list for today.

ffromani:
> Thanks @ffromani, this is on my priority list for today.

Thank you! If it helps, this is where I believe most (like 95%+) of the differences wiht #156 lie:

AI-assisted research to motivate the approach to gather HW data. Plain markdown doc.
https://github.com/kubernetes-sigs/dra-driver-cpu/pull/163/changes/97e8bf4de3e219b77c00d401e083ab06e19336bc

HW detection core change
https://github.com/kubernetes-sigs/dra-driver-cpu/pull/163/changes/4dc7f26878eda32cc8e9934eaf483d19623e5163

Simplified counterparts wrt #156. Same commit, adapted (and actually simplified) to consume the new scanning algo
https://github.com/kubernetes-sigs/dra-driver-cpu/pull/163/changes/7438c3363c1213ff41354ba3b041bb527a347c5c
https://github.com/kubernetes-sigs/dra-driver-cpu/pull/163/changes/d23eeea92bd621f18a58d338c95399b8000958dc

pkg/driver/dra_hooks.go
```diff
@@ -82,10 +84,14 @@ func (cp *CPUDriver) createGroupedCPUDeviceSlices(logger logr.Logger) [][]resour
...
-				"dra.cpu/numCPUs":    {IntValue: ptr.To(availableCPUsInSocket)},
-				"dra.cpu/smtEnabled": {BoolValue: ptr.To(cp.cpuTopology.SMTEnabled)},
+				AttributeSocketID:   {IntValue: ptr.To(int64(socketID))},
+				AttributeNumCPUs:    {IntValue: ptr.To(availableCPUsInSocket)},
+				AttributeSMTEnabled: {BoolValue: ptr.To(cp.cpuTopology.SMTEnabled)},
 			}
+			// PCIeDomain lacks a SocketID field. This is because we can't do a direct mapping
+			// using sysfs data - we will need to correlate. Therefore we can't filter domains
+			// directly like we do for NUMA nodes. Fall back to the CPU-based lookup.
+			cp.setPCIeRootsAttribute(deviceAttrs, socketCPUSet.UnsortedList()...)
```
AutuSnow: Should this use `allocatableCPUs` instead of `socketCPUSet`?
﻿
The published device capacity is based on `allocatableCPUs`, but the PCIe roots attribute is computed from all CPUs in the socket, including reserved CPUs. In an edge case where the only CPUs local to a PCIe root are reserved, the ResourceSlice could still advertise that root even though the driver cannot actually allocate CPUs from it.
﻿
Same comment applies to the NUMA-grouped path below.

pkg/driver/dra_hooks.go
```diff
@@ -117,12 +123,13 @@ func (cp *CPUDriver) createGroupedCPUDeviceSlices(logger logr.Logger) [][]resour
...
-				"dra.cpu/socketID":   {IntValue: ptr.To(socketID)},
-				"dra.cpu/smtEnabled": {BoolValue: ptr.To(cp.cpuTopology.SMTEnabled)},
-				"dra.cpu/numCPUs":    {IntValue: ptr.To(availableCPUsInNUMANode)},
+				AttributeNUMANodeID: {IntValue: ptr.To(int64(numaID))},
+				AttributeSocketID:   {IntValue: ptr.To(socketID)},
+				AttributeSMTEnabled: {BoolValue: ptr.To(cp.cpuTopology.SMTEnabled)},
+				AttributeNumCPUs:    {IntValue: ptr.To(availableCPUsInNUMANode)},
 			}
 			device.SetCompatibilityAttributes(deviceAttrs, int64(numaID))
+			cp.setPCIeRootsAttribute(deviceAttrs, numaNodeCPUSet.UnsortedList()...)
```
AutuSnow: Same concern as above: I think this should use `allocatableCPUs.UnsortedList()` rather than `numaNodeCPUSet.UnsortedList()` so the attribute describes the CPUs that this DRA device can actually allocate.

pkg/driver/driver.go
```diff
@@ -133,6 +144,27 @@ func Start(ctx context.Context, clientset kubernetes.Interface, config *Config)
 		return nil, asyncErr, fmt.Errorf("failed to get CPU topology: topology is nil")
 	}
 	plugin.cpuTopology = topo
+
+	plugin.pcieDomains, err = device.PCIeDomainsFromFS(logger, sysfs)
```
AutuSnow: This makes PCIe root discovery a driver startup dependency. Since `dra.cpu/pcieRoots` is still an auxiliary/experimental attribute, I think we should avoid failing the whole CPU driver on sysfs PCIe discovery errors unless the feature was explicitly enabled.

Could we make PCIe root publication opt-in and either skip this scan when disabled, or log-and-omit the attribute when discovery fails?

pkg/driver/dra_hooks.go
```diff
@@ -42,9 +43,10 @@ import (
 
 const (
 	// maxDevicesPerResourceSlice is the maximum number of devices that can be packed into a single
-	// ResourceSlice object. This is a hard limit defined in the Kubernetes API at
-	// https://github.com/kubernetes/kubernetes/blob/8e6d788887034b799f6c2a86991a68a080bb0576/pkg/apis/resource/types.go#L245
-	maxDevicesPerResourceSlice = 128
+	// ResourceSlice object.
+	// We use the lower "advanced features" limit because the driver
+	// may set list-type attributes (StringValues) such as PCIe roots.
+	maxDevicesPerResourceSlice = resourceapi.ResourceSliceMaxDevicesWithAdvancedFeatures
```
AutuSnow: Once PCIe roots become opt-in/default-off, we should not always use `ResourceSliceMaxDevicesWithAdvancedFeatures`. Otherwise users pay the 64-device slice limit even when no list attributes are emitted.

Can this limit depend on whether list attributes are actually enabled/published?

README.md
```diff
@@ -147,6 +147,67 @@ However, this is only a partial replacement of the corresponding CPU Manager opt
 We hardcode the NUMA split and, unlike the cpumanager feature, it won't automatically adapt if the same claim is handled by a 1-NUMA, 2-NUMA or 4-NUMA machine;
 the claim would need to be updated or recreated manually.
 
+### Exposing PCIe roots
```
AutuSnow, line 151: Could we document the operational contract here more explicitly?

This should mention that `dra.cpu/pcieRoots` uses list attributes and therefore requires `DRAListTypeAttributes=true`; that the intended safe consumption path for now is `matchAttribute` / derived attributes; and that this custom attribute does not yet provide direct interoperability with drivers exposing `resource.kubernetes.io/pcieRoot`.

AutuSnow (commented):
I like the bus-scan approach in this PR. It is simpler than #156 and avoids the complexity of scanning all PCI devices and deduplicating roots.
﻿
Before merging, I think we should tighten a few behavior/UX points: make PCIe roots opt-in/default-off, avoid failing driver startup on PCIe sysfs discovery unless explicitly enabled, compute grouped-mode roots from allocatable CPUs rather than the full CPU set, and document the DRAListTypeAttributes / matchAttribute constraints clearly.
﻿
So overall +1 on the direction,

pkg/driver/dra_hooks.go
AutuSnow: If we make PCIe roots opt-in/default-off as discussed, we may also want to keep using `ResourceSliceMaxDevices` when the feature is disabled. The current 64-device limit is correct while list attributes may be emitted, but it would be unnecessarily conservative when no list attributes are published.

ffromani:
> I like the bus-scan approach in this PR. It is simpler than #156 and avoids the complexity of scanning all PCI devices and deduplicating roots. ﻿ Before merging, I think we should tighten a few behavior/UX points: make PCIe roots opt-in/default-off, avoid failing driver startup on PCIe sysfs discovery unless explicitly enabled, compute grouped-mode roots from allocatable CPUs rather than the full CPU set, and document the DRAListTypeAttributes / matchAttribute constraints clearly. ﻿ So overall +1 on the direction,

Thanks for the review. Fair points, which in the current shape apply to both approaches. I also need to catch up with last comments and updates you reviewers requested in #156 .
Honestly I'm waiting for consensus about which direction should we pick so I can post updates just to one PR :)

pkg/driver/dra_hooks.go
ffromani: it totally should be `allocatableCPUs`, thanks for catching!

pkg/driver/dra_hooks.go
ffromani: absolutely, thanks for catching

README.md
ffromani: very fair. Updated the README.md

ffromani:
edited inline the documentation as pointed out; the opt-in behavior is implemented in the last 2 commits to minimize the changes and then the review churn.

I think that this split has merit, but I'll leave to reviewers if I should merge the changes in the parent commits in the next upload.

pkg/driver/dra_hooks.go
ffromani: good point, addressed

pkg/driver/driver.go
```diff
@@ -103,6 +107,16 @@ type Config struct {
...
+	ExposePCIeRoots  bool
+}
+
+func (cfg Config) DevicesPerResourceSlice() int {
+	if cfg.ExposePCIeRoots {
+		// We use the lower "advanced features" limit because the driver
+		// may set list-type attributes (StringValues) such as PCIe roots.
+		return resourceapi.ResourceSliceMaxDevicesWithAdvancedFeatures
+	}
+	return maxDevicesPerResourceSlice
```
pravk03: We can use resourceapi.ResourceSliceMaxDevices here and in tests and remove maxDevicesPerResourceSlice

pravk03 (commented):
Thanks. I'm still catching up on this PR and the discussions in #156, but my initial impression is that the top-down scanning from `/sys/devices/pci*` and reading CPU locality directly from the roots looks to be simpler and faster.

ffromani:
> Thanks. I'm still catching up on this PR and the discussions in #156, but my initial impression is that the top-down scanning from `/sys/devices/pci*` and reading CPU locality directly from the roots looks to be simpler and faster.

no problem at all. I think the list type convo will probably not be settled before the next WG device management mtg, so there's time to review. No rush, especially considering we're waist-deep the KEP season.

pkg/driver/driver.go
ffromani: True! I don't have preferences. @AutuSnow pointed out that's unnecessary to restrict the slice size if we don't expose list attributes, and both approaches have pros and cons. Open to chat about this :)

pkg/device/pciescan_amd64.go
```diff
@@ -0,0 +1,85 @@
...
+		roots = append(roots, dom)
+
+		logger.V(6).Info("found PCIe root", "domain", dom.String())
+	}
+
+	return roots, nil
+}
+
+// IsPCIeRootName matches the constructions of the pcie roots in the linux kernel.
+// The kernel format string is "pci%04x:%02x" (drivers/pci/probe.c:1028).
```
ffromani: nit: I should mention this applies to linux 7.0.9.

pkg/driver/driver.go
AutuSnow: Agreed with @pravk03. I think using the API constants directly is better than keeping a local `maxDevicesPerResourceSlice` alias.
﻿
The main behavior I care about is that the limit remains conditional on whether we may publish list-type attributes. So I would prefer something like:
﻿
```go
func (cfg Config) DevicesPerResourceSlice() int {
if cfg.ExposePCIeRoots {
return resourceapi.ResourceSliceMaxDevicesWithAdvancedFeatures
}
return resourceapi.ResourceSliceMaxDevices
}

Comment (ffromani) [reply pkg/driver/driver.go]: SGTM, applying the suggestion, thanks!

File: pkg/device/pciescan_amd64.go (hunk 7 above)
Comment (ffromani) [reply pkg/device/pciescan_amd64.go]: done

File: tools/gen-pcie-testdata/Dockerfile [hunk 8]
```diff
@@ -0,0 +1,22 @@
...
+RUN go mod download
+
+# build
+COPY . .
+RUN go build -ldflags "${LDFLAGS}" -o /go/bin/gen ./tools/gen-pcie-testdata/
+
+# copy binary onto base image
+FROM gcr.io/distroless/base-debian12
+COPY --from=builder --chown=root:root /go/bin/gen /gen
+CMD ["/gen"]
```
fmuyassarov, line 22: not objecting, do you expect the tool to be used as container ?

fmuyassarov (commented):
Thanks @ffromani for the great work on this. To me it looks pretty good and I had a chance to test out the latest changes on my machine (though not full fledged yet).

One thing I hope people won't miss is the feature gate. We can't easily verify the gates from within the driver and I think this transition period is going to be short depending on how the community moves with DRAListTypeAttributes.

tools/gen-pcie-testdata/Dockerfile
ffromani: Let me elaborate a bit. I won't (can't :) ) assume how ppl have access to "exotic" hardware. Sometimes they have ssh access, sometimes they access through a cloud provider, sometimes both. In case they access big iron _through_ a cloud provider first and foremost, gathering infos using containers can be simpler - this is how I had to do in some cases. And a example Dockerfile helps.

Now: admittedly the example Dockerfile is very simple, so there's a good argument to _not_ ship it to not confuse users.
So I'm questioning my own choice even if you're not objecting :) 
I'm fine both ways. If reviewers find it confusing, I can trivially drop; otherwise, we can keep it. Or I can add a small README to explain. I'm fine with all the approaches.

fmuyassarov: I think it’s fine to have it as a base for cases where running things within the container is easier. It wasn’t confusing to me what this particular Dockerfile is intended for while reviewing the patch. However, once it lands, there’s a risk that it becomes somewhat of a hidden file. Well, maybe “hidden” isn’t quite the right word, since it’s intended for developers and they can likely figure out why it’s there. I just wanted to mention that a short one line note could be helpful (like you pointed out already)not only for discoverability, but also as a reminder for us in the future about the purpose of the tool. But again, this just a nit. Feel free to drop it or if you agree add one liner along with other possible comment addresses.

ffromani: Very fair. I'll add a short README to carry this information. Good we had this thread.

ffromani:
important update: https://kubernetes.slack.com/archives/C0409NGC1TK/p1781021728981689?thread_ts=1780483228.626699&cid=C0409NGC1TK - the standard attribute is back on the PR!

pkg/device/pciescan_amd64.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/go-logr/logr"
+	"k8s.io/utils/cpuset"
+)
+
+func PCIeDomainsFromFS(logger logr.Logger, sysfs SysFS) ([]PCIeDomain, error) {
+	logger.V(6).Info("begin: processing PCIe domains")
+	defer logger.V(6).Info("end: processing PCIe domains")
+
+	entries, err := fs.ReadDir(sysfs, "devices")
```
pravk03, line 32: nit: Thinking if need to mount the PCIE paths in the daemonset manifest ?. May be not, since the change verified to be working and we also don't mount to read CPU info.

README.md
```diff
@@ -147,6 +147,76 @@ However, this is only a partial replacement of the corresponding CPU Manager opt
...
+
+The DRA CPU Driver can expose the PCIe root locality of CPU devices via the standard `resource.kubernetes.io/pcieRoot` attribute.
+This feature is opt-in, and requires _both_ the `DRAListTypeAttributes` Feature Gate (see KEP-5491) enabled in the cluster and the `--expose-pcie-roots` command line
```
pravk03, line 154: nit: For completeness, lets also add this flag to the `## Driver configuration` section

README.md
```diff
@@ -147,6 +147,76 @@ However, this is only a partial replacement of the corresponding CPU Manager opt
...
+flag in the driver. The driver has no way to introspect the cluster feature gate states, so care must be taken to keep the configuration consistent.
+
+**IMPORTANT NOTE**: it is recommended to consume the pcieRoots list attributes using the `matchAttribute` or [the derived attributes](https://github.com/kubernetes/enhancements/issues/6080).
```
pravk03: nit: `pcieRoot` for consistency with attribute name ?. (I wish the standard attribute was plural though)

pravk03 (commented):
Thanks @ffromani! This looks great, and it's even better that we are aligning with the standardattribute :). Thanks @fmuyassarov as well for verifying the changes on real hardware.

Added some minor nits, but otherwise LGTM. 

/lgtm
/hold 
(holding for others to review the latest updates)

pkg/device/pciescan_amd64.go
ffromani: `/sys` and `/proc` are usually mounted on containers - many programs depend on the availability of pseudo/virtual FSes anyway, but I realized I didn't check the container promises and guarantees lately.
Do we want to block on this?

pkg/driver/dra_hooks.go
```diff
@@ -82,10 +79,11 @@ func (cp *CPUDriver) createGroupedCPUDeviceSlices(logger logr.Logger) [][]resour
...
 
 			deviceAttrs := map[resourceapi.QualifiedName]resourceapi.DeviceAttribute{
-				"dra.cpu/socketID":   {IntValue: ptr.To(int64(socketID))},
-				"dra.cpu/numCPUs":    {IntValue: ptr.To(availableCPUsInSocket)},
-				"dra.cpu/smtEnabled": {BoolValue: ptr.To(cp.cpuTopology.SMTEnabled)},
+				AttributeSocketID:   {IntValue: ptr.To(int64(socketID))},
+				AttributeNumCPUs:    {IntValue: ptr.To(availableCPUsInSocket)},
+				AttributeSMTEnabled: {BoolValue: ptr.To(cp.cpuTopology.SMTEnabled)},
 			}
+			cp.setPCIeRootsAttribute(deviceAttrs, allocatableCPUs.UnsortedList()...)
```
AutuSnow, line 86: Here we publish `resource.kubernetes.io/pcieRoot` for the grouped device from all allocatable CPUs in the group. Since `GetPCIeRootsForCPU` returns the union of roots across those CPUs, a socket/NUMA grouped device can advertise `[rootA, rootB]`.
﻿
With `DRAListTypeAttributes`, `matchAttribute` uses set-intersection semantics. That means the scheduler may match this CPU group with another device on `rootA`, but during `PrepareResources` the grouped allocator still only constrains by socket/NUMA and does not know that `rootA` was the matching root. It can therefore assign CPUs from the same group that are not actually local to `rootA`.
﻿
This matches the tradeoff discussed in #114: NUMA/socket grouping plus a list of PCIe roots can make allocation succeed, but does not strictly guarantee that the allocated CPUs are closest to the matched PCIe root.
﻿
Would it be safer to publish the intersection of per-CPU roots for grouped devices, or only publish the standard `pcieRoot` attribute in individual mode / single-root grouped cases until grouped allocation can honor the matched root?

AutuSnow: My concern is mostly that once this is exposed through the standard `resource.kubernetes.io/pcieRoot` attribute, users may naturally use `matchAttribute` for cross-driver locality and expect the matched root to constrain the final CPU assignment. If we intentionally keep union semantics for grouped devices, I think the README should call out that `pcieRoot` in grouped mode means “the group may contain CPUs associated with these roots”, not “the allocated CPUs are guaranteed to match the selected root”.

ffromani: Thanks for your comments. I understand the concern here. The proper fix as I see it requires too stages
1. consume the pcieRoot locality data in the cpu allocation algos - which requires us to go and effectively fork the `cpumanager` code we import and pseudo-vendor. This is IMO not a problem per se, but I'm reluctant to add it now in the 0.2.0 cycle. I'm more than happy and willing to do this in 0.3.0 however
2. actually having fine grained pcie Root data to feed the allocator into - this is tied to the previous overlay idea which I proposed and then ditched in #156 . Likewise, I'd be happy to work on this in the 0.3.0 cycle, this PR offers a much better base to reintroduce the overlay concept.

The problem therefore **does** exist in theory, but the research (docs/dev/topology-linux-sysfs.md) I've done makes me reasonnably confident this is not a problem in practice because the pcie root locality data we are currently consuming **collapses into the NUMA locality data** because how the firmware and then the Linux kernel reports it.

The pcie root locality data is **meant** to be a finer grained sub-NUMA locality signal, but the current Linux/firmware implementation only reports a coarser grained signal, which is not wrong (according to my research) but always less precise than it should be.

The coarsiness and NUMA aliasing is therefore incidental, not structural - the pcie Root locality is a useful signal to carry around and expose and consume in general, not an elaborate workaround to approximate the NUMA locality signal. It can and should be used in conjunction with the (upcoming) NUMA locality signal, and can only improve in precision (the overlay idea can help here at least short term).

Therefore, before the granularity mismatch and the fact the signal devolves in "NUMA affinity" (or "Socket affinity") I do think the weaker guarantees are semantically OK.

I'm willing to document this better (independently from how the rest of this conversation unfolds) but I'm not sure how to do it better because this is a very nuanced topic.

README.md
ffromani: thanks, done

README.md
ffromani: fixed, thanks

ffromani:
/hold cancel

addressed the nits (LGTM gone) and open comments, so the hold is no longer necessary

pkg/driver/dra_hooks.go
AutuSnow: Thanks, that matches my understanding.
﻿
I agree this does not need to block the 0.2.0 cycle on teaching the grouped allocator to consume PCIe root locality. That is a larger change and probably belongs in a follow-up.
﻿
One thing I want to separate is the allocator work from the publication semantics. Using the intersection of per-CPU roots for grouped devices would not make the allocator smarter, but it would avoid advertising a PCIe root unless every possible CPU allocation from that group can satisfy it.
﻿
If the current sysfs data effectively collapses to NUMA locality, then for NUMA-grouped devices the union and intersection should usually be the same in practice. If the data becomes finer-grained later, for example through the overlay idea, the intersection behavior would keep the standard `resource.kubernetes.io/pcieRoot` semantics conservative. For socket grouping, an empty intersection would also be a useful signal that the socket group cannot honestly provide a single guaranteed PCIe-root locality contract.
﻿
That said, I am okay treating this as non-blocking if we explicitly document the current grouped-mode contract. I think the important bit is to say that in grouped mode `pcieRoot` currently means “the group may contain CPUs associated with these roots”, not “the final allocated CPU set is guaranteed to match the root selected by `matchAttribute`”.
﻿
Could we also track the stronger follow-up for 0.3.0, either to make grouped allocation consume PCIe root locality or to revisit grouped-mode publication semantics once finer-grained data is available?

ffromani: The limitations were already documented in the README.md, I moved them in a specific section, more explicit, with versioning information and timeline sketch for the future improvements. I will file tracking issues when we merge this PR.

AutuSnow:
LGTM For me

pkg/driver/dra_hooks.go
ffromani: > Could we also track the stronger follow-up for 0.3.0, either to make grouped allocation consume PCIe root locality or to revisit grouped-mode publication semantics once finer-grained data is available?

Absolutely. And I'm very interested in working on the followup for 0.3.0. I'm totally willing to file tracking issues (for obvious reasons only after this PR merges :) ) and working in this area.

ffromani:
@fmuyassarov howdy! any further concern, comment about the PR? want to make sure your perspective is taken into account.

fmuyassarov:
> @fmuyassarov howdy! any further concern, comment about the PR? want to make sure your perspective is taken into account.

Thanks @ffromani . 
I'm doing one last review right now. Give me few hours to finish it and do one more test on my environment.

ffromani:
> > @fmuyassarov howdy! any further concern, comment about the PR? want to make sure your perspective is taken into account.
> 
> Thanks @ffromani . I'm doing one last review right now. Give me few hours to finish it and do one more test on my environment.

thanks for your effort. And no problem at all, this is why I asked!

fmuyassarov:
Thanks @ffromani. I think we are in good shape now from the changes and I did a quick final test with individual mode and it went smooth. Thanks for adding this feature!
/lgtm

ffromani:
thanks everyone folks! this was a long journey and the comprehensive, detailed reviews directly contributed to shape a better codebase and a stronger feature. I call this a joint effort and a shared success.

PR #167, opened 2026-06-06
Make Helm chart portable by removing hardcoded private image registry

saros-dev:
## Problem

The Helm chart currently uses a hardcoded image from a private GCP Artifact Registry:

us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu:latest

This causes `ImagePullBackOff` for contributors who do not have access to the registry.

## Impact

- Blocks external contributors from running the chart
- Makes local testing impossible without GCP authentication
- Reduces portability of the project

## Changes in this PR

- Made image configuration fully customizable via `values.yaml`
- Added support for `imagePullSecrets`
- Removed hardcoded dependency on private registry assumptions

## How to test

helm template ./deployment/helm/dra-driver-cpu

## Notes

This change improves chart portability without changing default behavior for existing users.
This change is backward compatible and does not modify default behavior.
Happy to adjust this if maintainers prefer a different approach.

### DIFF
--- a/deployment/helm/dra-driver-cpu/values.yaml
+++ b/deployment/helm/dra-driver-cpu/values.yaml
@@ -20,9 +20,9 @@ fullnameOverride: ""
 # @schema additionalProperties:false
 image:
   # -- Container image repository
-  repository: us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu # @schema required:true
+  repository: dra-driver-cpu # @schema required:true
   # -- Image tag; defaults to `.Chart.AppVersion` when empty, which is set to the release tag at package time
-  tag: ""
+  tag: "latest"
   # -- Image pull policy
   pullPolicy: IfNotPresent # @schema required:true;enum:[Always, IfNotPresent, Never]
 


saros-dev:
Thanks for the review requests.

I have completed the CLA authorization. Please let me know if any additional changes are needed.

Thank you!

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -20,9 +20,9 @@ fullnameOverride: ""
 # @schema additionalProperties:false
 image:
   # -- Container image repository
-  repository: us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu # @schema required:true
+  repository: dra-driver-cpu # @schema required:true
   # -- Image tag; defaults to `.Chart.AppVersion` when empty, which is set to the release tag at package time
-  tag: ""
+  tag: "latest"
```
ffromani, line 25: I think this should be left `""` because this way helm can fall back to `.Chart.AppVersion`. Then yes, nowadays `.Chart.Appversion` is `latest`, but I think we want to keep the indirection

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -20,9 +20,9 @@ fullnameOverride: ""
 # @schema additionalProperties:false
 image:
   # -- Container image repository
-  repository: us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu # @schema required:true
+  repository: dra-driver-cpu # @schema required:true
```
ffromani, line 23: I think `--set image.repository` should already allow to override this value. Is that the case? is there another issue which is eluding us?

ffromani (commented):
Thanks for your PR!
Improving devex/ux is surely something we want to do, and indeed we can have better defaults. Likely, we should default to the previous released version by default and document how to point to a custom-built development version.
That said, unfortunately it is not obvious to me how the PR solves the problem you are describing. Perhaps you could maybe elaborate a bit?

saros-dev:
Thanks for the review.

The issue I encountered was during a clean Helm installation using the chart defaults. The resulting DaemonSet attempted to pull the image from:

`us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu`

and consistently failed with `ImagePullBackOff` due to `403 Forbidden` responses from the registry.

My initial assumption was that the chart defaults should point to an image that is directly consumable by users evaluating the project or running it in local development environments. Based on that assumption, I proposed changing the default repository to avoid the dependency on the staging registry.

That said, I understand your point that `image.repository` is already configurable via values or `--set`, and that changing the default repository may not be the right solution if the current staging image is intentionally used as the canonical default.

Perhaps a better approach would be to document the expected workflow for local development and clarify which image repository users should override to when deploying the chart outside of the Kubernetes release infrastructure.

I am happy to adjust the PR direction if documentation or another approach would be preferred.

fmuyassarov:
1. `us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu:latest` isn't private registry in fact. It is a public but owned/hosted in GCP with CNCF. 
**EDIT**: I guess I was too fast to write. It is private repo, you are right. Because, it is only driver authors who have rights to publish images to that repo. But you are free to pull it without any credentials. 
2. I would not necessarily call the defaults as hardcoded. Those are default values and can be easily tuned to your custom one. And we are using latest as the way to promote the latest version of the project to try out. If there is a preference to use the older version, just pass down the helm flag. 

> Blocks external contributors from running the chart

This is not true, perhaps I'm missing, but if you just for example spin up an empemeral environment which has no credentials, I'm sure you still should be able to install the driver with Helm as long as you have access to the internet. 
> Makes local testing impossible without GCP authentication

Again, this is not correct. This staging repo is not private, you don't need any credentials to pull the images, of course you can't push( but that's another story).

One thing I could suggest is to use the IfNotPresent image pull policy instead of Always, but even this can be changed during the Helm install with Helm flag. But we can make it by default. However, it won't do anything else apart from searching the local registry for the image and then falling back to the upstream one.

fmuyassarov:
> Thanks for the review.
> 
> The issue I encountered was during a clean Helm installation using the chart defaults. The resulting DaemonSet attempted to pull the image from:
> 
> `us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu`
> 
> and consistently failed with `ImagePullBackOff` due to `403 Forbidden` responses from the registry.
> 
this is from en ephemeral GitHub codespace
```
➜ /workspaces/dra-driver-cpu (main) $ docker pull us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu:latest
latest: Pulling from k8s-staging-images/dra-driver-cpu/dra-driver-cpu
811bf8dbf451: Pull complete 
52630fc75a18: Pull complete 
dd64bf2dd177: Pull complete 
dcaa5a89b0cc: Pull complete 
39dc083afc39: Pull complete 
7c12895b777b: Pull complete 
b839dfae01f6: Pull complete 
3214acf345c0: Pull complete 
2780920e5dbf: Pull complete 
7678d39f4ae5: Pull complete 
bf7a4185f015: Pull complete 
c3582e9a7cc7: Pull complete 
990a9c434e5e: Pull complete 
235c3625d753: Pull complete 
069d1e267530: Pull complete 
875ea9878944: Pull complete 
a585a1c35e73: Download complete 
Digest: sha256:49957462c7819503cbf5f85da505ad91d2e619103f4094ee8d558a5525151866
Status: Downloaded newer image for us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu:latest
us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu/dra-driver-cpu:latest
```

saros-dev:
Thanks for taking the time to verify this and for the detailed explanation.

Sorry for the delayed response.

It looks like my issue was environment-specific rather than a chart-level problem. I likely hit a network or regional restriction that caused the misleading 403 behavior on my side.

Given your confirmation that the registry is publicly accessible and the current defaults are intentional, I agree that changing the chart defaults is not necessary.

I'll close this PR.

Thanks again for the review and clarification.

ffromani:
/close

per https://github.com/kubernetes-sigs/dra-driver-cpu/pull/167#issuecomment-4673908659

PR #170, opened 2026-06-08
Fix PrepareResourceClaims allocation commit ordering

### LINKED ISSUE #169: PrepareResourceClaims mutates CPU allocation store before CDI device creation succeeds
## Problem
﻿
In both grouped and individual modes, the driver commits the claim CPU allocation before writing the CDI device:
﻿
- grouped mode: `AddResourceClaimAllocation` happens before `cdiMgr.AddDevice`
- individual mode: same ordering
﻿
If CDI creation fails, `PrepareResourceClaims` returns an error, but the in-memory allocation store has already removed those CPUs from the shared pool.
﻿
## Impact
﻿
A failed Prepare can poison the driver's allocation state. Later retries or unrelated claims may see fewer available CPUs than actually allocated.
﻿
## Expected behavior
﻿
Prepare should be atomic from the driver's perspective: either both allocation state and CDI spec are committed, or neither is.

AutuSnow:
/kind bug
﻿
This fixes `PrepareResourceClaims` so CPU allocation state is only committed after CDI device creation succeeds.
﻿
Previously, both grouped and individual prepare paths updated the in-memory CPU allocation store before calling `cdiMgr.AddDevice`. If CDI device creation failed, prepare returned an error but the CPUs remained removed from the shared pool, leaving the driver state polluted.
﻿
This PR changes prepare to:
﻿
- compute the desired CPUSet first
- create/update the CDI device before committing allocation state
- replace the claim allocation only after CDI succeeds
- keep same-claim repeated Prepare calls idempotent by allowing the claim to reuse its existing CPUs
- reuse an existing grouped allocation when it still satisfies the current claim, instead of silently selecting a different CPUSet
﻿
﻿
Fixes https://github.com/kubernetes-sigs/dra-driver-cpu/issues/169
﻿

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -408,9 +408,12 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		return kubeletplugin.PrepareResult{}
 	}
 
+	result := cp.prepareDevices(logger, claim, cpuAssignment)
+	if result.Err != nil {
+		return result
+	}
 	cp.cpuAllocationStore.AddResourceClaimAllocation(logger, claim.UID, cpuAssignment)
-
-	return cp.prepareDevices(logger, claim, cpuAssignment)
+	return result
 }
 
 func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi.ResourceClaim) kubeletplugin.PrepareResult {
@@ -461,8 +464,12 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 		}
 	}
 
+	result := cp.prepareDevices(logger, claim, claimCPUSet)
+	if result.Err != nil {
+		return result
+	}
 	cp.cpuAllocationStore.AddResourceClaimAllocation(logger, claim.UID, claimCPUSet)
-	return cp.prepareDevices(logger, claim, claimCPUSet)
+	return result
 }
 
 func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.ResourceClaim, claimCPUSet cpuset.CPUSet) kubeletplugin.PrepareResult {
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -821,6 +821,125 @@ func TestPrepareResourceClaims(t *testing.T) {
 	}
 }
 
+func TestPrepareResourceClaimsDoesNotCommitAllocationWhenCDIFails(t *testing.T) {
+	logger := testr.New(t)
+	claimUID := types.UID("claim-cdi-fails")
+	existingCPUs := cpuset.New(0, 1)
+
+	individualDriver := func(withExistingAllocation bool) *CPUDriver {
+		mockProvider := &cpuinfo.MockCPUInfoProvider{CPUInfos: mockCPUInfos_SingleSocket_4CPUS_HT}
+		topo, err := mockProvider.GetCPUTopology(logger)
+		require.NoError(t, err)
+		driver := &CPUDriver{
+			driverName: testDriverName,
+			deviceNameToCPUID: map[string]int{
+				"cpudev0": 0,
+				"cpudev1": 1,
+				"cpudev2": 2,
+				"cpudev3": 3,
+			},
+			cpuAllocationStore: store.NewCPUAllocation(topo, cpuset.New()),
+		}
+		if withExistingAllocation {
+			driver.cpuAllocationStore.AddResourceClaimAllocation(logger, claimUID, existingCPUs)
+		}
+		return driver
+	}
+
+	groupedDriver := func(withExistingAllocation bool) *CPUDriver {
+		mockProvider := &cpuinfo.MockCPUInfoProvider{CPUInfos: mockCPUInfos_SingleSocket_4CPUS_HT}
+		topo, err := mockProvider.GetCPUTopology(logger)
+		require.NoError(t, err)
+		driver := &CPUDriver{
+			driverName:             testDriverName,
+			cpuDeviceMode:          CPU_DEVICE_MODE_GROUPED,
+			cpuDeviceGroupBy:       GROUP_BY_SOCKET,
+			cpuTopology:            topo,
+			deviceNameToSocketID:   map[string]int{"cpudevsocket0": 0},
+			deviceNameToNUMANodeID: map[string]int{},
+			cpuAllocationStore:     store.NewCPUAllocation(topo, cpuset.New()),
+		}
+		if withExistingAllocation {
+			driver.cpuAllocationStore.AddResourceClaimAllocation(logger, claimUID, existingCPUs)
+		}
+		return driver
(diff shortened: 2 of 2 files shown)


pravk03:
Thanks. Will look at the changes soon but I am curious to know what failed during CDI update and if we do anything to improve.

ffromani:
I'll also be looking to this in addition to #171 . This issue is making me thinking again to how the driver manages the few little state it needs. In the recovery path (`NRI.synchronize`) we trust the runtime. Up until we pass the baton to the runtime - basically up till `NRI.createContainer` returns succesfully, we should probably use CDI files as durable store and manage the data flow accordingly. This was a pattern more or less already emerging from the code, perhaps we should fully make explicit and embrace it.
From the opposite angle: I _think_ CDI files are all the state we need; I'd be in general quite reluctant to _introduce_ extra state to manage (e.g. checkpoint files, db...) unless there's a crystal clear need.
All that said, I'm curious to review the PR.

AutuSnow:
> Thanks. Will look at the changes soon but I am curious to know what failed during CDI update and if we do anything to improve.

The failure I am fixing here is any error returned by cdiMgr.AddDevice / cache.WriteSpec.

In my environment I hit this while investigating node/runtime disruption, but I do not yet have a final RCA for the underlying CDI write failure. The driver bug is that even when AddDevice failed and PrepareResourceClaims returned an error, we had already updated the in-memory CPU allocation store. That polluted the shared CPU pool for later retries or unrelated claims.

This PR improves that by making PrepareResourceClaims atomic from the driver's perspective: the CPU allocation is committed only after the CDI device has been written successfully. I also added tests that inject a CDI add error for both grouped and individual modes and verify the allocation store is left unchanged.

AutuSnow:
> I'll also be looking to this in addition to #171 . This issue is making me thinking again to how the driver manages the few little state it needs. In the recovery path (`NRI.synchronize`) we trust the runtime. Up until we pass the baton to the runtime - basically up till `NRI.createContainer` returns succesfully, we should probably use CDI files as durable store and manage the data flow accordingly. This was a pattern more or less already emerging from the code, perhaps we should fully make explicit and embrace it. From the opposite angle: I _think_ CDI files are all the state we need; I'd be in general quite reluctant to _introduce_ extra state to manage (e.g. checkpoint files, db...) unless there's a crystal clear need. All that said, I'm curious to review the PR.

@ffromani Thanks, this matches how I was thinking about the fix.

This PR does not introduce any extra durable state. It keeps CDI files as the handoff point before the runtime consumes the allocation, and only updates the in-memory allocation store after the CDI spec has been written successfully.

So the intended ordering here is:

1. compute the desired CPUSet
2. write/update the CDI device
3. commit the allocation into the driver memory state

If CDI writing fails, PrepareResourceClaims returns the error and leaves the previous allocation state untouched. After the runtime has created the container, I agree that NRI.synchronize should continue to rebuild from what the runtime reports.

I also agree that adding checkpoint files or a DB would need a much clearer reason. This PR is meant to stay within the existing CDI-as-durable-handoff model.

ffromani:
>     * keep same-claim repeated Prepare calls idempotent by allowing the claim to reuse its existing CPUs

I only begun started looking at this PR but it seems to me that you are fixing 2 things at once which deserve separate tracking and separate fixes.

I suspect we already have a problem in main branch that retries are not idempotent - and they should be.
If so, we need a separate issue to track the fact PrepareResourceClaim must be idempotent - repeated calls must do the right thing, and I'm not sure today they are, and this bug should be fixed _first_.

I'm confident that once PrepareResourceClaim is fully idempotent the fix for AddDevice failing can be simpler and clearer - hopefully "just" a reordering of operations.

Both issues seems to be blocking for 0.2.0.

AutuSnow:
@ffromani 
I saw #175, which extracts the PrepareResourceClaims idempotency coverage from this PR and makes the pre-existing retry bug explicit.
﻿
That split makes sense to me. I will treat idempotency as a separate prerequisite and then reduce this PR back to the #169 scope: keep allocation commit ordering atomic around CDI device creation.
﻿
Once the idempotency fix lands, I can rebase this PR on top of it and remove the duplicated idempotency-specific changes from #170, leaving only:
1. compute the desired CPUSet;
2. write/update the CDI device;
3. commit the allocation store only after CDI succeeds;
4. preserve the previous allocation when CDI update fails.

pravk03:
> I suspect we already have a problem in main branch that retries are not idempotent - and they should be.
If so, we need a separate issue to track the fact PrepareResourceClaim must be idempotent - repeated calls must do the right thing, and I'm not sure today they are, and this bug should be fixed first.

This is right. I have a fix for this in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/165/changes/ceeec6086749cfd303cfb9e4a39a54eb596c6e29. I think we can absorb the change in this PR ? 

[Edit] I just saw https://github.com/kubernetes-sigs/dra-driver-cpu/pull/175, the fix can go in a separate PR too, but there seems to be enough overlap between the changes https://github.com/kubernetes-sigs/dra-driver-cpu/pull/165/changes/ceeec6086749cfd303cfb9e4a39a54eb596c6e29 and this PR. So, I am ok either way.

ffromani:
> > I suspect we already have a problem in main branch that retries are not idempotent - and they should be.
> > If so, we need a separate issue to track the fact PrepareResourceClaim must be idempotent - repeated calls must do the right thing, and I'm not sure today they are, and this bug should be fixed first.
> 
> This is right. I have a fix for this in [ceeec60](https://github.com/kubernetes-sigs/dra-driver-cpu/commit/ceeec6086749cfd303cfb9e4a39a54eb596c6e29). I think we can absorb the change in this PR ?
> 
> [Edit] I just saw #175, the fix can go in a separate PR too, but there seems to be enough overlap between the changes [ceeec60](https://github.com/kubernetes-sigs/dra-driver-cpu/commit/ceeec6086749cfd303cfb9e4a39a54eb596c6e29) and this PR. So, I am ok either way.

Great that you have a fix! If you can split the fix + its tests (perhaps we can incorporate testcases from #175? not sure there's any overlap) I think that we can merge it quickly

ffromani:
Folks, I took the liberty to extract and adapt https://github.com/kubernetes-sigs/dra-driver-cpu/pull/165/changes/ceeec6086749cfd303cfb9e4a39a54eb596c6e29 into my #175 - mostly because I'm curious and interested about the issue. We can proceed with #175 or do a cleaner extraction from #165, whatever works best.

Disclosure: I'm still verifying my own #175 to make sure I didn't introduce bugs with my extraction (therefore WIP)

AutuSnow:
@pravk03 @ffromani 
Rebased on latest main after #175 merged.
﻿
#170 is now narrowed to the #169 atomicity fix only: CDI AddDevice runs before committing CPU allocation, and allocation state is preserved on CDI failure. Duplicated idempotency/store changes from #175 were removed, with tests updated for grouped and individual modes.

ffromani (commented):
/approve
/lgtm

the production code changes match my expectation.
The test looks good, albeit borderline too smart. But still OK, and most notably I don't have obvious suggestions so I won't block over unclear improvements.
We will probably need to review and improve tests guideline in the 0.3.0 cycle.

PR #171, opened 2026-06-08
Initialize device lookup maps before publishing

### LINKED ISSUE #168: PrepareResourceClaims depends on PublishResources side effects for device topology lookup maps
## Problem
﻿
`PrepareResourceClaims` resolves allocated device names through in-memory maps such as `deviceNameToSocketID`, `deviceNameToNUMANodeID`, and `deviceNameToCPUID`.
﻿
Those maps are currently populated as a side effect of ResourceSlice generation in `PublishResources`:
﻿
- grouped socket: `pkg/driver/dra_hooks.go`
- grouped NUMA: `pkg/driver/dra_hooks.go`
- individual CPU: `pkg/driver/dra_hooks.go`
﻿
However, `Start` launches `PublishResources` asynchronously. After a driver restart, kubelet may call `PrepareResourceClaims` for an already allocated claim before those maps have been rebuilt.
﻿
## Impact
﻿
Prepare may fail with errors like "no valid socket ID found for device" even though the allocation is valid. The state also depends on publish timing instead of a deterministic device model.
﻿
Kubernetes DRA helper requires `PrepareResourceClaims` to be idempotent and restart-safe.
﻿

AutuSnow:
/kind bug
﻿
﻿
Fixes `PrepareResourceClaims` so it no longer depends on `PublishResources` side effects to resolve allocated device names.
﻿
Previously, the driver populated `deviceNameToCPUID`, `deviceNameToSocketID`, and `deviceNameToNUMANodeID` while generating ResourceSlice devices. Because `PublishResources` runs asynchronously, kubelet could call `PrepareResourceClaims` after a driver restart before those maps were rebuilt, causing valid allocations to fail prepare.
﻿
This PR changes the driver to build those lookup maps during startup, before the kubelet plugin can serve prepare calls.
﻿
The ResourceSlice publication path now uses the same stable device enumeration helpers, but no longer mutates the lookup maps.
﻿
Fixes #168
﻿

### DIFF
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -351,7 +351,7 @@ func TestPublishResources(t *testing.T) {
 				totalDevices += len(s.Devices)
 			}
 			require.Equal(t, tc.expectedDevices, totalDevices)
-			require.Equal(t, tc.expectedDevices, len(cp.deviceNameToCPUID))
+			require.Empty(t, cp.deviceNameToCPUID)
 
 			// Verify device attributes
 			cpuInfosMap := make(map[int]cpuinfo.CPUInfo)
@@ -362,19 +362,12 @@ func TestPublishResources(t *testing.T) {
 			seenCPUIDs := make(map[int]bool)
 			for _, slice := range pool.Slices {
 				for _, device := range slice.Devices {
-					cpuID, ok := cp.deviceNameToCPUID[device.Name]
-					require.True(t, ok)
+					cpuID := int(*device.Attributes[AttributeCPUID].IntValue)
 					require.False(t, seenCPUIDs[cpuID], "duplicate cpuID found in slices: %d", cpuID)
 					seenCPUIDs[cpuID] = true
 
-					var cpuInfo *cpuinfo.CPUInfo
-					for i := range tc.cpuInfos {
-						if tc.cpuInfos[i].CpuID == cpuID {
-							cpuInfo = &tc.cpuInfos[i]
-							break
-						}
-					}
-					require.NotNil(t, cpuInfo, "could not find matching cpuInfo for device %s", device.Name)
+					cpuInfo, ok := cpuInfosMap[cpuID]
+					require.True(t, ok, "could not find matching cpuInfo for device %s", device.Name)
 
 					numaNode := int64(cpuInfo.NUMANodeID)
 					CacheL3ID := int64(cpuInfo.UncoreCacheID)
@@ -395,8 +388,11 @@ func TestPublishResources(t *testing.T) {
 
 			// Test if hyperthreads are given successive device names
 			cpuIDToDeviceName := make(map[int]string)
-			for devName, cpuID := range cp.deviceNameToCPUID {
-				cpuIDToDeviceName[cpuID] = devName
+			for _, slice := range pool.Slices {
+				for _, device := range slice.Devices {
+					cpuID := int(*device.Attributes[AttributeCPUID].IntValue)
+					cpuIDToDeviceName[cpuID] = device.Name
+				}
 			}
 			for _, info := range tc.cpuInfos {
 				if info.SiblingCPUID == -1 || info.CpuID > info.SiblingCPUID {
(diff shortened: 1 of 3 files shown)


pkg/driver/dra_hooks_test.go
```diff
@@ -411,6 +407,122 @@ func TestPublishResources(t *testing.T) {
 	}
 }
 
+func TestInitializeDeviceLookupMaps(t *testing.T) {
```
ffromani, line 423: so, the effect of the bug would have been that a `PrepareResource` call received before the first `PublishResources` async call would have failed unnecessarily and wrongly.
I guess this may happen on driver restart when `ResourceSlice` was not cleaned up. IIRC the system should clean up stale ResourceSlices, but if the restart is fast enough, we may have a race.

I think the test which we want is indeed that a call to `PrepareResource` _succeeds_ before the first `PublishResources` completes. To enable this testing we may want to split driver.Start() in driver.Setup() (prepare all the internal data, gets ready) and a proper driver.Start() (assumes internal data ready, starts the loops/registers/listens).

pkg/driver/driver.go
ffromani: this should probably hoisted up right before `kubeletplugin.Start`

pkg/driver/driver.go
```diff
@@ -135,6 +135,7 @@ func Start(ctx context.Context, clientset kubernetes.Interface, config *Config)
 	plugin.cpuTopology = topo
 	plugin.cpuAllocationStore = store.NewCPUAllocation(plugin.cpuTopology, config.ReservedCPUs)
 	plugin.podConfigStore = store.NewPodConfig()
+	plugin.initializeDeviceLookupMaps()
```
ffromani, line 169: if we go with my proposed split, `Setup` could probably stop here and `Start` should indeed begin with the `os.MkdirAll` call.

ffromani (commented):
I'm trying to narrow down the issue and fully understand the root cause of the issue you reported.

That said, avoiding side effects in PublishResources and moving the initialization to be full and done earlier are good changes anyway.

pkg/driver/driver.go
AutuSnow: I agree with the direction. For this PR I kept the change scoped and added the regression test directly without introducing a new exported setup/start split. The updated startup sequence now prepares all state needed by `PrepareResourceClaims` before `kubeletplugin.Start`.
﻿
If you still prefer the explicit split, I can follow up by extracting the internal setup portion into a helper.

pkg/driver/driver.go
AutuSnow: Agreed. `cdiMgr` is also part of the state required by prepare, so it should be initialized before `kubeletplugin.Start` can expose the plugin to kubelet calls. I moved CDI manager initialization before starting the kubelet plugin.

pkg/driver/dra_hooks_test.go
AutuSnow: Good point. I agree that the regression test should exercise the actual failure mode: `PrepareResourceClaims` succeeding before the first `PublishResources` call.
﻿
I updated the PR to add a prepare-before-publish test which initializes the driver lookup maps during setup, does not call `PublishResources`, and verifies prepare succeeds for individual, socket-grouped, and NUMA-grouped modes.

ffromani:
thanks for the updates. I'm reviewing your PR and experimenting with a slightly different approach. The issue most certainly is a blocker for 0.2.0, let me do some more exploration and we will figure out the best way forward.

ffromani:
/lgtm

This issue fixes the issue at hand. I'd double down on this road and refactor more and deeper, but we can also merge as-is and I can post a followup PR building on this one to cleanup things more.

For example, device data should be done once at init time - we don't support hotplug/hotunplug anyway so redoing work every time it's just wasteful. But I can take care.

Let's wait to see if #163 merges this week. If not, we can go ahead with this and I can post a followup shortly after

AutuSnow:
@ffromani Thanks for the review and LGTM. That follow-up direction makes sense to me. I’ll keep this PR scoped to fixing the prepare-before-publish issue, and we can build the init-time device data cleanup on top of it after #163 settles.

ffromani:
Proposal: merge this fix in 0.2.0, postpone my followup to 0.3.0. I don't want to delay 0.2.0 to add more content. Would that work?

ffromani:
> @ffromani Thanks for the review and LGTM. That follow-up direction makes sense to me. I’ll keep this PR scoped to fixing the prepare-before-publish issue, and we can build the init-time device data cleanup on top of it after #163 settles.

perfect. Per https://github.com/kubernetes-sigs/dra-driver-cpu/pull/171#issuecomment-4680827519 I think we better _postpone_ my own followup to 0.3.0 - let's wrap up 0.2.0 and the fix is good enough. Please rebase and we can merge.

PR #175, opened 2026-06-13
dra: ensure idempotency of PrepareResourceClaims

ffromani:
Ensure DRA calls are idempotent. From #169 and the review of #170 (and independently fixed in #165) it emerged the driver is not correctly implementing this constraint.

Spawned from #170
Includes the fix https://github.com/kubernetes-sigs/dra-driver-cpu/pull/165/changes/ceeec6086749cfd303cfb9e4a39a54eb596c6e29

### DIFF
--- a/pkg/driver/cdi_test.go
+++ b/pkg/driver/cdi_test.go
@@ -180,3 +180,35 @@ func TestRemoveDevice(t *testing.T) {
 		})
 	}
 }
+
+func TestAddDeviceOverwrite(t *testing.T) {
+	logger := testr.New(t)
+	tempCDIDir := t.TempDir()
+
+	mgr, err := NewCdiManager(logger, testDriverName, tempCDIDir)
+	require.NoError(t, err)
+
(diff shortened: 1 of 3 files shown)


ffromani:
test failures are expected and meant to highlight and exercise we have indeed a bug in the DRA flows

pravk03:
I think this should work.

/lgtm
/hold

ffromani:
/hold cancel

ok, self-reviewed and I'm happy to propose this PR. Thanks @pravk03 for your review. I only added a minor cleanup to make the tests more consistent. Removed WIP, fully ready now.

@AutuSnow PTAL. If we're happy with this PR we can move forward and then rebase #170 on top. I expect it to become something quite similar to #180.

AutuSnow:
/lgtm
I’ll keep #175 as the idempotency prerequisite and rebase #170 on top of it.
Then #170 will be narrowed to the atomic allocation-commit fix around CDI device creation, with the overlapping idempotency changes removed.

PR #177, opened 2026-06-14
device: move IsPCIeRootName to arch-neutral source so gen-pcie-testdata builds on all arches

### LINKED ISSUE #176: tools/gen-pcie-testdata fails to build on non-amd64 (undefined: device.IsPCIeRootName)
Building the module on a non-amd64 host fails:

```
$ GOARCH=arm64 go build ./...
tools/gen-pcie-testdata/main.go:131:14: undefined: device.IsPCIeRootName
```

`make lint` reports the same error, since golangci-lint typechecks `./...`.

Cause: `IsPCIeRootName` is defined in `pkg/device/pciescan_amd64.go`, which the `_amd64.go` suffix restricts to amd64. The `//go:build !amd64` sibling `pkg/device/pciescan_other.go` does not define it, and `tools/gen-pcie-testdata` (`main.go:131`) calls it unconditionally, so the tool does not compile on other architectures. The amd64 build is unaffected, so the amd64 build/lint workflow does not catch it.

`IsPCIeRootName` is pure string parsing (it matches the kernel `pci%04x:%02x` root naming) with no architecture-specific behaviour, so one fix is to move it from `pciescan_amd64.go` to the shared `pkg/device/pciescan.go`, and its test from `pciescan_amd64_test.go` to a shared test file. With that change `go build ./...`, `go vet ./...` and `go test ./pkg/device/...` pass on arm64, and amd64 behaviour is unchanged.

I can send a PR for that if it looks right.

rocker-zhang:
tools/gen-pcie-testdata fails to build on any non-amd64 architecture with "undefined:
device.IsPCIeRootName". The tool calls device.IsPCIeRootName unconditionally, but the
function is defined in pkg/device/pciescan_amd64.go, which is compiled only on amd64. On
other architectures the symbol does not exist and the build of the tool breaks.

IsPCIeRootName only matches the shape of a sysfs PCIe-root directory name against the
kernel's "pci%04x:%02x" format, with no architecture-specific behavior. This patch moves
it from pciescan_amd64.go to the shared pciescan.go, and relocates TestIsPCIeRootName from
pciescan_amd64_test.go to the shared pciescan_test.go so the test compiles and runs on
every architecture. The function body is unchanged, so amd64 compiles the same code it did
before.

Splitting the test is deliberate. TestIsPCIeRootName covers a pure string-parsing function
and now runs on every architecture, while TestPCIeDomainsFromFS stays in the amd64 test
file because it depends on the amd64-only PCIeDomainsFromFS and on amd64-tagged testdata
fixtures.

This follows the direction ffromani suggested on the issue: moving IsPCIeRootName into the
arch-independent source.

Verified with go build ./... and go vet ./... on both arm64 and amd64. go test
./pkg/device/... passes on both; TestIsPCIeRootName runs on either architecture, and the
amd64-only TestPCIeDomainsFromFS still passes when run on amd64. The function body is
unchanged, so amd64 builds the same code as before.

Fixes #176


ffromani:
/lgtm

but I dunno why the `invalid-commit-message` label

ffromani:
https://www.kubernetes.dev/docs/guide/pull-requests/#do-not-use-github-keywords-or-mentions-within-your-commit-message

AutuSnow:
The `invalid-commit-message` label is caused by the commit body containing:
`Fixes #176`
﻿
Kubernetes PR checks do not allow GitHub auto-closing keywords or `#` issue mentions in commit messages, because they can automatically close issues when the commit is merged.
﻿
The code change itself looks fine. To clear the label, please amend the commit message and remove that line, or move the issue reference to the PR description/comment instead.

ffromani:
/approve

per previous comment a trivial commit message edit will make this PR mergeable.

rocker-zhang:
Thanks @ffromani @AutuSnow. Amended the commit to drop the `Fixes #176` line and force-pushed, so the invalid-commit-message check is cleared. The issue link stays in the PR description, so it still closes on merge. Also rebased onto current main while at it; the diff is unchanged (the pure move of `IsPCIeRootName`).

ffromani:
/lgtm

thanks!

PR #180, opened 2026-06-15
Fix UnprepareResourceClaims CDI removal ordering

### LINKED ISSUE #178: UnprepareResourceClaims releases CPU allocation before CDI device removal succeeds
unprepareResourceClaim currently removes the claim allocation from the in-memory CPU allocation store before removing the corresponding CDI device

AutuSnow:
This PR updates `UnprepareResourceClaims` to remove the CDI device before releasing the CPU allocation from the driver store.
Previously, `unprepareResourceClaim` removed the claim allocation first and then removed the CDI device. If CDI removal failed due to a real filesystem/permission/IO error, the CPUs were returned to the shared pool while the stale CDI spec could still remain on disk.
﻿
Fixes #178


pravk03:
Thanks for the fix. 

/lgtm
/approve

Issue #132, opened 2026-04-28
Review and improve e2e tests


ffromani:
From the kubernetes experience, e2e tests need constant care. It's due time we run a sweeping cleanup to make sure they are reliable, tidy and actually a good example. This includes and is not limited to
1. general cleanup
2. document best practices
3. add linters/checkers borrowing from k/k where feasible
4. add and use custom matchers to encapsulate common patterns and checks

xref: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/115

PR #182, opened 2026-06-17
fix ResourceSlices device attributes check in machine groupBy mode

pravk03:
Fixes an E2E test failure in `test/e2e/resource_slice_test.go` where device attribute validation failed for the `grouped` device mode with `machine` grouping configuration.


pravk03:
This change got left out in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/165

ffromani (commented):
/lgtm

thanks for the followup.

The code in resource_slice_test.go is also borderline too branchy/smart, but this is largely (like 95%) my fault. Another reason to keep #132 on the radar for 0.3.0 and, perhaps a recurring task.

PR #183, opened 2026-06-17
refactor and split driver.Start()

### LINKED ISSUE #174: Split `driver.New()` from `driver.Start()`, move static inizialization in `driver.New()`
Followup of https://github.com/kubernetes-sigs/dra-driver-cpu/pull/171

the idea is to introduce `driver.New()` which will return a CPUDriver object fully initialized and ready to be registered, and `<instance>.Start()` which will register the instance CPUDriver in the runtime and start serving. All the blocking (context-requiring) operations must be moved in` Start()`.

This refactoring will prevent all the class of bugs #171 fixed and will lead to a cleaner and more testable codebase.

ffromani:
The overarching goal is to avoid entirely the class of issues like https://github.com/kubernetes-sigs/dra-driver-cpu/issues/168

The core idea: move all the data initialization in the newly-introduced `driver.New()`, moving there all the initialization previously scatthered in the flows (e.g. publish resources).
With this series applied, the idea is when New() ends, the driver data initialization is complete, and the
object is ready to be wired using Start(); after that, the driver is finally ready to serve.

This split unlocks some further cleanups and code moving, e.g. computing device data once vs every time needed - device data is functionally tied to the driver lifecycle anyway, we don't support hot(un)plug so actually recomputing opens the door to subtle bugs.

**Notes for the reviewers**

- due to the simple and code-movement-driven nature of this PR, LLM assistance was used. All the changes are human-driven and thoroughfully reviewed by humans prior to submission.
- I have followup PR in the pipeline to split the device enumeration responsability from the CPUDriver object. I'd like to push back against the gravitational pull of `CPUDriver` and prevent it to become a god object - the trend is already slowly starting.

Followup of #171 and #175
Fixes: #174

### DIFF
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -162,7 +162,7 @@ func run(logger logr.Logger) error {
 	}()
 	signal.Notify(signalCh, os.Interrupt, unix.SIGINT)
 
-	driverConfig := &driver.Config{
+	driverConfig := driver.Config{
 		DriverName:       driverName,
 		NodeName:         nodeName,
 		ReservedCPUs:     reservedCPUSet,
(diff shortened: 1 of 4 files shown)


ffromani:
/hold cancel

ready for review

pravk03:
Thank you!

/lgtm

Issue #57, opened 2026-02-06
PLANNING: Release 0.2.0


ffromani:
## Introduction
We keep the simple model begun in #34 till we figure out a better approach. It's lightweight enough which is serving us well in our initial stages.

## Highlights
0.2.0 would be the second alpha version.
This version will default to individual mode
This version would *deprecate* the grouped mode. This operational mode would be implemented into an upcoming `cpumem` integrated driver which will only support the grouped mode

### Note
We prefer issues as tracking mechanism. PRs will fix/implement issues. We will list explictly the rare PRs which don't have a corresponding issue. Please file issues for tracking and discussion as default contribution vehicle.

PRs and issues already merged at time of updating are not listed.

## Issues to fix (not in priority order)

* #118 
* #109
* #107
* #94
* #89
* #72
* #58
* #56 
* #48
* #44  

## PRs to include (not in priority order)

* #126 
* #117 
* #83
* #76
* #68 (prereq: rebase on top of 1.36)

## Nice to have - we can postpone those
* #133 
* #63
* #60

ffromani:
/cc @pravk03 @catblade 

As we start to converge for 0.1.0, let's create a placeholder for the next release. The focus is still wrapping up 0.1.0.

fmuyassarov:
It would be great to include the following patches before the next release:

- [ ] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/76
- [ ] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/68
- [x] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/83
- [x] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/84
- [x] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/95
- [x] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/75
- [x] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/106

holysheep123:
Hi **ffromani** 鈥?happy to help debug. To speed things up, could you share:

1. The exact error message (full text, not just the code)
2. What model/endpoint you're trying to use
3. Whether this was working before or is a fresh setup

鈥?holysheep.ai

pravk03:
We may also want to include arm support ?

- [x] - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/106

fmuyassarov:
[...]

PR #186, opened 2026-06-18
feat: expose cpuset via env vars

### LINKED ISSUE #181: Support assigned.cpuset exposure via CDI (Parity with KEP-6122)
https://github.com/kubernetes/enhancements/pull/6123 is introducing a new field `assigned.cpuset` to the downward API. This allows containers to discover their exclusive CPU assignments  by reading a file projected into the pod. To achieve parity, the DRA driver should expose this information using CDI. This is likely possible through `containerEdits.mounts` but opening this issue to explore more. We can target this change for a release version with 1.37 or later.

xref from discussion in https://github.com/kubernetes/enhancements/pull/6123#discussion_r3416703416

fmuyassarov:
Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/181

Regarding exposing the assigned set via a CDI file mount, I checked the possible use cases vs the trade offs from a security perspective and based on the discussion on the issue, at least for now,  I thought I won't to include it. But, if there is an interest in replacing the env var with a file based approach or even having both, then I'm happy to improve & cherry pick https://github.com/kubernetes-sigs/dra-driver-cpu/commit/85640eaf7c5863a285a4cc6c0b8ecbec98598293. 

Example run: 

```YAML
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
spec:
  resourceClaims:
  - name: pod-cpu-claim-ref
    resourceClaimName: my-cpus
  containers:
  - name: app
    image: busybox
    command: ["sh", "-c", "env | grep DRA_CPUSET; sleep 3600"]
    resources:
      claims:
      - name: pod-cpu-claim-ref
```

```
$ kubectl logs my-pod
DRA_CPUSET_fdab0575-4c48-4566-a4aa-31d8d6552956=0,64
DRA_CPUSET_ASSIGNED=0,64
```

### DIFF
--- a/pkg/driver/cdi.go
+++ b/pkg/driver/cdi.go
@@ -28,7 +28,9 @@ const (
 	cdiVendor       = "dra.k8s.io"
 	cdiClass        = "cpu"
 	cdiEnvVarPrefix = "DRA_CPUSET"
-	cdiSpecDir      = "/var/run/cdi"
+	// This mirrors the semantics of the assigned.cpuset introduced in k/k KEP #6122.
+	cdiEnvVarAssigned = "DRA_CPUSET_ASSIGNED"
+	cdiSpecDir        = "/var/run/cdi"
 )
 
 // CdiManager handles the lifecycle of CDI allocations for the driver.
@@ -66,7 +68,7 @@ func (c *CdiManager) getSpecName(deviceName string) string {
 }
 
 // AddDevice writes a dedicated CDI spec file for a single device allocation.
-func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar string) error {
+func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVars []string) error {
 	specName := c.getSpecName(deviceName)
 
 	spec := &cdiSpec.Spec{
@@ -76,7 +78,7 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
 			{
 				Name: deviceName,
 				ContainerEdits: cdiSpec.ContainerEdits{
-					Env: []string{envVar},
+					Env: envVars,
 				},
 			},
 		},
@@ -86,7 +88,7 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
 		return fmt.Errorf("failed to write CDI spec %q: %w", specName, err)
 	}
 
-	logger.V(4).Info("Added CDI device", "deviceName", deviceName, "specName", specName, "env", envVar)
+	logger.V(4).Info("Added CDI device", "deviceName", deviceName, "specName", specName, "envVars", envVars)
 	return nil
 }
 
--- a/pkg/driver/cdi_test.go
+++ b/pkg/driver/cdi_test.go
@@ -81,7 +81,7 @@ func TestAddDevice(t *testing.T) {
 			expectedSpecName := mgr.getSpecName(tc.deviceName)
 			expectedFilePath := filepath.Join(tempCDIDir, expectedSpecName)
 
-			err = mgr.AddDevice(logger, tc.deviceName, tc.envVar)
+			err = mgr.AddDevice(logger, tc.deviceName, []string{tc.envVar})
 
 			if tc.expectedError != "" {
 				require.Error(t, err)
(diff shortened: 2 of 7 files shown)


fmuyassarov:
Let's target this for the v0.3.0 release as discussed shortly at https://github.com/kubernetes-sigs/dra-driver-cpu/issues/57#issuecomment-4765818262

ffromani:
the change per se LGTM.

the theme which is emerging here is however where we should draw the seam between the env var which we use internally to communicate across the DRA/NRI boundary and which ones we use as public API.
This topic will require more thought in general.

fmuyassarov:
> the change per se LGTM.
> 
> the theme which is emerging here is however where we should draw the seam between the env var which we use internally to communicate across the DRA/NRI boundary and which ones we use as public API. This topic will require more thought in general.

Point taken. I think I didn't fully capture the problem(or other pieces). I will revisit it and perhaps discuss it first over the issue.

ffromani:
> > the change per se LGTM.
> > the theme which is emerging here is however where we should draw the seam between the env var which we use internally to communicate across the DRA/NRI boundary and which ones we use as public API. This topic will require more thought in general.
> 
> Point taken. I think I didn't fully capture the problem(or other pieces). I will revisit it and perhaps discuss it first over the issue.

I'm leaning towards merging this PR anyway and do a cleanup sweep after the fact though. There other PRs touching this area like #192 and #194 and each of them (including obv. this one) makes sense in isolation.

The problem I highlighted needs to be addressed by the time we release 0.3.0, I don't think it should be a blocker for this PR.

fmuyassarov:
Then even better :) 
And we just came out of v0.2.0, meaning we will have enough time to come to best conclusion before next release.

AutuSnow:
I think this needs one more check before merging. `DRA_CPUSET_ASSIGNED` is emitted from each per-claim CDI device, but a container can consume multiple CPU claims. In that case the runtime will inject multiple `DRA_CPUSET_ASSIGNED=<claimCPUSet>` entries with different partial values, and the value observed by the workload becomes runtime/order dependent.
﻿
If the intended semantics mirrors assigned.cpuset, this should be the aggregate cpuset for the container, not the per-claim set. Otherwise we should avoid calling it aggregate/assigned semantics, or document that it is per CDI device. A regression test with a container using two CPU claims would make this visible.

fmuyassarov:
> I think this needs one more check before merging. `DRA_CPUSET_ASSIGNED` is emitted from each per-claim CDI device, but a container can consume multiple CPU claims. In that case the runtime will inject multiple `DRA_CPUSET_ASSIGNED=<claimCPUSet>` entries with different partial values, and the value observed by the workload becomes runtime/order dependent. ﻿ If the intended semantics mirrors assigned.cpuset, this should be the aggregate cpuset for the container, not the per-claim set. Otherwise we should avoid calling it aggregate/assigned semantics, or document that it is per CDI device. A regression test with a container using two CPU claims would make this visible.

Thanks, good point. I think we can mention it in the docs for example that `DRA_CPUSET_ASSIGNED` is scoped per claim to fix/lower the the expectation of the users or then the better approach would be that we compute the union of CPUs for a container using more than a single claim. Though,  this will not be an easy fix because the container to claim mapping is not available to the driver at CDI write time. That info lives in the pod spec, which isn't passed to `PrepareResourceClaims`. I guess it is doable but requires some mechanism (NRI hook...) to inject the correct union.

ffromani:
@AutuSnow good point about the union and multiple claims. I need to spend more time (swamped with other stuff atm) thinking about it, but atm I tend to agree that presenting 1 variable with the union is indeed a requirement, as it mimics the kubelet behavior. Let's investigate.

AutuSnow:
Yeah, I think the union requirement is the important bit here.

The one-claim case makes the current approach look fine, but with multiple CPU claims the CDI device boundary starts leaking into the user-facing API: each claim can inject its own `DRA_CPUSET_ASSIGNED=<claimCPUSet>`, so the container may see multiple values for the same env var. At that point the observed value depends on runtime/env ordering, which feels too shaky for something meant to mirror `assigned.cpuset`.

I’d also separate this from the internal `DRA_CPUSET_<claimUID>` namespace. That env is useful as DRA/NRI plumbing and for recovery, but if we put the public aggregate value under the same `DRA_CPUSET_` prefix we end up needing parser exceptions and it conflicts with #192’s reserved-prefix cleanup.

My current preference would be:

- keep `DRA_CPUSET_<claimUID>` as the internal per-claim metadata;
- use a different name for the public aggregate value, maybe `DRACPU_ASSIGNED_CPUSET`;
- add that aggregate in the NRI create-container path, where all per-claim envs for the container are visible and can be unioned;
- cover it with a multi-claim container test.

I checked the NRI API and `ContainerAdjustment.AddEnv` exists, so this path seems workable without trying to infer container-to-claim mapping inside `PrepareResourceClaims`.
/cc @pravk03

fmuyassarov:
I will explore the options for computing the aggregated CPUs of container and come back to this soon.
putting it on hold for now
/hold

PR #190, opened 2026-06-23
Fail closed on malformed DRA_CPUSET env

### LINKED ISSUE #185: Make DRA_CPUSET env parsing fail closed in NRI
`CreateContainer` currently logs `DRA_CPUSET_*` parsing errors but continues as if the container had no DRA CPU allocation, which makes it fall back to the shared CPU pool.
﻿
`Synchronize` has a similar recovery concern: if an existing container has a malformed `DRA_CPUSET_*` env, the container is skipped while rebuilding allocation state.
﻿
This should not happen in the normal path because the env is generated by this driver through CDI, but fail-closed behavior would be safer for corrupted CDI specs, runtime anomalies, or future env contract changes.
﻿
Related:
- #181
- #110
- kubernetes/enhancements#6123

AutuSnow:
Fixes #185
- Make the NRI `CreateContainer` hook return an error when parsing a driver-owned `DRA_CPUSET_*` env fails.
- Add unit coverage to ensure malformed `DRA_CPUSET_*` no longer falls back to the shared CPU pool.
- Add e2e coverage for the kubelet-facing `CreateContainerError` behavior.
﻿
## Why
`DRA_CPUSET_*` is generated by the driver through CDI. If this env is malformed, treating the container as shared can silently break the workload's CPU allocation guarantees. Failing container creation is easier to debug and safer than running with broken promises.

### DIFF
--- a/pkg/driver/nri_hooks_test.go
+++ b/pkg/driver/nri_hooks_test.go
@@ -133,6 +133,7 @@ func TestCreateContainer(t *testing.T) {
 		container                   *api.Container
 		expectedContainerAdjustment *api.ContainerAdjustment
 		expectedContainerUpdates    []*api.ContainerUpdate
+		expectedErrorContains       string
 	}{
 		{
 			name:               "guaranteed container triggers container adjustment with cpus in resource claim",
@@ -186,7 +187,7 @@ func TestCreateContainer(t *testing.T) {
 			},
 		},
 		{
-			name:               "guaranteed container with malformed env falls back to shared",
+			name:               "guaranteed container with malformed env fails closed",
 			podConfigStore:     store.NewPodConfig(),
 			cpuAllocationStore: store.NewCPUAllocation(topo, cpuset.New()),
 			claimTracker:       store.NewClaimTracker(),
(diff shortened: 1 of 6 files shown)


pkg/driver/nri_hooks_test.go
```diff
@@ -211,8 +209,15 @@ func TestCreateContainer(t *testing.T) {
 				claimTracker:       tc.claimTracker,
 			}
 			adjust, updates, err := driver.CreateContainer(context.Background(), pod, tc.container)
-			require.NoError(t, err)
+			if tc.expectedErrorContains != "" {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tc.expectedErrorContains)
```
ffromani, line 214: we need to do a sweeping pass in our tests to find a better way to check errors which is not the string comparison, which is fragile and awkward. Nothing a concern of this PR though.

test/e2e/e2e_suite_test.go
```diff
@@ -121,6 +121,18 @@ func BeFailedToCreate(fxt *fixture.Fixture) types.GomegaMatcher {
 	}).WithTemplate("Pod {{.Actual.Namespace}}/{{.Actual.Name}} UID {{.Actual.UID}} was not in failed phase")
 }
 
+func EventuallyFailedToCreate(ctx context.Context, fxt *fixture.Fixture, pod *v1.Pod) {
```
ffromani, line 124: this helper will backfire the moment we will need a different polling or timeout. At that point we will just inline the helper back in the tests. Fine for now.

ffromani (commented):
/approve
/lgtm

thanks for taking care

ffromani:
followups:
1. the issue mention `synchronize` - fixing with a separate PR is arguably better
2. `DRACPU_`  or `DRA_CPUSET_` env var prefixes are functionally reserved. This should be documented and enforced somewhere
3. just realized: or NRI hooks fire on _every_ container, not just the ones prepared through DRA. This needs to be doublechecked and, if confirmed, addressed. Immediate response: keep a in-memory set of containers the NRI hooks should process, ignore everything not in that set. This should be fine because when CreateContainer terminates, the handshake between DRA and NRI is completed, so there's no risk of stale data. What would happen on `Synchronize` though?

PR #192, opened 2026-06-23
Tighten DRA_CPUSET env prefix handling

AutuSnow:
- Treat only the exact `DRA_CPUSET_` env var prefix as driver-owned allocation metadata.
- Reuse the same prefix constant when generating and parsing CDI-injected env vars.
- Reject malformed driver-owned env vars with an empty claim UID and document the reserved prefix.
﻿
This addresses the reserved-prefix follow-up from #190. The driver should not treat similar user env names like `DRA_CPUSETFOO` as driver-owned metadata, while still failing closed for malformed `DRA_CPUSET_` values.
﻿
Follow-up for #185 / #190 review comment:


fmuyassarov:
I think this will conflict with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/186, in a sense that `DRA_CPUSET_ASSIGNED` sits within the same prefix. So, `DRA_CPUSET_ASSIGNED` will be treated as claim with uid `ASSIGNED`. We can of course rename `DRA_CPUSET_ASSIGNED` to something else that falls outside the DRA_CPUSET prefix, so that `TrimPrefix()` logic doesn't catch it.

AutuSnow:
Agreed, this conflict is real. #192 makes `DRA_CPUSET_` the reserved per-claim namespace, so `DRA_CPUSET_ASSIGNED` would be parsed as claim UID `ASSIGNED`.
﻿
That would either create a fake claim allocation with #192 alone, or fail the prepared-claim validation once #194 is applied. I think #192 should keep the stricter prefix handling, and #186 should move the public aggregate env outside the `DRA_CPUSET_` namespace.
﻿
This also fits the #181/#186 discussion: `DRA_CPUSET_<claimUID>` is internal per-claim DRA/NRI metadata; the user-facing assigned cpuset should be a separate API name and, if it mirrors kubelet `assigned.cpuset`, should represent the per-container union across all CPU claims.

fmuyassarov:
> This also fits the #181/#186 discussion: `DRA_CPUSET_<claimUID>` is internal per-claim DRA/NRI metadata; the user-facing assigned cpuset should be a separate API name and, if it mirrors kubelet `assigned.cpuset`, should represent the per-container union across all CPU claims.

Yup, I'm on it as mentioned at https://github.com/kubernetes-sigs/dra-driver-cpu/pull/186#issuecomment-4793010250

PR #193, opened 2026-06-23
Fail closed during NRI synchronize

AutuSnow:
- Make NRI `Synchronize` fail closed when parsing driver-owned `DRA_CPUSET_*` env fails.
- Rebuild `ClaimTracker` in temporary state during synchronization.
- Commit synchronized state only after the full sync succeeds.
﻿
#190 made `CreateContainer` fail closed for malformed `DRA_CPUSET_*` env, but `Synchronize` still logged parse errors and skipped affected containers.
﻿
This can rebuild incomplete in-memory state after runtime/plugin restart. This follow-up keeps synchronization atomic: either runtime state is rebuilt successfully, or the previous state is preserved.
﻿
Follow-up for #185 / #190 review:

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -39,6 +39,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cpuAllocationStore := store.NewCPUAllocation(cp.cpuTopology, cp.reservedCPUs)
 	podConfigStore := store.NewPodConfig()
+	claimTracker := store.NewClaimTracker()
 	var containerUpdates []*api.ContainerUpdate
 
 	for _, pod := range pods {
@@ -53,7 +54,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 			claimAllocations, err := parseDRAEnvToClaimAllocations(cLogger, container.Env)
 			if err != nil {
 				cLogger.Error(err, "error parsing DRA env for container")
-				continue
+				return nil, err
 			}
 			containerUID := types.UID(container.GetId())
 			var state *store.ContainerState
@@ -64,7 +65,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				allGuaranteedCPUs := cpuset.New()
 				for uid, cpus := range claimAllocations {
 					caLogger := cLogger.WithValues("claimUID", uid)
-					err := cp.claimTracker.SetOwner(caLogger, uid, types.UID(pod.Uid), container.Name)
+					err := claimTracker.SetOwner(caLogger, uid, types.UID(pod.Uid), container.Name)
 					if err != nil {
 						return nil, err
 					}
@@ -89,6 +90,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cp.podConfigStore = podConfigStore
 	cp.cpuAllocationStore = cpuAllocationStore
+	cp.claimTracker = claimTracker
 
 	// Reconcile container CPU masks to handle cases where the NRI plugin might have crashed
 	// or restarted and missed updating the cgroup settings.
--- a/pkg/driver/nri_hooks_test.go
+++ b/pkg/driver/nri_hooks_test.go
@@ -305,7 +305,9 @@ func TestNRISynchronize(t *testing.T) {
 		runtimePods     []*api.PodSandbox
 		runtimeCtrs     []*api.Container
 		expectedUpdates []*api.ContainerUpdate
-		expectedError   bool
+		expectedError   string
+		expectStoreLen  int
+		expectClaims    int
 	}{
 		{
 			name: "empty runtime state clears the store",
@@ -322,6 +324,8 @@ func TestNRISynchronize(t *testing.T) {
 			runtimePods:     []*api.PodSandbox{},
 			runtimeCtrs:     []*api.Container{},
 			expectedUpdates: []*api.ContainerUpdate{},
+			expectStoreLen:  0,
+			expectClaims:    0,
 		},
 		{
 			name: "mixed containers across multiple pods",
@@ -351,6 +355,8 @@ func TestNRISynchronize(t *testing.T) {
 					Linux:       &api.LinuxContainerUpdate{Resources: &api.LinuxResources{Cpu: &api.LinuxCPU{Cpus: "2-7"}}},
 				},
 			},
+			expectStoreLen: 2,
+			expectClaims:   1,
 		},
 		{
 			name: "only shared containers",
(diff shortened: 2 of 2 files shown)


pkg/driver/nri_hooks_test.go
```diff
@@ -418,21 +428,47 @@ func TestNRISynchronize(t *testing.T) {
...
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			updates, err := tc.driver.Synchronize(context.Background(), tc.runtimePods, tc.runtimeCtrs)
 
-			if tc.expectedError {
+			if tc.expectedError != "" {
 				require.Error(t, err)
+				require.Contains(t, err.Error(), tc.expectedError)
+				require.Equal(t, tc.expectStoreLen, tc.driver.podConfigStore.Len())
+				require.Equal(t, tc.expectClaims, tc.driver.claimTracker.Len())
```
pravk03: nit: I think we should avoid checking internal states and only test the function behavior.

AutuSnow: Updated. The test no longer checks `podConfigStore` or `claimTracker` directly.

It now verifies the same regression through public behavior: after `Synchronize` fails, subsequent `CreateContainer` calls still observe the previous state instead of partially synchronized state.

pkg/driver/nri_hooks_test.go
```diff
@@ -419,32 +420,70 @@ func TestNRISynchronize(t *testing.T) {
...
+				driver.podConfigStore.SetContainerState(types.UID("existing-pod"), store.NewContainerState("existing-ctr", "existing-id"))
+				return driver
+			}(),
+			runtimePods: []*api.PodSandbox{pod1},
+			runtimeCtrs: []*api.Container{
+				{Id: "p1-guaranteed", PodSandboxId: pod1.Id, Name: "guaranteed-ctr", Env: []string{fmt.Sprintf("%s_claim-A=%s", cdiEnvVarPrefix, "0,1")}},
+				{Id: "p1-malformed", PodSandboxId: pod1.Id, Name: "malformed-ctr", Env: []string{fmt.Sprintf("%s_claim-A=%s", cdiEnvVarPrefix, "a-b")}},
+			},
+			expectedError: "failed to parse cpuset value",
+			postSyncCheck: func(t *testing.T, driver *CPUDriver) {
```
pravk03: This feels like an integration test wrapped as a unit test and I am thinking if we need to go so far. It seems like we are checking the CPUSet does not get committed if the env var cannot be parsed ?. This should already be tested in CreateContainer unit tests ?. Wondering if there is a realistic scenario where CPUSet is not applied due during original CreateContainer but gets applied during Synchronize() - but even if this happens the `expectedError` check should be sufficient ?.

AutuSnow: I see what you mean. The part I was trying to cover here is not cpuset parsing itself; that is already covered elsewhere. This test is for the `Synchronize` restart path and, more specifically, that we do not commit partially rebuilt state if sync fails halfway through.
﻿
That path is different from `CreateContainer`: after an NRI/plugin restart, `Synchronize` reconstructs state from already-running containers. If one container parses fine and a later one has a malformed driver-owned env, `expectedError` alone proves fail-closed but not that we preserved the previous state.
﻿
I used `CreateContainer` for the post-check because of the earlier feedback to avoid checking internal stores directly. That said, I agree this became a bit heavy. I can simplify the public-behavior check, or drop it and keep only `expectedError` if you think the atomicity check is not worth the extra test weight.

pravk03: > Synchronize reconstructs state from already-running containers.

The fact that container is running would mean that parsing passed in CreateContainer ?. So, the checks for `expectedError` feels sufficient.

AutuSnow: Thank you for your guidance. I am no longer troubled and have updated

pravk03:
Thanks!! 

/lgtm

pkg/driver/nri_hooks.go
```diff
@@ -53,7 +54,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 			claimAllocations, err := parseDRAEnvToClaimAllocations(cLogger, container.Env)
 			if err != nil {
 				cLogger.Error(err, "error parsing DRA env for container")
-				continue
+				return nil, err
```
ffromani, line 57: nit: should we move the logging in the caller? not sure myself, so let's save this for a later review of the logging

pkg/driver/nri_hooks.go
```diff
@@ -89,6 +90,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 
 	cp.podConfigStore = podConfigStore
 	cp.cpuAllocationStore = cpuAllocationStore
+	cp.claimTracker = claimTracker
```
ffromani, line 93: yes, atomic replacement of the finished data structure vs updating incrementally is a better pattern.
thought for a future hardening: doublecheck we use the pattern consistently (arguably the fact we used `claimTracker` incrementally was an existing red flag)

pkg/driver/nri_hooks_test.go
```diff
@@ -305,7 +305,7 @@ func TestNRISynchronize(t *testing.T) {
 		runtimePods     []*api.PodSandbox
 		runtimeCtrs     []*api.Container
 		expectedUpdates []*api.ContainerUpdate
-		expectedError   bool
+		expectedError   string
```
ffromani, line 308: a refactoring priority for the near future is for us to find a way to better match errors than string matching. Not a concern of this PR though.

pkg/driver/nri_hooks_test.go
```diff
@@ -419,31 +419,34 @@ func TestNRISynchronize(t *testing.T) {
...
-				for _, container := range tc.runtimeCtrs {
-					if container.PodSandboxId != pod.Id {
-						continue
-					}
-					state := tc.driver.podConfigStore.GetContainerState(types.UID(pod.Uid), container.Name)
-					require.NotNil(t, state)
-				}
-			}
-
-			// Verify that the returned updates match expectations
```
ffromani, line 446: I strongly agree we should prefer testing public APIs and not against internal data. This change is good.
Still, we are loosening the checks and we should think of a way to re-enable stricter checking somehow, without peeking in the internal data structure. Again not a direct concern of this PR but more thoughts for us for the v0.3.0 cycle.

ffromani (commented):
/approve

thanks!

added inline comments about possible future improvements, nothing required (or, I'm afraid, actionable) for this PR.

AutuSnow:
@ffromani Filed follow-up issues for the post-merge review notes:
﻿
https://github.com/kubernetes-sigs/dra-driver-cpu/issues/203: Audit atomic state replacement in NRI synchronization paths
https://github.com/kubernetes-sigs/dra-driver-cpu/issues/204: Improve NRI test assertions and error matching
﻿
I did not file a separate issue for the logging-location nit for now, since it sounded more exploratory and can fit into a broader logging cleanup/review later.

ffromani:
perfect, thanks @AutuSnow

PR #194, opened 2026-06-23
Validate DRA env against prepared claims

AutuSnow:
- Validate `DRA_CPUSET_*` env entries against claims prepared by this driver.
- Reject containers whose DRA env references an unprepared claim or mismatched cpuset.
- Keep ordinary non-DRA containers in the shared CPU pool.
﻿
NRI `CreateContainer` runs for every container, not only containers prepared through DRA. A container-provided `DRA_CPUSET_*` env should not be trusted unless it matches driver-owned prepare state.
﻿
This addresses the #190 follow-up while preserving the intended shared-pool behavior for non-DRA containers.
﻿
Follow-up:

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -128,6 +128,19 @@ func parseDRAEnvToClaimAllocations(logger logr.Logger, envs []string) (map[types
 	return allocations, nil
 }
 
+func (cp *CPUDriver) validatePreparedClaimAllocations(claimAllocations map[types.UID]cpuset.CPUSet) error {
+	for uid, cpus := range claimAllocations {
+		preparedCPUs, ok := cp.cpuAllocationStore.GetResourceClaimAllocation(uid)
+		if !ok {
+			return fmt.Errorf("claim %q is not prepared by this driver", uid)
+		}
+		if !preparedCPUs.Equals(cpus) {
+			return fmt.Errorf("claim %q has cpuset %q, expected prepared cpuset %q", uid, cpus.String(), preparedCPUs.String())
+		}
+	}
+	return nil
+}
+
 func (cp *CPUDriver) getSharedContainerUpdates(logger logr.Logger, excludeID types.UID) []*api.ContainerUpdate {
 	updates := []*api.ContainerUpdate{}
 	sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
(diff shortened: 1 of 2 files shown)


pkg/driver/nri_hooks.go
```diff
@@ -128,6 +128,19 @@ func parseDRAEnvToClaimAllocations(logger logr.Logger, envs []string) (map[types
 	return allocations, nil
 }
 
+func (cp *CPUDriver) validatePreparedClaimAllocations(claimAllocations map[types.UID]cpuset.CPUSet) error {
+	for uid, cpus := range claimAllocations {
+		preparedCPUs, ok := cp.cpuAllocationStore.GetResourceClaimAllocation(uid)
```
pravk03: nit: Maybe do the check in parseDRAEnvToClaimAllocations() instead of looping through all the claims again ?.  Since `CreateContainer` is on the critical startup path of container.

AutuSnow: Folded this into the existing CreateContainer claim loop so we avoid the extra pass on the startup path.

I kept parseDRAEnvToClaimAllocations() as a pure parser because Synchronize() also uses it while rebuilding state from runtime containers. The validation itself is still factored into a small per-claim helper, so the main loop does the trust-but-verify bit without making the parser know about prepared state.

pkg/driver/nri_hooks.go
```diff
@@ -128,6 +128,17 @@ func parseDRAEnvToClaimAllocations(logger logr.Logger, envs []string) (map[types
 	return allocations, nil
 }
 
+func (cp *CPUDriver) validatePreparedClaimAllocation(uid types.UID, cpus cpuset.CPUSet) error {
+	preparedCPUs, ok := cp.cpuAllocationStore.GetResourceClaimAllocation(uid)
+	if !ok {
+		return fmt.Errorf("claim %q is not prepared by this driver", uid)
+	}
+	if !preparedCPUs.Equals(cpus) {
+		return fmt.Errorf("claim %q has cpuset %q, expected prepared cpuset %q", uid, cpus.String(), preparedCPUs.String())
```
pravk03: nit: 

`validation failed for claim %q: cpuset mismatch (expected %q, got %q)`

pravk03 (commented):
Thanks for this important change. 

/lgtm

pravk03:
Thanks for this!! 
/lgtm

ffromani (commented):
/approve

thanks!

PR #198, opened 2026-06-24
image: replace staging with production images

fmuyassarov:
The default registries are  pointing to the staging registry `us-central1-docker.pkg.dev/k8s-staging-images.` We should instead advertise the production registry `registry.k8s.io` so that users pull from the canonical location by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ $(OUT_DIR):  ## creates the output directory (used internally)
 CLUSTER_NAME=dra-driver-cpu
 IMAGE_NAME=dra-driver-cpu
 # docker image registry, default to upstream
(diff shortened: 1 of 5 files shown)


ffromani (commented):
/hold

to make sure @pravk03 has a chance to take a look

ffromani:
/approve
/lgtm

see https://github.com/kubernetes-sigs/dra-driver-cpu/pull/198#pullrequestreview-4568288013 about unholding

pravk03:
Thanks. I actually looked at this along with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/199, but I missed adding lgtm. 

/lgtm
/unhold

PR #199, opened 2026-06-24
helm: publish chart only on tagged releases

### LINKED ISSUE #196: Use git tags for helm chart versioning
Apparently, chart versioning wasn't set correctly as `GIT_VERSION` was computed as a date+short-sha string which means that even when building from a tagged release commit the `appVersion` in the released chart does not reflect the release tag and instead point to the sha.

I will file a fix later on today.

fmuyassarov:
Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/196

Pass down `TAG_NAME` as `IMG_TAG` in cloudbuild.yaml so that tag triggered builds use the correct version like v0.2.0. Also,  protect helm-push target to run only when `TAG_NAME` is set, otherwise we end up publishing charts during branch push builds as well. `TAG_NAME` is well known cloudbuild default substitution. xref: https://docs.cloud.google.com/build/docs/configuring-builds/substitute-variable-values

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,3 +7,4 @@ steps:
     entrypoint: ./scripts/build-and-publish-image.sh
     env:
     - IMG_PREFIX=us-central1-docker.pkg.dev/k8s-staging-images/dra-driver-cpu
+    - IMG_TAG=$TAG_NAME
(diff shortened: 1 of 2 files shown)


pravk03:
Thanks!

/lgtm
/approve

PR #206, opened 2026-06-25
docs: run kind e2e tests  

Tal-or:
Document the simplest way for developers to run e2e tests locally.
Also add a troubleshooting guide for common mishaps.

 Signed-off-by: Talor Itzhak <titzhak@redhat.com>

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -71,6 +71,16 @@ To run against an existing cluster with the driver already deployed:
 make test-e2e
 ```
 
+Note: `make test-e2e` does not build or load the test image; it only runs the Go e2e suite.
+The `make test-e2e-kind` target builds and loads the image
+automatically into a kind cluster, but for an arbitrary cluster the default `IMAGE_TEST` value
+points at the local CI image name and will not be pullable. Either make sure the
+test image is available to the cluster, or override the image explicitly:
+
+```bash
+DRACPU_E2E_TEST_IMAGE=<image> make test-e2e
+```
+
 ### troubleshooting
 
 - **Kubelet fails with `unknown service runtime.v1.RuntimeService`**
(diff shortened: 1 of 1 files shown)


ffromani (commented):
/approve
/lgtm

thanks, this helps. We are close to the point we need to split our README.md file in targeted readmes, but not a concern of this PR.

AutuSnow:
One small doc caveat here: for the existing-cluster case, I think we should also mention the test image.
﻿
`make test-e2e` does not build/load the test image by itself; it only runs the Go e2e suite with `DRACPU_E2E_TEST_IMAGE=$(IMAGE_TEST)`. In the `make test-e2e-kind` path that image is built and loaded by `ci-kind-setup`, but for an arbitrary existing cluster the default `IMAGE_TEST` value points at the local CI image name and probably is not pullable.
﻿
So maybe this should say that the driver must already be deployed *and* the test image must be available to the cluster, or that users should override `DRACPU_E2E_TEST_IMAGE` with a pullable test image.

Tal-or:
> One small doc caveat here: for the existing-cluster case, I think we should also mention the test image. ﻿ `make test-e2e` does not build/load the test image by itself; it only runs the Go e2e suite with `DRACPU_E2E_TEST_IMAGE=$(IMAGE_TEST)`. In the `make test-e2e-kind` path that image is built and loaded by `ci-kind-setup`, but for an arbitrary existing cluster the default `IMAGE_TEST` value points at the local CI image name and probably is not pullable. ﻿ So maybe this should say that the driver must already be deployed _and_ the test image must be available to the cluster, or that users should override `DRACPU_E2E_TEST_IMAGE` with a pullable test image.

Thank you @AutuSnow for the constructive comment. I'll update that ASAP

ffromani:
/approve cancel

@AutuSnow has a good point

Tal-or:
Added the requested note. see last commit.
Feel free to unhold and approve if happy

ffromani:
/approve

thanks!

PR #207, opened 2026-06-25
remove legacy install.yaml from release artifacts and update related documentation

### LINKED ISSUE #191: do not ship `install.yaml` as release artifact
In the github release page, we should not add `install.yaml` anymore as release artifact as we prefer the helm charts. Shipping both is confusing and can cause friction as we move the focus to helm - `install.yaml` will get out of date fast

swatisehgal:
This PR removes `install.yaml` from the release process and legacy generation paths to reduce confusion and keep Helm as the preferred distribution method. Maintaining both Helm and `install.yaml` was confusing and risked drift. This change keeps the project focused on Helm and avoids stale legacy artifacts in release workflows.

Fixes:  #191

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,33 +166,17 @@ delete-kind-cluster: ## delete kind cluster
 dist:
 	@mkdir -p dist
 
-manifests-legacy: dist install-yq ## create the install manifest (legacy, pre-helm)
-	@cd dist && cp -a ../manifests/base/*.part.yaml .
-ifeq ($(OVERRIDE_IMAGE),true)
-	@$(YQ) -i '.spec.template.spec.containers[0].image = "${IMAGE}"' dist/daemonset-dracpu.part.yaml
-	@$(YQ) -i '.spec.template.spec.containers[0].imagePullPolicy = "IfNotPresent"' dist/daemonset-dracpu.part.yaml
-	@$(YQ) -i '.spec.template.metadata.labels["build"] = "${GIT_VERSION}"' dist/daemonset-dracpu.part.yaml
-endif
-	@$(YQ) '.' \
-		dist/clusterrole-dracpu.part.yaml \
-		dist/serviceaccount-dracpu.part.yaml \
-		dist/clusterrolebinding-dracpu.part.yaml \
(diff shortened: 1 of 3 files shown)


ffromani (commented):
/lgtm

thanks!

AutuSnow (commented):
One issue to fix before merging: this PR changes `make manifests` to write `dist/helm-manifest.yaml`, but the docs still point users at `dist/install.yaml`.

In particular, `README.md` still has the deprecated install.yaml flow as:

```sh
make manifests
kubectl apply -f dist/install.yaml
```

After this change, that command no longer produces `dist/install.yaml`, so the documented flow is broken. The release docs also still talk about generating manifests from `make manifests` without clarifying the new output name. Could you either keep the output path compatible, or update/remove the remaining `install.yaml` documentation so it matches the new `dist/helm-manifest.yaml` behavior?

fmuyassarov (commented):
Thanks. Like @AutuSnow already mentioned it, we need a bit of an update the root readme as well to sync up the changes. Otherwise, lgtm

ffromani:
> One issue to fix before merging: this PR changes `make manifests` to write `dist/helm-manifest.yaml`, but the docs still point users at `dist/install.yaml`.
> 
> In particular, `README.md` still has the deprecated install.yaml flow as:
> 
> ```shell
> make manifests
> kubectl apply -f dist/install.yaml
> ```
> 
> After this change, that command no longer produces `dist/install.yaml`, so the documented flow is broken. The release docs also still talk about generating manifests from `make manifests` without clarifying the new output name. Could you either keep the output path compatible, or update/remove the remaining `install.yaml` documentation so it matches the new `dist/helm-manifest.yaml` behavior?

right, good point. README.md could use a review and an update.
/lgtm cancel

swatisehgal:
Thanks all for the feedback, I will address this in the PR shortly!

AutuSnow:
/lgtm

Thanks！！

fmuyassarov (commented):
LGTM. Thanks. 
/lgtm

pravk03:
/lgtm
Thank you!

ffromani:
/approve

thanks!

Issue #35, opened 2025-12-04
Refactor CPU Topology Representation for Asymmetric/Heterogeneous Systems


pravk03:
This came up during PR review - https://github.com/kubernetes-sigs/dra-driver-cpu/pull/16#discussion_r2588301122

The current CPU topology representation in `pkg/cpuinfo` (specifically the `CPUTopology` struct and its related functions like `CPUsPerCore`, `CPUsPerSocket`, `CPUsPerUncore`) assumes a symmetrical and uniform distribution of CPUs. This oversimplification can lead to inaccurate potential issues in the DRA driver's CPU allocation logic  when CPUs are manually offlined or in systems with heterogenous architectures. 

/kind feature

ffromani:
I didn't check lately. I recommend check your usecase against the general DRA documentation, the scope of this project is node-level, IIUC you need a cluster-level solution.

AutuSnow:
> My musings: This is something for the mid-term future and is tied to other decisions
> 
> * which hw representation we want to use?
> * how much, if at all, should we diverge from borrowed cpumanager code?
> 
> that said, the best solution (and IMO sustainable and achievable) is to stop inferring properties and actually checking what hardware reports. So, instead of dividing total cores for number of sockets, actually have a detailed enough representation and actually enumerate the cores per socket.
> 
> sysfs already exposes all the data we care about, we """just""" need to find a good representation (not the ultra-simplistic very ancient cpumanager representation) and query it.
> 
> On the other hand, we don't need to solve this problem once and for all: we """just""" need a richer representation which fits our needs.

@pravk03 @ffromani In my opinion, a clear data structure, O (1) query, and a clear display of topological relationships can be used to add precomputed mappings in CPUTopology, but it will increase memory usage and result in data duplication with CPUDetails
Example:
  CPUsBySocket map[int]cpuset.CPUSet
  CPUsByCore map[CoreKey]cpuset.CPUSet
  CPUsByUncoreCache map[int]cpuset.CPUSet

Ru13en:
Hi @AutuSnow, @pravk03  and @ffromani 
I drafted a small prototype with the O(1) query structure that @AutuSnow suggested and made it public here:
https://github.com/Ru13en/simple-cpu-topology
The idea is to make a common hierarchical model across architectures.
Maybe it can help the DRA CPU topology parser since it also takes care of L3/ComputeDomain mapping.
[...]

Issue #173, opened 2026-06-11
Expose and consume finer-grained pcie root locality data in the cpu allocator


ffromani:
Raised in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/163#discussion_r3385716792

We need to:
1. consume the pcieRoot locality data in the cpu allocation algos - which requires us to go and effectively fork the cpumanager code we import and pseudo-vendor.

2. actually having fine grained pcie Root data to feed the allocator into - this is tied to the previous overlay idea which I proposed and then ditched in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/156 .

Feeding the allocator with finer-grained data is actually pretty-easy. We need to revamp the `overlay` concept originally added in #156. #163 merged and gives us a simple and straightforward way to feed the data, fully opt-in, using the `sysfs` API.

The main challenge is how to consume sub-NUMA locality data. This is also tied to the LLC affinity data. Currently we mostly borrow the cpumanager allocation code, its limitations and design decision. A prerequisite for this work is to fork that codebase.

Once we have a custom allocator which can take into account sub-NUMA locality, we can fix the pcie root locality problem and then iterate over the LLC locality data, providing better solutions.

PR #210, opened 2026-06-25
use exact CPUDetails queries for asymmetric topology

AutuSnow:
This is a WIP first step for #35.
﻿
Issue #35 has been open for a while, and I do not think this PR should try to settle the larger allocator/model direction in one shot. The goal here is smaller: keep the current CPUManager-style allocator flow, but replace a few symmetric-topology assumptions with exact `CPUDetails` queries where the data is already available.
﻿
I am marking this as WIP because the broader question is still open: whether we continue to stay close to the kubelet CPUManager allocator code, or eventually split/diverge more intentionally. This PR is meant to make the smallest concrete part reviewable first, then iterate based on feedback.

### DIFF
--- a/pkg/cpuinfo/cpuinfo_utils.go
+++ b/pkg/cpuinfo/cpuinfo_utils.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cpuinfo
 
 import (
+	"fmt"
 	"sort"
 
 	"k8s.io/utils/cpuset"
@@ -48,6 +49,10 @@ func (k CoreKey) Less(other CoreKey) bool {
 		return k.SocketID < other.SocketID
 	}
 	return k.ClusterID < other.ClusterID
+}
+
+func (k CoreKey) String() string {
+	return fmt.Sprintf("socket=%d,cluster=%d,core=%d", k.SocketID, k.ClusterID, k.CoreID)
 }
 
 func (i CPUInfo) CoreKey() CoreKey {
--- a/pkg/cpuinfo/cpuinfo_utils.go
+++ b/pkg/cpuinfo/cpuinfo_utils.go
@@ -18,6 +18,7 @@ package cpuinfo
 
 import (
 	"fmt"
+	"slices"
 	"sort"
 
 	"k8s.io/utils/cpuset"
@@ -222,12 +223,7 @@ func (d CPUDetails) CoresInSockets(ids ...int) cpuset.CPUSet {
 // socket IDs.
 func (d CPUDetails) CoreKeysInSockets(ids ...int) []CoreKey {
 	return d.coreKeysFor(func(info CPUInfo) bool {
-		for _, id := range ids {
-			if info.SocketID == id {
-				return true
-			}
-		}
-		return false
+		return slices.Contains(ids, info.SocketID)
 	})
 }
 
@@ -263,12 +259,7 @@ func (d CPUDetails) CoresInNUMANodes(ids ...int) cpuset.CPUSet {
 // NUMA node IDs.
 func (d CPUDetails) CoreKeysInNUMANodes(ids ...int) []CoreKey {
 	return d.coreKeysFor(func(info CPUInfo) bool {
-		for _, id := range ids {
-			if info.NUMANodeID == id {
-				return true
-			}
-		}
-		return false
+		return slices.Contains(ids, info.NUMANodeID)
 	})
 }
 
@@ -311,12 +302,7 @@ func (d CPUDetails) coresInUncoreCache(ids ...int) cpuset.CPUSet {
 
 func (d CPUDetails) coreKeysInUncoreCache(ids ...int) []CoreKey {
 	return d.coreKeysFor(func(info CPUInfo) bool {
-		for _, id := range ids {
-			if info.UncoreCacheID == id {
-				return true
-			}
-		}
-		return false
+		return slices.Contains(ids, info.UncoreCacheID)
 	})
 }
 
--- a/pkg/cpuinfo/cpuinfo_utils.go
+++ b/pkg/cpuinfo/cpuinfo_utils.go
@@ -17,6 +17,10 @@ limitations under the License.
 package cpuinfo
 
 import (
+	"fmt"
+	"slices"
+	"sort"
+
 	"k8s.io/utils/cpuset"
 )
 
@@ -28,6 +32,41 @@ import (
 
 // CPUDetails is a map from CPU ID to Core ID, Socket ID, and NUMA ID.
 type CPUDetails map[int]CPUInfo
+
+// CoreKey identifies a physical core within the topology. CoreID alone is not
+// globally unique on all systems, so callers that need core-level precision
+// should use this key instead of the raw CoreID.
+type CoreKey struct {
+	SocketID  int
+	ClusterID int
+	CoreID    int
+}
+
+// Less provides a deterministic tie-breaker for sorted CoreKey slices. The
+// ordering is CoreID-first to stay aligned with the existing CPUManager-style
+// traversal, where callers already group by socket or NUMA before ordering
+// cores.
+func (k CoreKey) Less(other CoreKey) bool {
+	if k.CoreID != other.CoreID {
+		return k.CoreID < other.CoreID
+	}
+	if k.SocketID != other.SocketID {
+		return k.SocketID < other.SocketID
+	}
+	return k.ClusterID < other.ClusterID
+}
+
+func (k CoreKey) String() string {
+	return fmt.Sprintf("socket=%d,cluster=%d,core=%d", k.SocketID, k.ClusterID, k.CoreID)
+}
+
+func (i CPUInfo) CoreKey() CoreKey {
+	return CoreKey{
+		SocketID:  i.SocketID,
+		ClusterID: i.ClusterID,
+		CoreID:    i.CoreID,
+	}
+}
 
 func (d CPUDetails) KeepOnly(cpus cpuset.CPUSet) CPUDetails {
 	result := CPUDetails{}
@@ -66,13 +105,27 @@ func (d CPUDetails) CPUsInNUMANodes(ids ...int) cpuset.CPUSet {
 // core IDs in this CPUDetails.
 // TODO: CoreID is only unique within a socket. On multi-socket systems where CoreIDs repeat
 // across sockets, this method may return CPUs from multiple sockets for the same CoreID.
-// Consider refactoring core-level methods to use CoreKey{SocketID, CoreID} for correct
+// Consider refactoring core-level methods to use CoreKey{SocketID, ClusterID, CoreID} for correct
 // disambiguation, similar to CPUTopology.CPUsByCore.
 func (d CPUDetails) CPUsInCores(ids ...int) cpuset.CPUSet {
 	var cpuIDs []int
 	for _, id := range ids {
 		for cpu, info := range d {
 			if info.CoreID == id {
+				cpuIDs = append(cpuIDs, cpu)
+			}
+		}
+	}
+	return cpuset.New(cpuIDs...)
+}
+
+// CPUsInCoreKeys returns all logical CPUs associated with the given physical
+// core keys.
+func (d CPUDetails) CPUsInCoreKeys(keys ...CoreKey) cpuset.CPUSet {
+	var cpuIDs []int
+	for _, key := range keys {
+		for cpu, info := range d {
+			if info.CoreKey() == key {
 				cpuIDs = append(cpuIDs, cpu)
 			}
 		}
@@ -170,6 +223,14 @@ func (d CPUDetails) CoresInSockets(ids ...int) cpuset.CPUSet {
 	return cpuset.New(coreIDs...)
 }
 
+// CoreKeysInSockets returns all physical core keys associated with the given
+// socket IDs.
+func (d CPUDetails) CoreKeysInSockets(ids ...int) []CoreKey {
+	return d.coreKeysFor(func(info CPUInfo) bool {
+		return slices.Contains(ids, info.SocketID)
+	})
+}
+
 // NUMANodesInSockets returns all of the logical NUMANode IDs associated with
 // the given socket IDs in this CPUDetails.
 func (d CPUDetails) NUMANodesInSockets(ids ...int) cpuset.CPUSet {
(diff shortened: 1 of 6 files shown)


AutuSnow:
Hi @ffromani @pravk03, I pushed two small follow-up commits:
﻿
The PR scope is still limited to replacing inferred symmetric topology sizes with exact CPUDetails/CoreKey queries.  could you take another look when you have a chance?

ffromani:
> Hi @ffromani @pravk03, I pushed two small follow-up commits: ﻿ The PR scope is still limited to replacing inferred symmetric topology sizes with exact CPUDetails/CoreKey queries. could you take another look when you have a chance?

thanks. A bit of crazy days for me, but rest assured this PR is high on my list

pkg/cpumanager/cpu_assignment.go
```diff
@@ -96,7 +96,7 @@ type numaOrSocketsFirstFuncs interface {
 	takeFullSecondLevel()
 	sortAvailableNUMANodes() []int
 	sortAvailableSockets() []int
-	sortAvailableCores() []int
+	sortAvailableCores() []topology.CoreKey
```
ffromani, line 99: this is probably unavoidable (or requires so much effort to avoid which makes no sense) also considering the strict coupling we discussed in #35 between the cpu topology representation and the allocator code needs.

But effectively means we are forking the cpumanager codebase.
"forking" is not a value judgement, but rather a objective metric.
At first glance, I'm pretty convinced already this looks like a genuine incremental improvement.
But still it also means that we now drifted the implementation and that the cpumanager resync requires nonzero copy/paste effort (hence we forked).

IOW the value of the cpumanager borrowed code largely lies in the fact we can trivially (= mechanical changes with close to zero review necessary) import it in our codebase. If we cross the "nonzero backport boundary" whatever the reason (and again: this PR is *good* and a careful incremental upgrade, very hard to do nay better), the value proposition takes a plunge.

All that said, I still like this PR and I will take the time as soon as I can to do a proper review, but this also means that the discussion about adding a parallel allocator/moving past cpumanager becomes more urgent and a hot topic for 0.3.0, also including our plans for sub-NUMA allocations (e.g. #173).

I'll sketch a design proposal and file an enhancement issue to discuss the matter as soon as time allows (and review this PR meantime)

AutuSnow: I agree. This crosses the nonzero-backport boundary for `cpu_assignment.go`; after this change we should treat the allocator as intentionally diverged from a trivial kubelet resync.
﻿
For this PR I tried to keep the scope limited to the #35 correctness issue: preserve the current allocator/resource model, but replace symmetric inferred sizes with exact `CPUDetails`/`CoreKey` queries where the data already exists.😔

AutuSnow: I can take the first pass on that enhancement issue / design proposal if that helps.
﻿
While working on this PR I collected a few thoughts around the allocator boundary, and I think the proposal should be grounded in a few existing facts:
﻿
- Kubernetes CPUManager is already a node-local placement policy for workloads where cache affinity / scheduling latency matter, and the static policy has kept accumulating placement options such as NUMA distribution, socket alignment, core spreading, and uncore-cache alignment:
https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#configuring-cpu-management-policies
- DRA is also moving toward explicit accounting for native node resources through Node Allocatable Resources, where DRA devices can declare a footprint in standard resources like `cpu`:
https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#node-allocatable-resources
- The kernel already exposes richer topology than the old CPUManager representation models directly: package, die, cluster, core, sibling/core/cluster cpumasks, and online/offline CPU sets are all available through sysfs:
https://docs.kernel.org/admin-guide/cputopology.html
- In this repo, #35 already points at the need to stop inferring topology properties from symmetric counts, and #173 ties sub-NUMA / PCIe root locality to the need for an allocator that can consume finer-grained locality data:
https://github.com/kubernetes-sigs/dra-driver-cpu/issues/35
https://github.com/kubernetes-sigs/dra-driver-cpu/issues/173
﻿
My initial proposal direction would be to make the allocator boundary explicit instead of treating it as accidental drift:
﻿
1. keep this PR as a narrow correctness fix for #35, using exact `CPUDetails` / `CoreKey` queries;
2. document that `cpu_assignment.go` has crossed the trivial kubelet-resync boundary;
3. evaluate a future allocator interface where the current CPUManager-derived allocator remains the baseline, but a DRA-specific allocator can consume richer topology/locality inputs such as CoreKey, LLC/uncore, PCIe root locality, and possibly future node-allocatable-resource metadata;
4. keep resource model changes separate from this PR, so we can discuss them explicitly instead of growing the current patch.
﻿
I can draft that as an enhancement issue / design proposal and link it back here, #35, and #173. For this PR, I would still prefer to keep the scope limited to the asymmetric-topology correctness fix.

ffromani: Thanks for the offer. I also want to sketch out a design document with my thoughts and drive the design, which of course will need review from all the major contributors and will incorporate their input. We will probably need to iterate. If you have time now, feel free to sketch your doc: we will need cross-review and discussion anyway, and I can only see a final design taking input from all the parties. This is how community-driven project works are meant to work!

I think there are many intertwined questions we should explore - and the key point is _explore_: we don't *need*, I believe, answer or solutions for all of them, but the key is to make conscious decisions.

For example, a topic worth exploring is how much the allocator design is tied with the resource representation. In concrete terms, how much kubelet is tied to cadvisor's MachineInfo or `pkg/kubelet/cm/cpumanager/topology/...` or our allocator is tied to `pkg/cpuinfo/...`.

So far, all my experiments and past experience are slowly convincing me that, ultimately, the very question is wrong. Unless we can design a model/abstraction/representation clearly good, we should _not_ and just use 1:1 the linux kernel sysfs shape or perhaps the `hwloc` shape. As you correctly pointed out, the existing models are already rich (far richer than the kubelet model we kinda ineherited), and these are also stable and well understood from the larger community. A trivial golang struct mapping would give us all the data we need at a minimal cost, and we could spend time where it matters: consuming the data.

Another hot topic is how much the allocator logic and the internal representation drivers the shape of what we expose in the `resourceslice`. IOW: how much should we be compatible with the device shape we already exposed? Is it ok, or desirable, that changing the allocator change that shape? I don't have clear answers yet but I'm pretty convinced this is a GA blocker, so not really a theoretical but a rather practical concern.

I'd be happy to read your thoughts, write mine and forster the conversation.

pkg/cpuinfo/cpuinfo_utils.go
```diff
@@ -29,6 +31,33 @@ import (
...
+// CoreKey identifies a physical core within the topology. CoreID alone is not
+// globally unique on all systems, so callers that need core-level precision
+// should use this key instead of the raw CoreID.
+type CoreKey struct {
+	SocketID  int
+	ClusterID int
+	CoreID    int
+}
+
+func (k CoreKey) Less(other CoreKey) bool {
```
Tal-or, line 49: IIUC, `Less` means that from HW layout hierarchy, a `k` CoreKey is mareked with a lower number than `other` right?

Assuming the hierarchy looks something like:
```
[ Server Motherboard ]
       │
       ├──► Socket 0 (Physical CPU 1)
       │     │
       │     ├──► Cluster 0 (e.g., CCX / L3 Cache Group)
       │     │     ├──► Core 0
       │     │     ├──► Core 1
       │     │     └──► Core 2
       │     │
       │     └──► Cluster 1
       │           ├──► Core 3
       │           └──► Core 4
       │
       └──► Socket 1 (Physical CPU 2)
             │
             └──► Cluster 0
                   ├──► Core 0
                   └──► Core 1
```
Shouldn't the function check the Socket ID first?

For example:

Core 8 socket 0  is "less" than core 2 socket 1

is that correct or am I missing something here?

pkg/cpuinfo/cpuinfo_utils.go
```diff
@@ -80,6 +109,20 @@ func (d CPUDetails) CPUsInCores(ids ...int) cpuset.CPUSet {
 	return cpuset.New(cpuIDs...)
 }
 
+// CPUsInCoreKeys returns all logical CPUs associated with the given physical
+// core keys.
+func (d CPUDetails) CPUsInCoreKeys(keys ...CoreKey) cpuset.CPUSet {
```
Tal-or, line 124: Is there a special reasons why all receivers are copy and not pointers? i.e `(d *CPUDetails)`?

pkg/cpumanager/cpu_assignment.go
```diff
@@ -597,7 +600,7 @@ func (a *cpuAccumulator) takePartialUncore(uncoreID int) {
 		"uncore", uncoreID,
 		"claimed", claimed,
 		"needed", a.numCPUsNeeded,
-		"cores", freeCores.String(),
+		"cores", freeCores,
```
Tal-or, line 603: Maybe it's worth to implement `String()` function for `CoreKey` to make the output more structured and clear

pkg/cpuinfo/cpuinfo_utils.go
```diff
@@ -221,3 +303,32 @@ func (d CPUDetails) coresInUncoreCache(ids ...int) cpuset.CPUSet {
...
+			if info.UncoreCacheID == id {
+				return true
+			}
+		}
+		return false
+	})
+}
+
+func (d CPUDetails) coreKeysFor(include func(CPUInfo) bool) []CoreKey {
+	seen := map[CoreKey]struct{}{}
```
Tal-or, line 316: you may use `set` from:
"k8s.io/apimachinery/pkg/util/sets"

and then

```
func (d CPUDetails) coreKeysFor(include func(CPUInfo) bool) []CoreKey {
	// 1. Initialize the Kubernetes generic set
	seen := sets.New[CoreKey]()
	
	// 2. Filter and insert directly into the set
	for _, info := range d {
		if include(info) {
			seen.Insert(info.CoreKey())
		}
	}

	// 3. Extract the underlying slice directly from the set
	coreKeys := seen.UnsortedList()

	// 4. Sort the keys deterministically
	sort.Slice(coreKeys, func(i, j int) bool {
		return coreKeys[i].Less(coreKeys[j])
	})
	
	return coreKeys
}
```

Tal-or (commented):
Thank you @AutuSnow,
This PR could really help with the problem I've described in:
https://github.com/kubernetes-sigs/dra-driver-cpu/issues/35#issuecomment-4798238891 

Small comments from my side

pkg/cpuinfo/cpuinfo_utils.go
AutuSnow: `Less()` here is only used as a deterministic tie-breaker once the higher-level topology grouping has already been decided by the surrounding allocator logic.
﻿
I tried changing it to a strict physical-hierarchy order (`socket -> cluster -> core`), but that ended up changing an existing allocator decision in `pkg/cpumanager` rather than just making the ordering clearer. Because of that I’d prefer to keep the current tie-break semantics in this PR and avoid changing allocation behavior as part of the asymmetric-topology fix.

pkg/cpuinfo/cpuinfo_utils.go
AutuSnow: These helpers are intentionally read-only, so I kept value receivers.
﻿
For `CPUDetails`, the receiver is a map, so using a value receiver only copies the map header, not the underlying data. For `CPUInfo.CoreKey()`, the receiver is a small value object and the method does not mutate state, so a pointer receiver would not buy us much here.

pkg/cpuinfo/cpuinfo_utils.go
AutuSnow: That would work too, but I kept the local `map[CoreKey]struct{}` here because the method still needs an explicit deterministic sort afterwards, and the plain map keeps that dedupe step very direct.
﻿
So for this PR I’d prefer to keep the current implementation unless we want a broader consistency cleanup across similar helpers.

pkg/cpumanager/cpu_assignment.go
AutuSnow: Fixed. Added `CoreKey.String()` so the debug output is easier to read.

thc1006:
Read through this on top of the latest main, built it and ran the cpuinfo and cpumanager tests locally, all green. The `CoreKey` unification is a nice cleanup, and swapping the `CPUsPerSocket`/`CPUsPerCore`/`CPUsPerUncore` assumptions for exact `CPUDetails` queries is the right direction for #35. Wiring the three asymmetric fixtures into real allocator tests is what makes it convincing rather than just mechanical.

One case I wanted to check on the `takePartialUncore` path. In the new `topoAsymmetricUncoreSMT` test, a 3-CPU request resolves to `{0,2,5}`, which spans uncore 0 and uncore 1 even though uncore 0 on its own has 4 free CPUs and could hold all three. Tracing it: `sortAvailableUncoreCaches` sorts fewest-free-first, so uncore 1 (2 CPUs) is visited first, `needsAtLeast(2)` is true for a 3-CPU request, so `takeFullUncore` claims uncore 1 and the remaining CPU spills into uncore 0. On a symmetric topology this never shows up because the uncores are the same size, the asymmetric sizes surface it. Since the point of this path is uncore locality, is `{0,2,5}` the behavior we want to lock in, or a known best-effort limitation worth a short comment next to the expected value? Not blocking, I mostly want the test to encode intent rather than current output.

Smaller thing: after this change `CPUsPerCore` / `CPUsPerSocket` / `CPUsPerUncore` have no in-tree callers left. The updated doc comment makes sense if you want to keep them available, otherwise they could be dropped in a follow-up.

Happy to be corrected on the uncore point if I am missing a step in the accumulator.

AutuSnow:
I re-checked the path and I think your tracing is right.

My intent here was to keep the allocator policy unchanged and only replace the symmetric topology assumptions with exact `CPUDetails` queries. So in this case `{0,2,5}` is documenting the current best-effort heuristic on an asymmetric uncore topology, not asserting a stronger “stay within one uncore whenever possible” rule. I added a short note next to the expected value to make that intent explicit.

Agree on `CPUsPerCore` / `CPUsPerSocket` / `CPUsPerUncore` too; I’d prefer to keep that as follow-up cleanup rather than widen this PR.

thc1006:
Makes sense, and that is the intent I was after. The note in `5e5bff9` makes it explicit that `{0,2,5}` is the current best-effort heuristic on an asymmetric uncore topology rather than a rule to stay within one uncore whenever possible, which is exactly the distinction I was not sure the test was capturing. Thanks for re-checking the path.

On `CPUsPerCore` / `CPUsPerSocket` / `CPUsPerUncore`, happy to send that cleanup as a follow-up once this lands so it stays out of this PR's scope.

PR #212, opened 2026-06-29
Chore: release and build consistency fixes

AutuSnow:
## What type of PR is this?
﻿
/kind cleanup
/kind documentation
﻿
## What this PR does / why we need it
﻿
- updates GitHub Actions to read the Go version from `go.mod`
- aligns the Helm chart default image registry with released artifacts
- fixes the release branch naming in `RELEASE.md` so it matches the release automation workflow
- makes `test/image/dracputester` build on non-Linux hosts by splitting Linux-specific affinity logic from shared code
﻿
These changes reduce configuration drift and make local builds work reliably on macOS hosts.
﻿
## Which issue(s) this PR is related to
﻿
N/A
﻿
## Special notes for reviewers
N/A
﻿
## Release note
﻿
```release-note
N/A
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
       with:
-        go-version: '1.25'
+        go-version-file: go.mod
     - name: Build
       run: make build
     - name: Test
@@ -31,7 +31,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
         with:
-          go-version: '1.25'
+          go-version-file: go.mod
       - name: Install golangci-lint
         run: make install-golangci-lint
       - name: Run golangci-lint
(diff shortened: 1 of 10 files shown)


ffromani:
LGTM,

will approve if @fmuyassarov is happy as well

fmuyassarov (commented):
Thanks, LGTM. 
nit, I wish we could use 2026 on the licensing header or even better no date at all as we do it now. But it doesn't have to be done here since other test files are also on 2025.

/lgtm

AutuSnow:
/hold
Ah, yes, I'm sorry I overlooked this issue when copying the license

ffromani:
> Thanks, LGTM. nit, I wish we could use 2026 on the licensing header or even better no date at all as we do it now. But it doesn't have to be done here since other test files are also on 2025.

In this specific case I think what @AutuSnow did is correct because they moved around code which already had a copyright year. Should that have been new code, I'd have asked to remove the copyright year, but for code movement with no additions I tend to believe preserving the original copyright boilerplate verbatim is the correct thing.

Little harm even if this is "wrong", anyway.

/approve

ffromani:
removing the year is also fine by me btw.

AutuSnow:
Thank you for the reminder. I have removed the year, please review again

fmuyassarov:
/lgtm 
thanks

Issue #142, opened 2026-05-12
PLANNING: Release 0.3.0


ffromani:
## Introduction
We keep the simple planning model (#34, #57). It's lightweight enough which is serving us well in our initial stages.

## Highlights
0.3.0 would be the first beta version.
This version will support only individual mode, and point to the bootstrapped `cpumem` driver for users which want to use the grouped mode. Therefore, having a running `cpumem` repo to point users to is a prerequisite for this release.

### Note
We prefer issues as tracking mechanism. PRs will fix/implement issues. We will list explictly the rare PRs which don't have a corresponding issue. Please file issues for tracking and discussion as default contribution vehicle.

PRs and issues already merged at time of updating are not listed.

## Roadmap-affecting items to figure out before release (0.3.0 blockers)

* #164 - DECISION: go/nogo/postpone
* #152 - hard depends on #164 promotion and revised decision on #112 

SUMMARY: #304 seems to be the solution external integrators/scheduler authors like, #112 is likely to be **abandoned**

## Issues to fix (not in priority order)
 
* #304 
* #299 (janitorial)
* #296 (docs)
* #266 
* #264
* #245 
* #234 
* #233 
* #231 
* #226 
* #219
* #218 
* #196
* #191 
* #185 and its fail-close followups
* #181 
* #174 
* #63 (the parts not covered by #143 )
* #48 (PR: #213)

## PRs to include (not in priority order)
* #249 
* #76

## Nice to have - we can postpone those
* #60 
* #173
* #161

ffromani:
@pravk03 @AutuSnow @catblade @fmuyassarov 

the 0.3.0 cycle opens, as usual please review the current list and feel free to comment suggesting things to add/remove

ffromani:
PROPOSAL: let us release 0.3.0 _after_ kubernetes 1.37 is out so we can consume KEP 5517 alpha2. Even with the usual caveats and rough edges stemming from the fact that the KEP implementation is alpha (e.g. need FG enablement) we will still be in a much better place and reduce the friction in consuming the driver significantly.

pravk03:
+1. I like the idea. This driver can be one of the use case to validate the KEP during alpha, will can help us gather feedback to shape the beta.

ffromani:
[...]

PR #213, opened 2026-06-29
feat: add configuration file support

### LINKED ISSUE #48: Configuration file support
We are close to the tipping point for command line options. At the beginning, it's easy and convenient to add command line options to tune the behavior of the software but unless we enforce strict discipline, it's easy to just add them till, quite too soon, we have a plethora of options which makes command lines long and confusing and harder than necessary to programmatically set.

If caught early, it's just easier to switch to a config file support. If we agree to turn the ship and prioritize/require configuration files over than command line options, we should consider the following

- widespread format like yaml or toml (json is often less suitable)
- configlet support from day 1
- optional aliasing of config file keys to commandline options
- versioned configuration and backward/forward compatibility is a must (and must be tested from the beginning)

fmuyassarov:
#### What type of PR is this?
/kind feature

Adds configuration file support. The driver now accepts `--config` pointing to a YAML file. Values from the file fill in any CLI flag not explicitly set. Deployments without `--config` are not affected. When deploying with helm, `driverConfig` in values.yaml is the primary way to pass configuration.  The chart serializes the map into yaml and stores it in a configMap, mounts it into the driver container and then passes `--config` automatically. Invalid values in the config file (unknown cpuDeviceMode, groupBy, or apiVersion) are rejected at startup. The helm schema also validates `driverConfig` fields at install time.

Fixes #48 

I've been going back and forth on the value this change brings vs the additional complexity it introduces, especially considering the current state of the project.

I can definitely see the value of introducing a config file, even though I wasn't sure fully while implementing this. My hesitation wasn't because I think it's a bad idea, but because we already have a mechanism for users to provide driver configuration through helm. For someone not aware of the follow up discussion, adding another configuration mechanism may seem unnecessary, especially since driver configuration can already be provided either through individual flags or the `driverConfig` flag.

That said, I do think this change adds value in the long run. We expect the number of configuration options to grow, whether because of internal changes or future changes to the DRA API. As that happens, the number of flags will inevitably grow as well. While we could continue exposing everything as flags and ask users to pass values through standard helm `--set` options, I think that eventually becomes a poor user experience from UX point.

@ffromani made a good point [here](https://github.com/kubernetes-sigs/dra-driver-cpu/issues/142#issuecomment-4806721823) about dropping the standalone flags in favor of a single configuration file. That argument ultimately convinced me as well that this direction makes sense.

So, what comes next? If everyone agrees, my proposal would be to remove the individual flags (which should be fairly straightforward after this change) and keep `driverConfig` as the single source of truth. Users would then be able to inspect the effective configuration either from the mounted ConfigMap or from the logs.

The remaining question is when we should make that transition. I'd really like to hear everyone's thoughts on this so we can pick a good time to do it without hurting the users. And sorry for the late submission.

#### Release note

```release-note
Add config file support via --config flag. When deploying with Helm, use driverConfig in values.yaml to pass driver settings as a map. The chart creates a ConfigMap and wires --config automatically behind the scenes. 
```

#### Documentation updates
- README: added configuration file section
- Helm chart README: updated with driverConfig usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ You need to disable the CPUManager on the nodes you wish to run this DRA driver.
 
 ## Driver configuration
 
+### Command-line flags
+
 The driver can be configured with the following command-line flags:
 
 - `--cpu-device-mode`: Sets the mode for exposing CPU devices.
@@ -38,13 +40,40 @@ When deploying with Helm, set `driverConfig` in your values file. The chart seri
 map to YAML, stores it in a ConfigMap, mounts it as `/etc/dracpu/config.yaml` inside the
 driver container, and passes `--config` automatically.
 
+#### Schema
+
+The config file is a **flat** YAML map - there are no nested groups. All fields are optional
+except where noted. Unknown fields are rejected at startup to catch typos early.
+
+| Field              | Type   | Default     | Description                                                                              |
+| ------------------ | ------ | ----------- | ---------------------------------------------------------------------------------------- |
+| `apiVersion`       | string | *(omitted)* | When present, must be `v1alpha1`. Rejected otherwise.                                    |
+| `cpuDeviceMode`    | string | `grouped`   | CPU exposure mode: `individual` or `grouped`.                                            |
+| `groupBy`          | string | `numanode`  | Grouping strategy when `cpuDeviceMode` is `grouped`: `numanode`, `socket`, or `machine`. |
+| `reservedCPUs`     | string | *(none)*    | CPUs excluded from allocation, e.g. `"0-1"`.                                             |
+| `hostnameOverride` | string | *(none)*    | Override the node hostname the driver registers under.                                   |
+| `exposePCIeRoots`  | bool   | `false`     | Add PCIe root attributes to CPU devices (requires `DRAListTypeAttributes` feature gate). |
+| `kubeconfig`       | string | *(none)*    | Path to a kubeconfig file (for out-of-cluster use).                                      |
+
+#### Versioning and backward compatibility
+
+The schema is versioned via the optional `apiVersion` field (currently `v1alpha1`). The layout
+is intentionally flat for now. If a nested hierarchy is introduced in the future, the
+`apiVersion` field will be bumped so that older config files continue to be accepted or produce a
+an error.
+
+#### Example
+
 ```yaml
 # values.yaml
 driverConfig:
+  apiVersion: v1alpha1
   cpuDeviceMode: grouped
   groupBy: numanode
   reservedCPUs: "0-3"
 ```
+
+The `apiVersion` field is optional. When present it must equal `v1alpha1` (the current version); any other value is rejected at startup.
 
 `driverConfig` is a single map covering all driver settings — there is no separate Helm value
 per field. `args.*` on the other hand exposes individual fields as explicit Helm values and
--- a/README.md
+++ b/README.md
@@ -32,6 +32,29 @@ The driver can be configured with the following command-line flags:
 - `--reserved-cpus`: Specifies a set of CPUs to be reserved for system and kubelet processes. These CPUs will not be allocatable by the DRA driver and would be excluded from the `ResourceSlice`. The value is a cpuset, e.g., `0-1`. This semantic is the same as the one the kubelet applies with its `static` CPU Manager policy and enabling [`strict-cpu-reservation`](https://kubernetes.io/blog/2024/12/16/cpumanager-strict-cpu-reservation/) flag and specifying the CPUs with the [`reservedSystemCPUs`](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#explicitly-reserved-cpu-list) to be reserved for system daemons. For correct CPU accounting, the number of CPUs reserved with this flag should match the sum of the kubelet's `kubeReserved` and `systemReserved` settings. This ensures the kubelet subtracts the correct number of CPUs from `Node.Status.Allocatable`.
 - `--expose-pcie-roots`: If enabled, adds the "resource.kubernetes.io/pcieRoot" standard value to CPU devices, to report the PCIe roots close to each device. Since it always reports values as list, this option requires the cluster Feature Gate `DRAListTypeAttributes` (see KEP 5491) to be enabled. The driver has no way to introspect the cluster Feature Gate, so care must be taken to enable first the Feature Gate then this option.
 
+### Configuration file
+
+When deploying with Helm, set `driverConfig` in your values file. The chart serializes the
+map to YAML, stores it in a ConfigMap, mounts it as `/etc/dracpu/config.yaml` inside the
+driver container, and passes `--config` automatically.
+
+```yaml
+# values.yaml
+driverConfig:
+  cpuDeviceMode: grouped
+  groupBy: numanode
+  reservedCPUs: "0-3"
+```
+
+`driverConfig` is a single map covering all driver settings — there is no separate Helm value
+per field. `args.*` on the other hand exposes individual fields as explicit Helm values and
+translates them to CLI flags. Both reach the same driver settings; `args.*` takes priority when
+both are set for the same field.
+
+Both `args.*` and `driverConfig` exist during a transition period. The intent is to eventually
+deprecate `args.*` in favour of `driverConfig` as the single configuration mechanism. The driver
+logs the effective configuration at startup so you can verify which values are active.
+
 ## How it Works
 
 The driver is deployed as a DaemonSet which contains two core components:
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -53,10 +53,18 @@ const (
 
 var (
 	driverFlags = driverconfig.Default()
-	ready       atomic.Bool
+	configFile  string
+
+	ready atomic.Bool
 )
 
 func init() {
+	// --config is kept outside Config to avoid a self-referential loop,
+	// following the convention used by kubeadm and similar tools.
+	flag.StringVar(&configFile, "config", "",
+		"Path to a YAML driver configuration file. Values in the file "+
+			"are applied for any flag not explicitly provided on the command line. "+
+			"If empty, only CLI flags and built-in defaults are used.")
 	driverFlags.AddFlags(flag.CommandLine)
 }
 
@@ -85,26 +93,32 @@ func main() {
 
 	logger := ctxlog.Setup()
 
-	if err := runDriver(logger); err != nil {
+	cfg, err := driverconfig.Load(driverFlags, configFile, flag.CommandLine)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "dracpu: failed to load configuration: %v\n", err)
+		os.Exit(1)
+	}
+
+	if err := runDriver(logger, cfg); err != nil {
 		os.Exit(1)
 	}
 }
 
-func runDriver(logger logr.Logger) error {
-	if err := run(logger); err != nil {
+func runDriver(logger logr.Logger, cfg driverconfig.Config) error {
+	if err := run(logger, cfg); err != nil {
 		logger.Error(err, "failed to run")
 		return err
 	}
 	return nil
 }
 
-func run(logger logr.Logger) error {
+func run(logger logr.Logger, cfg driverconfig.Config) error {
 	printVersion(logger)
-	flag.VisitAll(func(f *flag.Flag) {
-		logger.Info("FLAG", "name", f.Name, "value", f.Value.String())
-	})
+	logger.Info("CONFIG", "bindAddress", cfg.BindAddress, "cpuDeviceMode", cfg.CPUDeviceMode,
+		"groupBy", cfg.GroupBy, "reservedCPUs", cfg.ReservedCPUs,
+		"hostnameOverride", cfg.HostnameOverride, "exposePCIeRoots", cfg.ExposePCIeRoots)
 
-	reservedCPUSet, err := cpuset.Parse(driverFlags.ReservedCPUs)
+	reservedCPUSet, err := cpuset.Parse(cfg.ReservedCPUs)
 	if err != nil {
 		return fmt.Errorf("failed to parse reserved CPUs: %w", err)
 	}
@@ -126,7 +140,7 @@ func run(logger logr.Logger) error {
 	// Add metrics handler
 	mux.Handle("/metrics", promhttp.Handler())
 	server := &http.Server{
-		Addr:              driverFlags.BindAddress,
+		Addr:              cfg.BindAddress,
 		Handler:           mux,
 		IdleTimeout:       120 * time.Second,
 		ReadTimeout:       10 * time.Second,
@@ -141,8 +155,8 @@ func run(logger logr.Logger) error {
 	}()
 
 	var restConfig *rest.Config
-	if driverFlags.Kubeconfig != "" {
-		restConfig, err = clientcmd.BuildConfigFromFlags("", driverFlags.Kubeconfig)
+	if cfg.Kubeconfig != "" {
+		restConfig, err = clientcmd.BuildConfigFromFlags("", cfg.Kubeconfig)
 	} else {
 		// creates the in-cluster config
 		restConfig, err = rest.InClusterConfig()
@@ -162,7 +176,7 @@ func run(logger logr.Logger) error {
 		return fmt.Errorf("can not create client-go client: %w", err)
 	}
 
-	nodeName, err := nodeutil.GetHostname(driverFlags.HostnameOverride)
+	nodeName, err := nodeutil.GetHostname(cfg.HostnameOverride)
 	if err != nil {
 		return fmt.Errorf("can not obtain the node name, use the hostname-override flag if you want to set it to a specific value: %w", err)
 	}
@@ -183,9 +197,9 @@ func run(logger logr.Logger) error {
 		DriverName:       driverName,
 		NodeName:         nodeName,
 		ReservedCPUs:     reservedCPUSet,
-		CPUDeviceMode:    driverFlags.CPUDeviceMode,
-		CPUDeviceGroupBy: driverFlags.GroupBy,
-		ExposePCIeRoots:  driverFlags.ExposePCIeRoots,
+		CPUDeviceMode:    cfg.CPUDeviceMode,
+		CPUDeviceGroupBy: cfg.GroupBy,
+		ExposePCIeRoots:  cfg.ExposePCIeRoots,
 		Metrics:          cpumetrics.New(prometheus.DefaultRegisterer),
 	}
 	driverProviders := driver.Providers{
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -62,8 +62,9 @@ func init() {
 	// --config is kept outside Config to avoid a self-referential loop,
 	// following the convention used by kubeadm and similar tools.
 	flag.StringVar(&configFile, "config", "",
-		"Path to a YAML driver configuration file. Values in the file "+
-			"are applied for any flag not explicitly provided on the command line. "+
+		"Path to a YAML driver configuration file. Configuration values are applied in order: "+
+			"built-in defaults, then file values, then explicit CLI flags. "+
+			"Only values explicitly set on the command line override earlier layers. "+
 			"If empty, only CLI flags and built-in defaults are used.")
 	driverFlags.AddFlags(flag.CommandLine)
 }
@@ -93,9 +94,9 @@ func main() {
 
 	logger := ctxlog.Setup()
 
-	cfg, err := driverconfig.Load(driverFlags, configFile, flag.CommandLine)
+	cfg, err := driverconfig.Load(driverFlags, configFile, flag.CommandLine, logger)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "dracpu: failed to load configuration: %v\n", err)
+		logger.Error(err, "failed to load configuration")
 		os.Exit(1)
 	}
 
@@ -114,16 +115,14 @@ func runDriver(logger logr.Logger, cfg driverconfig.Config) error {
 
 func run(logger logr.Logger, cfg driverconfig.Config) error {
 	printVersion(logger)
-	logger.Info("CONFIG", "bindAddress", cfg.BindAddress, "cpuDeviceMode", cfg.CPUDeviceMode,
-		"groupBy", cfg.GroupBy, "reservedCPUs", cfg.ReservedCPUs,
-		"hostnameOverride", cfg.HostnameOverride, "exposePCIeRoots", cfg.ExposePCIeRoots)
+	logger.Info("CONFIG", cfg.LogValues()...)
 
 	reservedCPUSet, err := cpuset.Parse(cfg.ReservedCPUs)
 	if err != nil {
 		return fmt.Errorf("failed to parse reserved CPUs: %w", err)
 	}
 
-	sfs, err := newSysFS(logger, driverFlags.SysFSOverlay)
+	sfs, err := newSysFS(logger, cfg.SysFSOverlay)
 	if err != nil {
 		return err
 	}
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -35,13 +35,13 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Affinity rules for scheduling the DaemonSet pods |
-| args.cpuDeviceMode | string | `"grouped"` | CPU exposure mode: `grouped` (expose NUMA nodes or sockets as devices) or `individual` (expose each CPU as a device) |
+| args.cpuDeviceMode | string | `""` | CPU exposure mode: `grouped` (expose NUMA nodes or sockets as devices) or `individual` (expose each CPU as a device); when empty, defers to `driverConfig.cpuDeviceMode` or the built-in default (`grouped`) |
 | args.exposePCIeRoots | bool | `false` | Discover and expose PCIe roots as device attributes. Requires the `DRAListTypeAttributes=true` feature gate in the cluster |
-| args.groupBy | string | `"numanode"` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode`, `socket` or `machine` |
+| args.groupBy | string | `""` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode`, `socket` or `machine`; when empty, defers to `driverConfig.groupBy` or the built-in default (`numanode`) |
 | args.hostnameOverride | string | `""` | Override the node name the driver registers under; omitted when empty |
 | args.logLevel | int | `4` | Log verbosity level passed as `--v` |
 | args.reservedCPUs | string | `""` | CPUs reserved for the OS and kubelet, excluded from DRA management (e.g. `"0-1"`); omitted when empty |
-| driverConfig | object | `{}` | Driver config file contents. When non-empty, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. CLI flags (args.*) always take priority over values set here. Example:   driverConfig:     cpuDeviceMode: grouped     groupBy: numanode     reservedCPUs: "0-3" |
+| driverConfig | object | `{}` | Driver config file contents. When non-empty, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. Fields in `args.*` that are non-empty always take priority over values set here. Example:   driverConfig:     cpuDeviceMode: individual     groupBy: socket     reservedCPUs: "0-3" |
 | extraArgs | list | `[]` | Extra command-line arguments appended to the driver arguments |
 | extraVolumeMounts | list | `[]` | Extra volume mounts for the driver container |
 | extraVolumes | list | `[]` | Extra volumes for the DaemonSet pod |
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -36,7 +36,7 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Affinity rules for scheduling the DaemonSet pods |
 | args.cpuDeviceMode | string | `""` | CPU exposure mode: `grouped` (expose NUMA nodes or sockets as devices) or `individual` (expose each CPU as a device); when empty, defers to `driverConfig.cpuDeviceMode` or the built-in default (`grouped`) |
-| args.exposePCIeRoots | bool | `false` | Discover and expose PCIe roots as device attributes. Requires the `DRAListTypeAttributes=true` feature gate in the cluster |
+| args.exposePCIeRoots | bool | `false` | Discover and expose PCIe roots as device attributes. Requires the `DRAListTypeAttributes=true` feature gate in the cluster. Not configurable via `driverConfig`; use this flag instead |
 | args.groupBy | string | `""` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode`, `socket` or `machine`; when empty, defers to `driverConfig.groupBy` or the built-in default (`numanode`) |
 | args.hostnameOverride | string | `""` | Override the node name the driver registers under; omitted when empty |
 | args.logLevel | int | `4` | Log verbosity level passed as `--v` |
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -41,6 +41,7 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
 | args.hostnameOverride | string | `""` | Override the node name the driver registers under; omitted when empty |
 | args.logLevel | int | `4` | Log verbosity level passed as `--v` |
 | args.reservedCPUs | string | `""` | CPUs reserved for the OS and kubelet, excluded from DRA management (e.g. `"0-1"`); omitted when empty |
+| driverConfig | object | `{}` | Driver config file contents. When non-empty, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. CLI flags (args.*) always take priority over values set here. Example:   driverConfig:     cpuDeviceMode: grouped     groupBy: numanode     reservedCPUs: "0-3" |
 | extraArgs | list | `[]` | Extra command-line arguments appended to the driver arguments |
 | extraVolumeMounts | list | `[]` | Extra volume mounts for the driver container |
 | extraVolumes | list | `[]` | Extra volumes for the DaemonSet pod |
--- a/deployment/helm/dra-driver-cpu/driverconfig.schema.json
+++ b/deployment/helm/dra-driver-cpu/driverconfig.schema.json
@@ -1,0 +1,33 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "driverconfig.schema.json",
+  "$comment": "Schema for driverConfig values. Mirrors internal/driverconfig/config.go Config struct, except bindAddress which is intentionally excluded — use the healthzPort Helm value instead. Update this file when fields are added or removed from Config.",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "apiVersion": {
+      "type": "string",
+      "enum": ["v1alpha1"]
+    },
+    "kubeconfig": {
+      "type": "string"
+    },
+    "hostnameOverride": {
+      "type": "string"
+    },
+    "reservedCPUs": {
+      "type": "string"
+    },
+    "cpuDeviceMode": {
+      "type": "string",
+      "enum": ["grouped", "individual"]
+    },
+    "groupBy": {
+      "type": "string",
+      "enum": ["numanode", "socket", "machine"]
+    },
+    "exposePCIeRoots": {
+      "type": "boolean"
+    }
+  }
+}
--- a/deployment/helm/dra-driver-cpu/driverconfig.schema.json
+++ b/deployment/helm/dra-driver-cpu/driverconfig.schema.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "driverconfig.schema.json",
-  "$comment": "Schema for driverConfig values. Mirrors internal/driverconfig/config.go Config struct, except bindAddress which is intentionally excluded — use the healthzPort Helm value instead. Update this file when fields are added or removed from Config.",
+  "$comment": "Schema for driverConfig values. Mirrors internal/driverconfig/config.go Config struct, except bindAddress, exposePCIeRoots, and showMetrics, whose config-file values are ignored — use healthzPort, args.exposePCIeRoots, and --show-metrics instead. Update this file when fields are added or removed from Config.",
   "type": "object",
   "additionalProperties": false,
   "properties": {
@@ -26,8 +26,8 @@
       "type": "string",
       "enum": ["numanode", "socket", "machine"]
     },
-    "exposePCIeRoots": {
-      "type": "boolean"
+    "sysfsOverlay": {
+      "type": "string"
     }
   }
 }
--- a/deployment/helm/dra-driver-cpu/templates/configmap.yaml
+++ b/deployment/helm/dra-driver-cpu/templates/configmap.yaml
@@ -1,0 +1,26 @@
+# Copyright The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+{{- if .Values.driverConfig }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "dra-driver-cpu.fullname" . }}-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dra-driver-cpu.labels" . | nindent 4 }}
+data:
+  config.yaml: |
+    {{- .Values.driverConfig | toYaml | nindent 4 }}
+{{- end }}
--- a/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
+++ b/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
@@ -64,8 +64,12 @@ spec:
         args:
           - /dracpu
           - --v={{ .Values.args.logLevel }}
+          {{- if .Values.args.cpuDeviceMode }}
           - --cpu-device-mode={{ .Values.args.cpuDeviceMode }}
+          {{- end }}
+          {{- if .Values.args.groupBy }}
           - --group-by={{ .Values.args.groupBy }}
+          {{- end }}
           {{- if .Values.healthzPort }}
           - --bind-address=:{{ .Values.healthzPort }}
           {{- end }}
--- a/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
+++ b/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
@@ -79,9 +79,7 @@ spec:
           {{- if .Values.args.hostnameOverride }}
           - --hostname-override={{ .Values.args.hostnameOverride }}
           {{- end }}
-          {{- if .Values.args.exposePCIeRoots }}
-          - --expose-pcie-roots
-          {{- end }}
+          - --expose-pcie-roots={{ .Values.args.exposePCIeRoots }}
           {{- with .Values.extraArgs }}
           {{- toYaml . | nindent 10 }}
           {{- end }}
--- a/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
+++ b/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
@@ -34,10 +34,13 @@ spec:
         {{- with .Values.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.podAnnotations }}
       annotations:
+        {{- if .Values.driverConfig }}
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- end }}
+        {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- end }}
     spec:
       hostNetwork: true
       priorityClassName: system-node-critical
@@ -78,6 +81,9 @@ spec:
           {{- with .Values.extraArgs }}
           {{- toYaml . | nindent 10 }}
           {{- end }}
+          {{- if .Values.driverConfig }}
+          - --config=/etc/dracpu/config.yaml
+          {{- end }}
         image: {{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
@@ -115,6 +121,11 @@ spec:
         {{- with .Values.extraVolumeMounts }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
+        {{- if .Values.driverConfig }}
+        - name: driver-config
+          mountPath: /etc/dracpu
+          readOnly: true
+        {{- end }}
       volumes:
       - name: device-plugin
         hostPath:
@@ -131,4 +142,9 @@ spec:
           type: DirectoryOrCreate
       {{- with .Values.extraVolumes }}
       {{- toYaml . | nindent 6 }}
+      {{- end }}
+      {{- if .Values.driverConfig }}
+      - name: driver-config
+        configMap:
+          name: {{ include "dra-driver-cpu.fullname" . }}-config
       {{- end }}
--- a/internal/driverconfig/cfgfile.go
+++ b/internal/driverconfig/cfgfile.go
@@ -17,6 +17,7 @@ limitations under the License.
 package driverconfig
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"maps"
@@ -74,13 +75,16 @@ func mergeMap(dst, src map[string]any) {
 }
 
 // applyMap applies only the keys present in m to cfg; absent keys are
-// untouched (encoding/json.Unmarshal semantics).
+// untouched (encoding/json.Unmarshal semantics). Unknown keys are rejected
+// to catch typos early rather than silently ignoring them.
 func applyMap(cfg *Config, m map[string]any) error {
 	data, err := json.Marshal(m)
 	if err != nil {
 		return fmt.Errorf("marshaling config map: %w", err)
 	}
-	if err := json.Unmarshal(data, cfg); err != nil {
+	dec := json.NewDecoder(bytes.NewReader(data))
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(cfg); err != nil {
 		return fmt.Errorf("applying config map: %w", err)
 	}
 	return nil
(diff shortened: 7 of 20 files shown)


fmuyassarov:
One thing I forgot to mention is - regarding the drop in configs. It's the same battle I had with myself as I raised in the description. I'm still leaning towards not adding them, but if you see value in having them for the driver at this stage, I'm happy to include them. I still have the commit ready.

internal/driverconfig/load.go
```diff
@@ -0,0 +1,71 @@
...
+	})
+
+	confMap := map[string]any{}
+	if err := buildConfMap(confMap, filePath); err != nil {
+		return Config{}, fmt.Errorf("config file %q: %w", filePath, err)
+	}
+
+	// CLI-explicit flags win; drop their keys so the file doesn't override them.
+	for jsonKey := range explicitJSONKeys {
+		delete(confMap, jsonKey)
```
AutuSnow, line 67: Why is this place so strange? When the user only sets driverControl: {cpuDeviceMode: individual} and ARGs to default, the driver actually runs' grouped 'without any error. This is exactly the field promoted by PR as the "main configuration method" for helm, which is equivalent to the silent failure of core functions on the main deployment path, exposing incorrect device topology. Only fields that are conditionally rendered (reserved CPUs/hostnameOverride/exposePCIeRots, {{- if}} package) will respect the driverConfig when ARGs is empty. Isn't that driverConfig silently disabling cpuDeviceMode/groupBy/bindAddress?

internal/driverconfig/cfgfile.go
```diff
@@ -0,0 +1,87 @@
...
+func applyMap(cfg *Config, m map[string]any) error {
+	data, err := json.Marshal(m)
+	if err != nil {
+		return fmt.Errorf("marshaling config map: %w", err)
+	}
+	if err := json.Unmarshal(data, cfg); err != nil {
+		return fmt.Errorf("applying config map: %w", err)
+	}
+	return nil
+}
```
AutuSnow, line 86: ApplyMap uses json. Unmarshal to apply to Config, but the json library ignores unknown fields by default. Writing the wrong key (
TotallyBogusField: 123 → Silent acceptance, no error reported
Reserved CPUss: "4-7" → One more letter, silently ignored, reserved CPUs remains default
) will not result in an error, and the fields will remain default. Helm side driverConfig is a bare "type": "object" in the schema, so there is no blocking during Helm installation

internal/driverconfig/flags.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("invalid cpuDeviceMode %q: must be %q or %q",
+			c.CPUDeviceMode, driver.CPU_DEVICE_MODE_GROUPED, driver.CPU_DEVICE_MODE_INDIVIDUAL)
+	}
+	if c.CPUDeviceMode == driver.CPU_DEVICE_MODE_GROUPED {
+		if c.GroupBy != driver.GROUP_BY_SOCKET && c.GroupBy != driver.GROUP_BY_NUMA_NODE && c.GroupBy != driver.GROUP_BY_MACHINE {
+			return fmt.Errorf("invalid groupBy %q: must be %q, %q, or %q",
+				c.GroupBy, driver.GROUP_BY_SOCKET, driver.GROUP_BY_NUMA_NODE, driver.GROUP_BY_MACHINE)
+		}
+	}
+	return nil
```
AutuSnow, line 67: Validate groupBy only when cpuDeviceMode==grouping; The unconditional validation of CLI's Set may have been brought in when I was writing code, Sorry

internal/gatherinfo/gatherinfo.go
```diff
@@ -263,14 +263,23 @@ func detectDriverConfig(defaults driverconfig.Config, driverCmdlinePath string)
...
 	}
 
-	return cfg
+	if configFile == "" {
+		return cfg
+	}
+	// If the file can't be read (e.g. run outside the driver container), fall back to defaults.
+	loaded, err := driverconfig.Load(cfg, configFile, fs)
+	if err != nil {
+		return cfg
```
AutuSnow, line 306: nit:There is an issue with the annotations above

deployment/helm/dra-driver-cpu/templates/configmap.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ include "dra-driver-cpu.fullname" . }}-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dra-driver-cpu.labels" . | nindent 4 }}
+data:
+  config.yaml: |
+    {{- .Values.driverConfig | toYaml | nindent 4 }}
+{{- end }}
```
AutuSnow, line 26: One more blocking issue I think we should address before treating `driverConfig` as the main Helm path:

this PR adds a ConfigMap-backed config file, but I don't see a rollout trigger tied to config changes in the DaemonSet template. The driver loads config at startup, so changing only `driverConfig` during `helm upgrade` can leave the running pods on the old effective configuration until they are restarted.

Can we add a checksum annotation (or equivalent rollout trigger) for the generated ConfigMap?

AutuSnow (commented):
Once the precedence issue is fixed, we should also decide what the single source of truth is for `bindAddress` vs `healthzPort`/probes in the chart, otherwise Helm health checks and the actual listener can drift apart.

AutuSnow:
My concern is that, in the current shape, `--config` does not really introduce a cleaner configuration model for Helm users. It adds another configuration layer on top of an already existing Helm/flag path, so the result is overlapping mechanisms rather than a simpler source of truth.
﻿
Because of that, I am not against config files in principle, but I am not convinced this PR shape is the right point to introduce `--config`.
@ffromani @pravk03

ffromani:
thanks for your review @AutuSnow . Your observation has been noted and I'll take into account in my review, which I'm hoping to be able to do ASAP.

internal/driverconfig/load.go
fmuyassarov: Thanks. Good point, the bug was pre existing (the unconditional `--cpu-device-mode` / `--group-by` args) but only became a problem with this PR, which introduced driverConfig and misleadingly suggested those fields could be set there. Fixed.

internal/driverconfig/cfgfile.go
fmuyassarov: Thanks, that's good point. I've now added `DisallowUnknownFields()`, we now return an error instead of radio silence.

internal/gatherinfo/gatherinfo.go
fmuyassarov: fixed

deployment/helm/dra-driver-cpu/templates/configmap.yaml
fmuyassarov: Fair point but I want ask first if we should expect user modifying config after the driver is deployed, in other words post install. I think even currently we don't have hot reload or sort of a signal handler for reloading config. So, all fields require a restart to take effect. If we want to stick to the same pattern then I don't see a value in making in rollout trigger.

fmuyassarov:
> My concern is that, in the current shape, `--config` does not really introduce a cleaner configuration model for Helm users. It adds another configuration layer on top of an already existing Helm/flag path, so the result is overlapping mechanisms rather than a simpler source of truth. ﻿ Because of that, I am not against config files in principle, but I am not convinced this PR shape is the right point to introduce `--config`. @ffromani @pravk03

You are correct that the Helm `driverConfig` field, which backs the driver's `--config` option, is currently very similar to the existing per flag Helm args. But note that this is intentional and is exactly what I mentioned in the description. During the transition period, while we deprecate the standalone flags(follow up not here), we need to support both options (_I think_) to avoid hurting existing users and to give them enough time to migrate to the new `driverConfig`.

Is it simpler at this stage? Probably not. However the expectation is that it will become much simpler over time, as users will no longer need to pass a large number of `--set` flags during installation and can instead provide the driver configuration in a more structured way and assuming the number of options will grow.

fmuyassarov:
> /test pull-dra-driver-cpu-modernize

unfortunately, we need https://github.com/kubernetes/test-infra/pull/37359 before we let it run

internal/driverconfig/load.go
AutuSnow: You were right about the precedence problem. The `cpuDeviceMode` / `groupBy` part is fixed now because those flags are no longer rendered unconditionally, so `driverConfig` can drive them when `args.*` is left empty.
﻿
However, `bindAddress` is still a real gap on the Helm path: the chart still renders `--bind-address=:{{ .Values.healthzPort }}` by default, which means `driverConfig.bindAddress` is still overridden by an explicit CLI flag. So I think this thread should stay open until we decide what the single source of truth is between `bindAddress` and `healthzPort` / probes in the chart.

internal/driverconfig/cfgfile.go
AutuSnow: This part is fixed now.
﻿
`driverConfig` no longer silently ignores unknown fields at runtime: `applyMap` now uses `DisallowUnknownFields()`, so misspelled or unsupported keys fail startup instead of being accepted and falling back to defaults.
﻿
One thing that is still true is that the Helm schema for `driverConfig` remains broad, so Helm itself does not validate those fields up front yet. But the silent runtime acceptance issue should be addressed.

deployment/helm/dra-driver-cpu/templates/configmap.yaml
AutuSnow: I think this is still worth addressing.
﻿
The issue is not hot reload by itself; it is that the new ConfigMap-backed path behaves worse than the existing `args.*` path during `helm upgrade`. Changing `args.*` changes the Pod template and naturally rolls the DaemonSet, while changing only `driverConfig` updates the ConfigMap without forcing a restart. That can leave the running Pods on the old effective configuration until they are restarted manually.
﻿
Because of that behavior difference, I think adding a checksum annotation (or equivalent rollout trigger) is justified here.

AutuSnow:
> > My concern is that, in the current shape, `--config` does not really introduce a cleaner configuration model for Helm users. It adds another configuration layer on top of an already existing Helm/flag path, so the result is overlapping mechanisms rather than a simpler source of truth. ﻿ Because of that, I am not against config files in principle, but I am not convinced this PR shape is the right point to introduce `--config`. @ffromani @pravk03
> 
> You are correct that the Helm `driverConfig` field, which backs the driver's `--config` option, is currently very similar to the existing per flag Helm args. But note that this is intentional and is exactly what I mentioned in the description. During the transition period, while we deprecate the standalone flags(follow up not here), we need to support both options (_I think_) to avoid hurting existing users and to give them enough time to migrate to the new `driverConfig`.
> 
> Is it simpler at this stage? Probably not. However the expectation is that it will become much simpler over time, as users will no longer need to pass a large number of `--set` flags during installation and can instead provide the driver configuration in a more structured way and assuming the number of options will grow.

So I don't think the direction is wrong, but I do agree the current PR still needs some tightening before the “main Helm path” positioning is convincing.

deployment/helm/dra-driver-cpu/templates/configmap.yaml
fmuyassarov: I think I now fully grasp your point. If nothing in the Pod spec changes during a helm upgrade, existing pods won't be restarted even if the configMap content has changed and adding the checksum annotation to force a rollout when driverConfig changes. Totally makes sense. Adding it....

internal/driverconfig/cfgfile.go
fmuyassarov: > so Helm itself does not validate those fields up front yet. But the silent runtime acceptance issue should be addressed.

We could use the Helm scheme validation to catch wrong values during install and that was my intention in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/213/changes/760147997f3944afae6d5d35344f18494de66b70 but it seems it was broken and not fully finalized really.  So, if you check the updated commit, I've now added driverconfig.schema.json, basically a file that lists all valid field names for `driverConfig` (this matches Go config struct), with their types and enum constraints. It is referenced from values.yaml via `$ref`. Also, added `--bundle` flag to the `helm-schema` Makefile target so that the generator inlines the referenced schema into values.schema.json under $defs, and we end up with a single self contained file that Helm can validate against without resolving external references.

deployment/helm/dra-driver-cpu/templates/configmap.yaml
fmuyassarov: fixed

internal/driverconfig/load.go
fmuyassarov: What if, we remove `bindAddress` from schema validation so that Helm rejects it if a user tries to set it. In other words, if a user tries 
```YAML
driverConfig:
  bindAddress: :9090 
```
Helm will reject it during install time with a validation error.  Basically, we make the constraint explicit that users must use `healthzPort` to change the bind address.

AutuSnow: I think this is still unresolved in the current head.
﻿
`driverConfig.bindAddress` is still advertised as a valid Helm-facing field, but the chart always renders `--bind-address=:{{ .Values.healthzPort }}` here, so the config-file value can never win on the Helm path.
﻿
I verified this by rendering the chart with:
﻿
`--set driverConfig.bindAddress=:9090`
﻿
and the manifest still contains:
﻿
`--bind-address=:8080`
﻿
together with `containerPort: 8080`.
﻿
So at the moment Helm accepts `driverConfig.bindAddress`, but the deployed driver silently ignores it. I think we should either:
﻿
1. remove `bindAddress` from the Helm-exposed `driverConfig` schema/docs and make `healthzPort` the only supported knob, or
2. rework the chart so listener/probes derive from a single source of truth instead of unconditionally overriding the config file.

fmuyassarov: yes it is not resolved, i was rather asking if what I think makes sense or good way to proceed. If yes, then I would make the changes.

AutuSnow: Yes, I think that is a reasonable way to proceed for this PR.
﻿
If Helm is going to keep `healthzPort` as the source of truth for the listener/probe settings, then `driverConfig.bindAddress` should not be exposed as a valid Helm-facing field. Otherwise we keep a silent precedence trap.

fmuyassarov: done

cmd/dracpu/app.go
```diff
@@ -52,10 +52,18 @@ const (
...
+	ready atomic.Bool
 )
 
 func init() {
+	// --config is kept outside Config to avoid a self-referential loop,
+	// following the convention used by kubeadm and similar tools.
+	flag.StringVar(&configFile, "config", "",
+		"Path to a YAML driver configuration file. Values in the file "+
+			"are applied for any flag not explicitly provided on the command line. "+
+			"If empty, only CLI flags and built-in defaults are used.")
```
ffromani: I would just say that configuration values are set this way: defaults -> file -> flags and only values set explicitly overrides the previous values.
IIRC we don't use env vars; if we introduce them, we will need to decide if they sit before or after the flags. But YAGNI.

cmd/dracpu/app.go
```diff
@@ -84,26 +92,32 @@ func main() {
...
 
-func run(logger logr.Logger) error {
+func run(logger logr.Logger, cfg driverconfig.Config) error {
 	printVersion(logger)
-	flag.VisitAll(func(f *flag.Flag) {
-		logger.Info("FLAG", "name", f.Name, "value", f.Value.String())
-	})
+	logger.Info("CONFIG", "bindAddress", cfg.BindAddress, "cpuDeviceMode", cfg.CPUDeviceMode,
+		"groupBy", cfg.GroupBy, "reservedCPUs", cfg.ReservedCPUs,
+		"hostnameOverride", cfg.HostnameOverride, "exposePCIeRoots", cfg.ExposePCIeRoots)
```
ffromani: This can be fine for the first iteration, but the right shape is IMO something like `flag.VisitAll` (note this NOT implies we would need introspection to implement it!) because otherwise we add some coupling between the logger in the call site and the shape of the Config object. Using a visitor approach removes that coupling. IOW, moving the responsability to find and iterate somehow over the config fields should be moved in a method of Config object/package.
Granted, the shape of the struct is this part of the config file API, but still we move in the right direction and we try to push back coupling rather than embracing.
TL;DR: a good improvement could be to just copypaste this line and wrap into a helper exposed by the COnfig struct or driverconfig package.

deployment/helm/dra-driver-cpu/README.md
```diff
@@ -35,12 +35,13 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
...
 | args.exposePCIeRoots | bool | `false` | Discover and expose PCIe roots as device attributes. Requires the `DRAListTypeAttributes=true` feature gate in the cluster |
-| args.groupBy | string | `"numanode"` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode`, `socket` or `machine` |
+| args.groupBy | string | `""` | Grouping criteria when `cpuDeviceMode=grouped`: `numanode`, `socket` or `machine`; when empty, defers to `driverConfig.groupBy` or the built-in default (`numanode`) |
 | args.hostnameOverride | string | `""` | Override the node name the driver registers under; omitted when empty |
 | args.logLevel | int | `4` | Log verbosity level passed as `--v` |
 | args.reservedCPUs | string | `""` | CPUs reserved for the OS and kubelet, excluded from DRA management (e.g. `"0-1"`); omitted when empty |
+| driverConfig | object | `{}` | Driver config file contents. When non-empty, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. Fields in `args.*` that are non-empty always take priority over values set here. Example:   driverConfig:     cpuDeviceMode: individual     groupBy: socket     reservedCPUs: "0-3" |
```
ffromani, line 44: regressing on the defaults is something worth noting though. I'm not saying it's a blocker, but maybe something we can improve later. Curious to learn why we have this little regression at all.

README.md
```diff
@@ -32,6 +32,29 @@ The driver can be configured with the following command-line flags:
...
 - `--expose-pcie-roots`: If enabled, adds the "resource.kubernetes.io/pcieRoot" standard value to CPU devices, to report the PCIe roots close to each device. Since it always reports values as list, this option requires the cluster Feature Gate `DRAListTypeAttributes` (see KEP 5491) to be enabled. The driver has no way to introspect the cluster Feature Gate, so care must be taken to enable first the Feature Gate then this option.
 
+### Configuration file
```
ffromani, line 37: thanks, this is fine and helps. We should add some lines describing the schema of the config file and possibly providing an example. 

Things we should document: is the layout flat? do we have groups? if we decide to move to a nested hierararchy, how will we handle backward compat (easy answer: add versioning somehow)

internal/driverconfig/load.go
```diff
@@ -0,0 +1,71 @@
...
+	"reserved-cpus":     "reservedCPUs",
+	"cpu-device-mode":   "cpuDeviceMode",
+	"group-by":          "groupBy",
+	"expose-pcie-roots": "exposePCIeRoots",
+}
+
+// Load merges the config file at filePath into base, giving CLI flags that were
+// explicitly set (reported by fs.Visit) priority over file values.
+// If filePath is empty, base is returned unchanged. fs must already be parsed.
+func Load(base Config, filePath string, fs *flag.FlagSet) (Config, error) {
```
ffromani: the general shape makes sense. There could be some implementation streamlining possible, but these can be deferred.
The general behavior should be that the new config layers should be purely additive and incremental; IOW, flags whose value we don't set from commandline flags should be inherit their value from the config file, and if not specified in the config file they should inherit the default value.

internal/driverconfig/load.go
```diff
@@ -0,0 +1,71 @@
...
+
+	explicitJSONKeys := map[string]bool{}
+	fs.Visit(func(f *flag.Flag) {
+		if jsonKey, ok := flagToJSONKey[f.Name]; ok {
+			explicitJSONKeys[jsonKey] = true
+		}
+	})
+
+	confMap := map[string]any{}
+	if err := buildConfMap(confMap, filePath); err != nil {
```
ffromani: Idea/nice to have: it could be useful to dump the config after each stage, at very high (V>=7) verbose level. IOW: defaults (full dump); after applying config values (full dump); after applying flags/final shape (full dump)

internal/driverconfig/cfgfile.go
```diff
@@ -0,0 +1,91 @@
...
+// validateAPIVersion checks confMap["apiVersion"] when present.
+func validateAPIVersion(confMap map[string]any) error {
+	raw, ok := confMap["apiVersion"]
+	if !ok {
+		return nil
+	}
+	apiVer, _ := raw.(string)
+	if apiVer != ConfigAPIVersion {
+		return fmt.Errorf("unsupported apiVersion %q, want %q", apiVer, ConfigAPIVersion)
+	}
```
ffromani, line 68: ok so config is already versioned (elsewhere I commented we would need versioning). Great. It should be documented somewhere in the user-facing docs

ffromani (commented):
I'm sharing thoughts about the general design and overall architecture; I didn't yet read prior comments and I'm consciously trying to defer implementation details unless there's something really catching my eye or interacting with the general design.

Retrofitting a config file support on an app which started with commandline flags only may turn dirty work and surprisingly hard (in relative terms ofc), this is one of the reasons the earlier we do the better. Nevertheless, it may require some iterations and may live some rough edges. I'm completely fine with that, and for an app whose version is 0.2.0 evolving to 0.3.0, I think it's fair.

fmuyassarov:
Thanks @ffromani for the reviews, appreciate it. I will go through each comment and update soon.

deployment/helm/dra-driver-cpu/README.md
fmuyassarov: It's a necessary trade off to let `driverConfig` to take effect. I would not do it otherwise. 

What happens is that, if you set `args.cpuDeviceMode: grouped` explicitly in values.yaml, the chart renders `--cpu-device-mode=grouped` as a CLI flag, which means `driverConfig.cpuDeviceMode` can never win. So, the defaults have to be  `""`  in  values.yaml to let `driverConfig` to the work.

The helm template has  `{{- if .Values.args.cpuDeviceMode }}`. When the value is empty string, Go templates treat it as falsy and the flag is not rendered. `fs.Visit` only sees flags that are passed on the command line. If the flag wasn't rendered, it's not in `explicitJSONKeys`, so it is not deleted from confMap and the config file value gets applied.

cmd/dracpu/app.go
fmuyassarov: updated

internal/driverconfig/load.go
fmuyassarov: Makes sense. Though, I used the same highest level we already have (`v(6)`). Let me know if you want me to go one layer higher.

internal/driverconfig/cfgfile.go
fmuyassarov: yup, updated the docs as well for this.

fmuyassarov:
> Thanks @ffromani for the reviews, appreciate it. I will go through each comment and update soon.

I think I have now addressed the comments. Please feel free to take your time and go through another round of reviews if you think it's necessary. We aren't in rush.

deployment/helm/dra-driver-cpu/README.md
ffromani: thanks for elaborating. I was suspecting this is a necessary evil. I'd be fine starting this way and perhaps iterating over time to try to remove this limitation, if possible at all.

internal/driverconfig/load.go
ffromani: V=6 is fine for starters.

ffromani:
> > Thanks @ffromani for the reviews, appreciate it. I will go through each comment and update soon.
> 
> I think I have now addressed the comments. Please feel free to take your time and go through another round of reviews if you think it's necessary. We aren't in rush.

thanks for your updates. I concur we need more review but because of the nature of retrofitting config files on an existing flag baseline, with all the quirks and the rough edges this exercise requires. This is a general problem, I think this PR is heading the right direction and it's minimizing the necessary pain we have to endure.

deployment/helm/dra-driver-cpu/README.md
fmuyassarov: yup, I believe&expect there will be several rounds of iterations/follow-ups to the change as a whole.

fmuyassarov:
> > > Thanks @ffromani for the reviews, appreciate it. I will go through each comment and update soon.
> > 
> > 
> > I think I have now addressed the comments. Please feel free to take your time and go through another round of reviews if you think it's necessary. We aren't in rush.
> 
> thanks for your updates. I concur we need more review but because of the nature of retrofitting config files on an existing flag baseline, with all the quirks and the rough edges this exercise requires. This is a general problem, I think this PR is heading the right direction and it's minimizing the necessary pain we have to endure.

and as I mentioned in a comment above, I'm almost certain that there will be things that will turn out to be not fully correct or require improvements. But that's fine at this stage and I will be happy to address and keep improving it even once changes land.

internal/driverconfig/load.go
fmuyassarov: Good point and this is now handled.  `Load()` layers defaults then file then CLI flags (via  `fs.Visit `). So unset fields fall through correctly.

fmuyassarov:
Hi @ffromani @AutuSnow is there some blocker that you would like me to address or are we good to move with the current version.

AutuSnow:
Thanks for the ping @fmuyassarov. From my side there's no blocker — all three threads I raised are addressed in the current head, and I re-verified each.
Good to move forward as far as I'm concerned. One small non-blocking note for later: driverconfig.schema.json is hand-maintained to mirror the Config struct — the comment reminder helps, but it'd be worth a follow-up to keep it in sync (a test or generator) so the two don't drift. Not something that needs to hold this PR.
﻿

pravk03:
Sorry for the delay in reviewing this (and the other PR's). Was OOO for a bit and also trying to make progress on the KEP. I plan to review this in the next couple of days and add my thoughts.

fmuyassarov:
> Sorry for the delay in reviewing this (and the other PR's). Was OOO for a bit and also trying to make progress on the KEP. I plan to review this in the next couple of days and add my thoughts.


No worries. As said, I don't have pressing deadlines. Just want to make sure that we all are on the same page. So, take tour time @pravk03 and thanks in advance.

fmuyassarov:
rebased

internal/driverconfig/load.go
```diff
@@ -0,0 +1,71 @@
...
+func Load(base Config, filePath string, fs *flag.FlagSet) (Config, error) {
+	if filePath == "" {
+		return base, nil
+	}
+
+	explicitJSONKeys := map[string]bool{}
+	fs.Visit(func(f *flag.Flag) {
+		if jsonKey, ok := flagToJSONKey[f.Name]; ok {
+			explicitJSONKeys[jsonKey] = true
+		}
```
pravk03, line 56: Should we add an error log if CLI flag is missing in `flagToJSONKey` ?

pravk03: Or, a unit test to make sure all CLI flags are present in `flagToJSONKey`. My concern is that since `flagToJSONKey` should be manually updated, its easy to miss. Although after this PR, ideally we should sop adding new flags.

internal/driverconfig/load.go
```diff
@@ -0,0 +1,71 @@
...
+package driverconfig
+
+import (
+	"flag"
+	"fmt"
+)
+
+// flagToJSONKey maps CLI flag names to their Config JSON keys.
+// Add an entry here whenever a new field is added to Config and AddFlags.
+var flagToJSONKey = map[string]string{
```
pravk03, line 28: `ShowMetrics`, `SysFSOverlay` missing ?

Comparing with https://github.com/kubernetes-sigs/dra-driver-cpu/blob/10df102c22d32623ee92851b060220cf81d8d1f4/internal/driverconfig/config.go#L26

internal/driverconfig/cfgfile.go
```diff
@@ -0,0 +1,88 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/yaml"
+)
+
+// buildConfMap loads the config file at filePath into confMap.
+// It validates and strips "apiVersion" before returning.
+func buildConfMap(confMap map[string]interface{}, filePath string) error {
```
pravk03: nit: 

`func buildConfMap(filePath string) (map[string]any, error)`
 
this maybe cleaner ?. The caller need not initialize empty map.

internal/driverconfig/cfgfile.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := validateAPIVersion(confMap); err != nil {
+		return err
+	}
+	delete(confMap, "apiVersion")
+
+	return nil
+}
+
+// loadAndMerge reads the YAML file at path and merges it into dst.
+func loadAndMerge(dst map[string]any, path string) error {
```
pravk03: nit: 

`func loadConfigFile(path string) (map[string]any, error)` ?

internal/driverconfig/cfgfile.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return fmt.Errorf("reading %s: %w", path, err)
+	}
+
+	src := map[string]interface{}{}
+	if err := yaml.Unmarshal(data, &src); err != nil {
+		return fmt.Errorf("parsing %s: %w", path, err)
+	}
+
+	mergeMap(dst, src)
```
pravk03: Can we not return `src` directly here ?.

cmd/dracpu/app.go
```diff
@@ -86,26 +95,30 @@ func main() {
 
 	logger := ctxlog.Setup()
 
-	if err := runDriver(logger); err != nil {
+	cfg, err := driverconfig.Load(driverFlags, configFile, flag.CommandLine, logger)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "dracpu: failed to load configuration: %v\n", err)
```
pravk03: nit: logger.Error ?

internal/gatherinfo/gatherinfo.go
```diff
@@ -287,14 +287,23 @@ func detectDriverConfig(defaults driverconfig.Config, driverCmdlinePath string)
...
 	if err := fs.Parse(knownConfigArgs(fs, cmdline[1:])); err != nil {
 		return defaults
 	}
 
-	return cfg
+	if configFile == "" {
+		return cfg
+	}
+	// If the file can't be read (e.g. run outside the driver container), fall back to CLI-parsed config.
+	loaded, err := driverconfig.Load(cfg, configFile, fs, logr.Discard())
```
pravk03, line 304: I think this will work when the gatherinfo tool is run from within the ds pod, but not when used as a standalone tool 
cc @AutuSnow

pravk03: But the fix seems easy, we can follow the same approach as with sysfs 

https://github.com/kubernetes-sigs/dra-driver-cpu/blob/10df102c22d32623ee92851b060220cf81d8d1f4/internal/gatherinfo/gatherinfo.go#L196

README.md
pravk03, line 21: nit: we can split this into 2 sub headings - Command-line flags, Configuration file

deployment/helm/dra-driver-cpu/driverconfig.schema.json
```diff
@@ -0,0 +1,33 @@
...
+    },
+    "cpuDeviceMode": {
+      "type": "string",
+      "enum": ["grouped", "individual"]
+    },
+    "groupBy": {
+      "type": "string",
+      "enum": ["numanode", "socket", "machine"]
+    },
+    "exposePCIeRoots": {
```
pravk03: `sysfsOverlay` and `showMetrics` missing.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
pravk03: CLI overriding config file settings would break in the case of boolean flags ?

Eg: CLI `exposePCIeRoots` arg is false -> so we do not add it in ds spec -> `exposePCIeRoots` is true in config file -> driver uses the config file value.

README.md
```diff
@@ -32,6 +32,56 @@ The driver can be configured with the following command-line flags:
...
+| `groupBy`          | string | `numanode`  | Grouping strategy when `cpuDeviceMode` is `grouped`: `numanode`, `socket`, or `machine`. |
+| `reservedCPUs`     | string | *(none)*    | CPUs excluded from allocation, e.g. `"0-1"`.                                             |
+| `hostnameOverride` | string | *(none)*    | Override the node hostname the driver registers under.                                   |
+| `exposePCIeRoots`  | bool   | `false`     | Add PCIe root attributes to CPU devices (requires `DRAListTypeAttributes` feature gate). |
+| `kubeconfig`       | string | *(none)*    | Path to a kubeconfig file (for out-of-cluster use).                                      |
+
+#### Versioning and backward compatibility
+
+The schema is versioned via the optional `apiVersion` field (currently `v1alpha1`). The layout
+is intentionally flat for now. If a nested hierarchy is introduced in the future, the
```
pravk03, line 61: +1 to making this nested in the future. As we add more features, we could use some logical grouping.

pravk03 (commented):
This is great! Thanks a lot for adding this. I think it will really help with future features by keeping the configuration more manageable. I like the direction and agree with the discussions and conclusions made so far in the discussions above. All of my comments are implementation-related, and many of them are nits. Feel free to postpone any of my comments to a follow-up PR.

internal/gatherinfo/gatherinfo.go
AutuSnow: Good catch, thanks @pravk03 — you're right. configFile here is the path as it appears in the driver's cmdline, i.e. a path inside the driver container. Inside the ds pod it resolves; run standalone it won't, and right now we silently fall back to the CLI-parsed cfg, so the report would quietly drop whatever the config file contributed.
I think we already have the pattern for this a few lines up: collectReport resolves the SysFSOverlay path through driverFilesystemPath(driverCmdlinePath, overlayPath) (→ /proc/<pid>/root/... for absolute, /proc/<pid>/cwd/... for relative). Since detectDriverConfig already receives driverCmdlinePath, we can route configFile through the same helper before driverconfig.Load, and then the standalone case reads the driver's real config file too — with the same access assumptions the overlay handling already relies on, so no new requirement.
One thing I'd add regardless of the resolution approach: the read failure shouldn't be silent. If we still can't load it (e.g. no access to /proc/<pid>/root), let's surface that in the report / log rather than returning a config that differs from the driver's actual effective one — same "no silent trap" line as the earlier threads.

ffromani:
> One thing I forgot to mention is - regarding the drop in configs. It's the same battle I had with myself as I raised in the description. I'm still leaning towards not adding them, but if you see value in having them for the driver at this stage, I'm happy to include them. I still have the commit ready.

context: drop-in configlets are the YAML snippets the driver loads and merges with the main config file, merging them into a final document before to process it (note made-up paths):
```
/etc/dracpu/config.yaml
/etc/dracpu/config.yaml.d/00-foo.yaml
/etc/dracpu/config.yaml.d/10-bar.yaml
/etc/dracpu/config.yaml.d/99-buz.yaml
```

these are useful for system integrators: easier to drop a configlet than read/process/write the whole config in one go.
As long as the code is reasonnably ready and simple to be extended with configlets/drop-ins at a later stage, I'm fine to defer them

cmd/dracpu/app.go
ffromani: yes please, we should have a logger ready when we are here. if not, maybe because we have or we will introduce circular deps between logger and config, I'd still prefer us to setup a bare-bones bootstrap logger (`stdr`?) to handle early processing errors, and then replace with the real logger as soon as possible

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
ffromani: great point. Can we perhaps workarounding this issue by setting all the flags all the time with explicit value?
e.g.: `exposePCIeRoots=false` -> `--expose-pcie-roots=false`
IOW:
```
  # instead of
  {{- if .Values.args.exposePCIeRoots }}
  - --expose-pcie-roots={{ .Values.args.exposePCIeRoots }}
  {{- end }}

  # always render
  - --expose-pcie-roots={{ .Values.args.exposePCIeRoots }}
```
now, this **IS** a little ugly, but we're trending in the direction of removing the flags anyway; and, most notably, we have still a manageable number of flags.

deployment/helm/dra-driver-cpu/driverconfig.schema.json
ffromani: I wonder if there's a way to generate the schema from the golang structs which define the `driverconfig` (or the other way around). IOW, if there's a way to have a single source of truth and automatically derive the rest.

If we go this way, it's totally fine to defer to a followup

internal/driverconfig/cfgfile.go
ffromani: +1

internal/driverconfig/config.go
```diff
@@ -35,85 +32,14 @@ type Config struct {
...
-		return ""
-	}
-	return *v.value
-}
-
-func (v *groupByValue) Set(s string) error {
-	if s != device.GROUP_BY_SOCKET && s != device.GROUP_BY_NUMA_NODE && s != device.GROUP_BY_MACHINE {
-		return fmt.Errorf("invalid value: %q, must be %s, %s or %s", s, device.GROUP_BY_SOCKET, device.GROUP_BY_NUMA_NODE, device.GROUP_BY_MACHINE)
+// LogValues returns key-value pairs for structured logging of the config.
+func (c Config) LogValues() []any {
```
ffromani, line 36: good idea

internal/driverconfig/config_test.go
```diff
@@ -0,0 +1,221 @@
...
+	require.NoError(t, os.WriteFile(path, []byte(content), 0600))
+	return path
+}
+
+// TestLoad_NoFile: empty filePath returns base unchanged.
+func TestLoad_NoFile(t *testing.T) {
+	cfg := driverconfig.Default()
+	fs := newFlagSet(t, &cfg, nil)
+
+	result, err := driverconfig.Load(cfg, "", fs, logr.Discard())
```
ffromani: if you resubmit, I realized `testr.New(t)` is likely better than `logr.Discard()`. Can totally be fixed as followup

internal/driverconfig/config_test.go
```diff
@@ -0,0 +1,221 @@
...
+	return fs
+}
+
+// writeFile creates name with content inside dir and returns the full path.
+func writeFile(t *testing.T, dir, name, content string) string {
+	t.Helper()
+	path := filepath.Join(dir, name)
+	require.NoError(t, os.WriteFile(path, []byte(content), 0600))
+	return path
+}
```
ffromani, line 51: scratch this, it's mostly a personal preference, we don't have a standard/convention there

~if you resubmit, and unless other reviewers prefer differently, move the helpers at the end of the file, so the noise is limited and the most important stuff, the tests proper, appear first~

internal/driverconfig/load.go
ffromani: +1 for unit test

internal/driverconfig/load.go
```diff
@@ -0,0 +1,77 @@
...
+	for jsonKey := range explicitJSONKeys {
+		delete(confMap, jsonKey)
+	}
+
+	result := base
+	if err := applyMap(&result, confMap); err != nil {
+		return Config{}, fmt.Errorf("applying config file %q: %w", filePath, err)
+	}
+
+	logger.V(6).Info("config: after file", result.LogValues()...)
```
ffromani, line 75: where do we log after flags?

ffromani (commented):
thanks for pushing this forward. I'm fine with the general direction, the current PR shape and the roadmap/followups. I don't see any major blocker but some details to iron out before we can merge.

Added some questions and suggestions inline

EDIT: forgot to mention, test coverage looks nice

internal/driverconfig/load.go
fmuyassarov: Definitely, otherwise we end up with something like you found out here: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/213#discussion_r3555453376

internal/driverconfig/cfgfile.go
fmuyassarov: +1, I've now refactored `buildConfMap/loadAndMerge` to drop dst, src mergeMap pattern entirely.

deployment/helm/dra-driver-cpu/driverconfig.schema.json
fmuyassarov: Good catch @pravk03 and thanks for the suggestion @ffromani. Totally agree and would love to have it too to avoid another accidental drift from Config struct. I think there are several options (with some libs) for deriving the scheme from Go structs. I will create a tracking issue and make a follow up changes on that front once we land the current changes here.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
fmuyassarov: Thanks @pravk03 @ffromani for the catch and the suggestion. Refactoring it now to unconditionally emit `expose-pcie-roots`

internal/driverconfig/load.go
fmuyassarov: Very valid catch, that log line was actually printing `Default(),` not base, so it always showed the same hardcoded values no matter what flags were passed. There was no log point after `fs.Parse` at all. Now fixed it to log base instead right after flags are parsed.

internal/driverconfig/config_test.go
fmuyassarov: +1, thanks

internal/driverconfig/load.go
fmuyassarov: thank you and sorry for the miss.

internal/gatherinfo/gatherinfo.go
fmuyassarov: thanks, done

cmd/dracpu/app.go
fmuyassarov: yup, updated.

internal/driverconfig/config.go
fmuyassarov: just realized after this that we are missing `kubeconfig`, `showMetrics` and `sysfsOverlay` here and ended up adding T`estConfigLogValues_CoversAllFields` to protect against `LogValues()` silently missing fields. Not very important but sort of a good to have test I thought.

fmuyassarov:
Thanks for the latest reviews @pravk03 @ffromani, those were good points and missed bugs. I've addressed the them and hopefully didn't miss any.

ffromani:
note about package `gatherinfo_test` - it's our only package I can think of whose tests sit on the _test package rather in the same package. I noticed but didn't register this now and connected the dots because you added the `_internal_test.go`.

This idiom - having the test in their separate package - is welcome but not preferred nor enforced. We can very much test the packages by their external API putting the tests in the same package. Not warrant any further change for now, let's see how the codebase evolves.

ffromani:
/lgtm

all my comments have been addressed. Since both @pravk03 and @AutuSnow did extensive reviews recently, and since this change affects all the project. I'd like to give them a chance to chime in.

fmuyassarov:
> note about package `gatherinfo_test` - it's our only package I can think of whose tests sit on the _test package rather in the same package. I noticed but didn't register this now and connected the dots because you added the `_internal_test.go`.
> 
> This idiom - having the test in their separate package - is welcome but not preferred nor enforced. We can very much test the packages by their external API putting the tests in the same package. Not warrant any further change for now, let's see how the codebase evolves.




Good observation @ffromani. No strong opinion from me to be frank. But let's see how we evolve. I will keep this point in mind.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
AutuSnow: Following up on the unconditional-emit change here — it does fix the boolean direction @pravk03 raised (an explicit `args.exposePCIeRoots=false` can now override a config-file `true`).
﻿
I just want to sanity-check the *other* direction on the Helm path. Because `--expose-pcie-roots={{ .Values.args.exposePCIeRoots }}` is now always rendered,`args.exposePCIeRoots` (default `false`) is always an explicit CLI flag, so `driverConfig.exposePCIeRoots` can never win when a user sets it via the config map and leaves `args` at the default. Rendered to confirm:
﻿
`# helm template ... --set driverConfig.exposePCIeRoots=true
data:
config.yaml: |
exposePCIeRoots: true      # config map says true
...
- --expose-pcie-roots=false    # but the flag is false → driver runs with false
- --config=/etc/dracpu/config.yaml`
﻿
This is the same silent-precedence shape we decided to avoid for `bindAddress`(removed from the `driverConfig` schema so `healthzPort` is the single knob).For `exposePCIeRoots` we've effectively made `args.*` the single source of truth on the Helm path — consistent with "args.* always wins" — but the field is still
advertised in the `driverConfig` schema, so a user can set it and have it silently ignored.
﻿
Would it make sense to reconcile the two the same way — either drop the bool fields from the Helm `driverConfig` schema (like `bindAddress`), or add a short note that bool settings must go through `args.*` on the Helm path? Happy either way; just flagging so we don't leave a trap that only bites bool fields. Does
that match your intent?

cmd/dracpu/app.go
AutuSnow: Small thing I noticed while tracing the config plumbing: `run()` reads every setting from the merged `cfg` (`cfg.ReservedCPUs`, `cfg.BindAddress`, `cfg.CPUDeviceMode`, `cfg.ExposePCIeRoots`, …) except this one line, which still reads `driverFlags.SysFSOverlay`:
﻿
sfs, err := newSysFS(logger, driverFlags.SysFSOverlay)
﻿
Since `sysfsOverlay` is now a first-class config field (it's in `flagToJSONKey`, the Helm `driverconfig.schema.json`, and `LogValues()`), a user who sets it only via the config file will have it loaded into `cfg.SysFSOverlay` and logged as active by `LogValues()`, but `newSysFS` would still use the empty flag value — so the overlay wouldn't take effect and the log line would disagree with the real
behavior.
﻿
Is reading `driverFlags` here intentional, or should it be `cfg.SysFSOverlay` to match the rest of `run()`? (The `driverFlags.ShowMetrics` check in `main()` has the same shape, but that one runs before config load, so it may be intended to stay flag-only.)

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
fmuyassarov: very good point @AutuSnow, you are right. Made `args.exposePCIeRoots` the single source of truth on the Helm and dropped `exposePCIeRoots` from the `driverConfig` schema (this is the same as bindAddress/healthzPort ).

cmd/dracpu/app.go
fmuyassarov: do'h that's a bug, `driverFlags.SysFSOverlay` was a leftover, should've been `cfg.SysFSOverlay` like everything else in run(). Now fixed and also added a regression test.

fmuyassarov:
@AutuSnow addressed your latest comment. Thanks for those. PTAL.

AutuSnow:
/lgtm
Thanks！！

deployment/helm/dra-driver-cpu/driverconfig.schema.json
fmuyassarov: filed the tracking issue in https://github.com/kubernetes-sigs/dra-driver-cpu/issues/233 for the improvement and hopefully will file up the required changes once we get this feature in.

ffromani:
/lgtm

thanks for the detailed review and for your attention to detail @AutuSnow

fmuyassarov:
Thanks @pravk03 @ffromani @AutuSnow for the great reviews. Let's wait for @pravk03 to get his final review.

deployment/helm/dra-driver-cpu/driverconfig.schema.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "cpuDeviceMode": {
+      "type": "string",
+      "enum": ["grouped", "individual"]
+    },
+    "groupBy": {
+      "type": "string",
+      "enum": ["numanode", "socket", "machine"]
+    },
+    "showMetrics": {
```
AutuSnow: Small follow-up now that `exposePCIeRoots` has been dropped from the `driverConfig` schema — should `showMetrics` get the same treatment?
﻿
My reasoning: the `--show-metrics` handler runs in `main()` *before* `driverconfig.Load()` and reads `driverFlags.ShowMetrics` (the flag), not the merged `cfg`. So a value set only via `driverConfig.showMetrics` is parsed and loaded into `cfg`, but the print-and-exit path never looks at it — it's silently inert on the config-file path. On top of that, `show-metrics` is really a "print metadata and exit" debugging flag, so it doesn't seem like something that belongs in a persistent deployment config in the first place.
﻿
I know `showMetrics`/`sysfsOverlay` were added here intentionally to mirror the `Config` struct (@pravk03's earlier point), so I might be missing a use case — `sysfsOverlay` now genuinely flows through `cfg`, so that one is fine. But for `showMetrics` specifically, would it make sense to exclude it from the Helm `driverConfig` schema the same way we did for `exposePCIeRoots`/`bindAddress`, and keep it as a CLI-only flag? Totally fine to defer to a follow-up if you'd rather not expand the scope of this PR — just flagging the inconsistency.

AutuSnow (commented):
Sorry, I saw this place while reviewing the modified code earlier, but I didn't pay attention to it. However, after thinking about it again, it seems that this is not appropriate. But I won't block the progress of this PR and can fix it in the subsequent PR

deployment/helm/dra-driver-cpu/driverconfig.schema.json
ffromani: I think you raised a good point @AutuSnow . It seems to me the general point is that we have - and we will have - some flags that don't map in the config because they modify too deeply the runtime behavior of the driver.

`--show-metrics` is a great example. It is a way for us to introspect the driver which is implemented as runtime option. But it's actually a different operational mode rather than a tunable of the basic operational mode.

Tuning the driver operational mode (configurable knobs) I think belong in the configuration file; changing the operational mode should be kept as flags, because it's the less friction/less surprising behavior.

I didn't connect the dots fully either previously, and only had a vague feeling sorting out all these details would require iteration. This conversation seems to confirm my istinct.
I think there's consensus forming this PR would be the first important step in the direction of rationalizing the general configuration side, and which we will need to dull a lot of sharp edges.

Looking forward, I think the best way here is to distinguish operational modes more firmly. We should probably add sub-commands to clearly convey which operational mode do we take. IOW,
it won't be
`dracpu --show-metrics`
but rather
`dracpu introspect metrics` (`metrics` being the positional argument of the `introspect` subcommand - can be `debug` or `info`, just to illustrate the point).

There are SOME backward compatibility concerns to address before to do this or any similar change; we're still 0.2.0 and alpha, but still let's try to be careful and don't users more than is necessary.

For the short term, I think we need to acknowledge this gap, commit to a resolution before 0.3.0 which leaves no sharp edge, but move forward and merge this PR to have the first important building block, especially we have universal consensus moving towards config files is a good thing.

Makes sense?

AutuSnow: Yes, makes sense — and I think the operational-mode vs. tunable-knob line is exactly the right abstraction to draw. "The config file holds tunables of the current mode; switching mode stays a flag/subcommand" is a clean, low-surprise rule, and `dracpu introspect metrics` reads much better than `--show-metrics`.
No objection from me to landing this as the first building block — the consensus on config files is clear and I don't want to hold up the foundation.
﻿
One thing worth deciding explicitly, since it falls straight out of your rule:
if the `driverConfig` schema should only ever carry tunable knobs, then `showMetrics` is the single current exception (and it's inert on that path today anyway). Would it be worth doing just the one-line schema removal now — same treatment as `exposePCIeRoots`/`bindAddress`, no Go or flag changes — so the Helm-facing surface is already clean when this lands, and keep the larger `introspect` subcommand rework (with the backward-compat handling) as the 0.3.0 item?
﻿
Either way is fine by me — if you'd rather keep the scope tight and fold it all
into the 0.3.0 change, I'm good with that too. Happy to file a tracking issue for
the operational-mode / subcommand direction so the pre-0.3.0 commitment doesn't
get lost. Thanks for writing this up so clearly.

fmuyassarov: Totally makes sense. I think we discussed somewhere (can't find among all the comments) that I and we all expect that there will be several rounds of iterations until we reach a solid configuration file for users. But as @ffromani pointed out, this is a starting point.

regarding `showMetrics`, I don't mind to refactor the scheme quickly. It is a tiny change.

ffromani: @AutuSnow filing a tracking issue would help, thanks.
@fmuyassarov since it's a tiny change and you are happy to do it, yes, let's do the 1-line removal. It's cheap and a little clearer, so a little less work for the future

fmuyassarov: on it

fmuyassarov: ok, done now.

ffromani:
/lgtm

thanks! I think we're ready again for the final review

pravk03:
Thanks a lot!

/lgtm
/approve

Issue #188, opened 2026-06-22
Track DRA/NRI allocation lifetime assumptions before future sharing or preemption support


AutuSnow:
`StopContainer` currently releases claim CPU allocations before `UnprepareResourceClaims` runs.

The code comments already document this as a workaround and note that it assumes ResourceClaims are not shared across pods/containers. That is fine for current behavior because claim sharing is explicitly unsupported.

This issue is to track the assumption before future work changes the lifecycle model.

Related:
- kubernetes/enhancements#6113
- kubernetes/enhancements#5987
- #103
- #149
- #179

Suggested:
- document the current DRA/CDI/NRI lifetime invariant
- add tests around `StopContainer`, `UnprepareResourceClaims`, and restart `Synchronize`
- revisit this before enabling claim sharing, preemption, or mixed allocation behavior

thc1006:
I'd like to take the docs + tests part of this. I read through the `StopContainer` early-release path in `nri_hooks.go` and the workaround comment, so the invariant is clear: because claims aren't shared, releasing the guaranteed CPUs back to the shared pool at `StopContainer` (and re-broadening the other shared containers there) is safe, whereas `UnprepareResourceClaims` is too late to update those other containers.

My plan would be unit tests in `nri_hooks_test.go` covering the three transitions: `StopContainer` releases the claim's CPUs and re-broadens shared containers; `UnprepareResourceClaims` cleans up the allocation store; and `Synchronize` rebuilds the right shared/guaranteed split on restart. Plus a short doc comment pinning down the lifetime invariant in one place.

Does that match the scope you had in mind here, leaving the "revisit before sharing/preemption" piece as the follow-up the issue flags? Happy to start once you confirm.

PR #215, opened 2026-06-30
test: cover StopContainer claim release; document lifetime invariant

thc1006:
#### What type of PR is this?

/kind documentation

`StopContainer` is where a guaranteed container's claim CPUs are released back into the shared pool and the surviving shared containers are re-broadened. The existing `TestStopContainer` only asserted which containers received an update, not that the release and re-broadening actually happen. This adds a unit test for that release invariant, plus a doc comment on `StopContainer` that writes down the CPU-allocation lifetime model across the DRA and NRI hooks (Prepare -> Create -> Stop -> Unprepare -> Synchronize), and trims the now-redundant inline comment.

Part of #188

The rest of the lifetime is already covered, so I did not duplicate it: `UnprepareResourceClaims` (`TestUnprepareResourceClaims*`), `Synchronize` (`TestNRISynchronize`), and the store-level idempotency of `RemoveResourceClaimAllocation` (`cpu_allocation_test.go` "remove non-existent"). That idempotency is what keeps the StopContainer early-release consistent with the authoritative Unprepare cleanup, which the new doc comment makes explicit.

cc @AutuSnow

#### Release note

```release-note
NONE
```

### DIFF
--- a/pkg/driver/nri_hooks_test.go
+++ b/pkg/driver/nri_hooks_test.go
@@ -460,3 +460,61 @@ func containerIDsFromUpdates(updates []*api.ContainerUpdate) []string {
 	sort.Strings(ids)
 	return ids
 }
+
+// TestStopContainerReleasesClaimToSharedPool asserts the DRA/NRI lifetime invariant tracked by #188:
+// stopping a guaranteed container releases its claim's CPUs back into the shared pool *during*
+// StopContainer (the documented early-release workaround), and immediately re-broadens the other
+// shared containers to the freed pool, rather than deferring the release to UnprepareResourceClaims.
+// The existing TestStopContainer only asserts which containers are updated; this asserts the actual
+// release (the shared-pool contents) and the re-broadened cpuset values.
+func TestStopContainerReleasesClaimToSharedPool(t *testing.T) {
+	logger := testr.New(t)
+	allCPUs := cpuset.New(0, 1, 2, 3, 4, 5, 6, 7)
+	var infos []cpuinfo.CPUInfo
+	for _, cpuID := range allCPUs.UnsortedList() {
+		infos = append(infos, cpuinfo.CPUInfo{CpuID: cpuID, CoreID: cpuID, SocketID: 0, NUMANodeID: 0})
+	}
+	mockProvider := &cpuinfo.MockCPUInfoProvider{CPUInfos: infos}
+	topo, _ := mockProvider.GetCPUTopology(logger)
+
+	guaranteedPod := &api.PodSandbox{Id: "pod-id-g", Name: "guaranteed-pod", Namespace: "ns", Uid: "pod-uid-g"}
+	guaranteedCtr := &api.Container{Id: "ctr-id-g", PodSandboxId: guaranteedPod.Id, Name: "guaranteed-ctr"}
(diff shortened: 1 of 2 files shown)


AutuSnow:
/ok-to-test

I reviewed this in the context of #188 and the related future-work threads around mixed allocation, sharing, and preemption.
﻿
The direction here looks good to me.
﻿
The new `StopContainer` doc comment matches the current implementation, and the added test closes a real gap in coverage: the old `TestStopContainer` only checked which containers received updates, while this one also verifies the actual invariant we rely on today:
- the claim CPUs are released back into the shared pool during `StopContainer`
- surviving shared containers are immediately re-broadened to that pool
﻿
I also like that this stays scoped to docs + tests and does not silently change behavior. Given the existing coverage around `UnprepareResourceClaims` and `Synchronize`, this rounds out the current lifetime story nicely.
﻿
LGTM from me.

pkg/driver/nri_hooks_test.go
```diff
@@ -460,3 +460,61 @@ func containerIDsFromUpdates(updates []*api.ContainerUpdate) []string {
...
+	driver := &CPUDriver{
+		podConfigStore:     store.NewPodConfig(),
+		cpuAllocationStore: cpuAllocationStore,
+		claimTracker:       store.NewClaimTracker(),
+		cpuTopology:        topo,
+	}
+	driver.podConfigStore.SetContainerState(types.UID(guaranteedPod.Uid),
+		store.NewContainerState(guaranteedCtr.Name, types.UID(guaranteedCtr.Id), claimUID))
+	driver.podConfigStore.SetContainerState(types.UID(sharedPod.Uid),
+		store.NewContainerState(sharedCtr.Name, types.UID(sharedCtr.Id)))
```
ffromani, line 499: I have mixed feelings about setting private data - we should avoid that as much as we can. So changes which do like this must be evaluate on a case-by-case basis. The general direction is test exercising only public API.

ffromani (commented):
thanks for your contribution! the lifecycle comment is good - but does it belont to StopContainer's docstring or rather as a general comment in `nri_hooks.go` ?

The test per se is nice, but the problem is we access private data. For the current code shape it seems a e2e test may serve us better. Do we have e2e coverage for this flow? how hard would be to add it if missing?

pkg/driver/nri_hooks_test.go
thc1006: Agreed, and the case-by-case framing makes sense to me. For this test I matched the existing setup in `nri_hooks_test.go` (the adjacent `TestStopContainer` and `TestNRISynchronize` build `CPUDriver` with the internal stores and seed via `SetContainerState`), but I am happy to hold the new code to the higher bar.

I can reseed this case through the public hooks: drive `Synchronize` with the guaranteed container's CDI env so the driver builds the stores itself, then call `StopContainer`, which leaves only `cpuTopology` set directly. Want me to rework it that way here, or fold the wider public-API conversion of this file into a separate cleanup? Whichever you prefer works for me.

ffromani: In this case it's ok to follow the existing convention. Better to have a followup sweeping cleanup - but sometimes to enable it we would need quite some refactoring, which can be unpractical.
So we will need to accept some private data testing (but never encourage it or normalize it) for the time being, I'm afraid.

ffromani (commented):
/approve
/lgtm

thanks for the contribution!

Issue #164, opened 2026-06-04
Proposal: Opaque config in ResourceClaim status instead of Individual Mode


pravk03:
**Note:** This is an exploratory proposal. This direction was @johnbelamaric's idea, which came up during a discussion about scalability concerns. Wanted to get general feedback on this direction before exploring it further.

# Background

The DRA CPU driver currently has 2 modes exposed via the `--cpu-device-mode` flag:
1. Grouped Mode (`--cpu-device-mode=grouped`): Exposes CPUs grouped by topology boundaries (e.g., NUMA nodes, sockets). The driver remains in control of selecting which specific CPU cores within the boundaries of the selected device are allocated to the container.
2. Individual Mode (`--cpu-device-mode=individual`): Exposes every CPU core as an distinct device in the `ResourceSlice`. The main benefit this provides it allows users,  or external schedulers to perform fine-grained selection of specific CPU cores at the claim level.

Individual mode while it technically works with default kube-scheduler the scalability challenges prevent it from being used broadly. For all practical purposes is primarily meant for external scheduler use-cases where there is a need to select exact CPUs. This mode presents the below scalability challenges 
1.  Kube-scheduler allocation logic does not scale well with 100's of devices per node/ResourceSlice. This performance bottleneck occurs when a claim requests a general count of devices (e.g., "give me any 4 CPUs") rather than requesting specific device instances (Eg: "give me CPUS 0-3"). This forces the kube-scheduler's allocation algorithm to search for and choose available devices and this performs poorly at scale. This is the main reason why `individual` mode is recommended only with external schedulers.
2. Requesting a large number of devices in a claim easily hits the DRA default claim limit of 32 devices (Issue [#5717](https://github.com/kubernetes/enhancements/issues/5717)). We need to split up requests into multiple claims or increase the default value.
3. Large machines easily hit limits on the maximum number of devices stored per ResourceSlice (Issue [#5718](https://github.com/kubernetes/enhancements/issues/5718)). We need to split into multiple ResourcesSlices. 
4. More etcd storage. 

Additionally, maintaining the two modes creates two separate code paths with very little overlap. https://github.com/kubernetes-sigs/dra-driver-cpu/issues/112 has more discussion and it was agreed that individual mode can be split as a seperate driver.

# Alternate Proposal

We could use `grouped` mode with the below changes as an alternative to Individual mode to achieve fine grain allocation:

[...]

PR #217, opened 2026-06-30
WIP: machine: expose topology information in group-by=machine

ffromani:
#### What type of PR is this?

/kind feature

Address https://github.com/kubernetes-sigs/dra-driver-cpu/issues/164#issuecomment-4831926896

Related to 164 - but not closing it

N/A

#### Release note

```release-note
TBD
```

#### Documentation updates

Readme updates pending

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,16 @@ The driver allocates the specified cores directly (after validating that they ar
 >   - The cores are not reserved using the driver's `--reserved-cpus` configuration flag.
 > - **Error Handling**: If validation fails (e.g. core conflict, size mismatch, duplicate target, or offline cores), the driver returns a failure immediately in Kubelet's `PrepareResourceClaims` hook, causing pod startup to fail.
 
+### Getting topology information
+
+In order to enable external scheduler to make decisions about placement, the `ResourceSlice` must still expose topological information.
+The driver reports the following attributes when operating in `machine` grouping mode:
+
+- `dra.cpu/numa{id}: <cpuset>` (`{id}` being a non-negative integer) - all online CPUs on NUMA node `{id}`, including reserved ones, expressed in the [linux cpuset list notation](https://man7.org/linux/man-pages/man7/cpuset.7.html).
+- `dra.cpu/socket{id}: <cpuset>` (`{id}` being a non-negative integer) - all online CPUs on Socket `{id}`, including reserved ones, expressed in the [linux cpuset list notation](https://man7.org/linux/man-pages/man7/cpuset.7.html).
+
+The driver will report entries for all the NUMA nodes and all the Sockets detected in the system.
+
 ## Prerequisites
 
 The driver relies on [NRI (Node Resource Interface)](https://github.com/containerd/nri) to pin containers to their
--- a/pkg/driver/attributes.go
+++ b/pkg/driver/attributes.go
@@ -29,4 +29,7 @@ const (
 	AttributeCoreID     resourceapi.QualifiedName = "dra.cpu/coreID"
 	AttributeCPUID      resourceapi.QualifiedName = "dra.cpu/cpuID"
 	AttributeNumCPUs    resourceapi.QualifiedName = "dra.cpu/numCPUs"
+
+	AttributePrefixNUMA   = "dra.cpu/numa"
+	AttributePrefixSocket = "dra.cpu/socket"
 )
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -219,6 +219,7 @@ func (cp *CPUDriver) createGroupedCPUDeviceSlices(logger logr.Logger) [][]resour
 				AttributeSMTEnabled: {BoolValue: new(cp.cpuTopology.SMTEnabled)},
 				AttributeNumCPUs:    {IntValue: new(availableCPUs)},
 			}
+			addMachineTopologyAttributes(deviceAttrs, cp.cpuTopology)
 			cp.setPCIeRootsAttribute(deviceAttrs, deviceInfo.cpus.UnsortedList()...)
 			devices = append(devices, resourceapi.Device{
 				Name:                     deviceInfo.name,
@@ -233,6 +234,23 @@ func (cp *CPUDriver) createGroupedCPUDeviceSlices(logger logr.Logger) [][]resour
 		return nil
 	}
 	return [][]resourceapi.Device{devices}
+}
+
+func addMachineTopologyAttributes(attrs map[resourceapi.QualifiedName]resourceapi.DeviceAttribute, topo *cpuinfo.CPUTopology) {
+	for _, numaNodeID := range topo.CPUDetails.NUMANodes().List() {
+		attrName := resourceapi.QualifiedName(fmt.Sprintf("%s%d", AttributePrefixNUMA, numaNodeID))
+		numaNodeCPUs := topo.CPUDetails.CPUsInNUMANodes(numaNodeID)
+		attrs[attrName] = resourceapi.DeviceAttribute{
+			StringValue: new(numaNodeCPUs.String()),
+		}
+	}
+	for _, socketID := range topo.CPUDetails.Sockets().List() {
+		attrName := resourceapi.QualifiedName(fmt.Sprintf("%s%d", AttributePrefixSocket, socketID))
+		socketCPUs := topo.CPUDetails.CPUsInSockets(socketID)
+		attrs[attrName] = resourceapi.DeviceAttribute{
+			StringValue: new(socketCPUs.String()),
+		}
+	}
 }
 
 // CreateCPUDeviceSlices creates Device objects based on the CPU topology.
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -448,6 +448,109 @@ func TestPublishResources(t *testing.T) {
 	}
 }
 
+func TestPublishResourcesGroupedMachine(t *testing.T) {
+	testCases := []struct {
+		name          string
+		cpuInfos      []cpuinfo.CPUInfo
+		reservedCPUs  cpuset.CPUSet
+		expectPublish bool
+		expectedAttrs map[string]string
+	}{
+		{
+			name:          "no cpu data",
+			cpuInfos:      []cpuinfo.CPUInfo{},
+			reservedCPUs:  cpuset.New(),
+			expectPublish: true,
+			expectedAttrs: map[string]string{},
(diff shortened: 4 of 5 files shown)


killianmuldoon:
When trying to run this I'm seeing:

```
E0630 12:08:21.715758       1 dra_hooks.go:571] "processing ResourceSlice objects" err="create resource slice: ResourceSlice.resource.k8s.io \"00000-dra.cpu-validation-k8s-gpu-03-tdgx2\" is invalid: [spec.devices[0].attributes[numa-1]: Invalid value: \"numa-1\": a valid C identifier must start with alphabetic character or   
  '_', followed by a string of alphanumeric characters or '_' (e.g. 'my_name',  or 'MY_NAME',  or 'MyName', regex used for validation is '[A-Za-z_][A-Za-z0-9_]*'), spec.devices[0].attributes[socket-268435456]: Invalid value: \"socket-268435456\": a valid C identifier must start with alphabetic character or '_', followed by a  
  string of alphanumeric characters or '_' (e.g. 'my_name',  or 'MY_NAME',  or 'MyName', regex used for validation is '[A-Za-z_][A-Za-z0-9_]*'), spec.devices[0].attributes[socket-285212672]: Invalid value: \"socket-285212672\": a valid C identifier must start with alphabetic character or '_', followed by a string of           
  alphanumeric characters or '_' (e.g. 'my_name',  or 'MY_NAME',  or 'MyName', regex used for validation is '[A-Za-z_][A-Za-z0-9_]*'), spec.devices[0].attributes[numa-0]: Invalid value: \"numa-0\": a valid C identifier must start with alphabetic character or '_', followed by a string of alphanumeric characters or '_' (e.g.    
  'my_name',  or 'MY_NAME',  or 'MyName', regex used for validation is '[A-Za-z_][A-Za-z0-9_]*')]" logger="ResourceSlice controller.UnhandledError"     
 ```

I think we need underscores rather than hyphens here for the attribute keys.

**Comment by ffromani** [comment]
fair point. Let me do some more testing before we iterate further.

**Comment by killianmuldoon** [comment]
I did a small patch to get past that issue and got a ResourceSlice:

```
spec:
  devices:
  - attributes:
      dra.cpu/numCPUs:
        int: 144
      dra.cpu/smtEnabled:
        bool: false
      numa_0:
        string: 0-71
      numa_1:
        string: 72-143
      socket_268435456:
        string: 0-71
      socket_285212672:
        string: 72-143
    capacity:
      dra.cpu/cpu:
        value: "144"
    name: cpudevmachine
  driver: dra.cpu
  nodeName: validation-k8s-gpu-03
  pool:
    generation: 1
    name: validation-k8s-gpu-03
    resourceSliceCount: 1
```

Let me think about how we would consume it a bit 🤔

AutuSnow:
direction looks right, gives slurm-bridge the per-domain cpuset it needs.
﻿
One caveat while it's WIP: the machine device now gets one attr per NUMA node + one per
socket (plus SMTEnabled/numCPUs/PCIe). That runs into two *enforced* ResourceSlice limits :
﻿
- `ResourceSliceMaxAttributesAndCapacitiesPerDevice = 32` 
- `DeviceAttributeMaxValueLength = 64`
﻿
Both are hard-validated server-side (validation.go: `>32 → Invalid`, `>64 → TooLong`), so a high-socket / Sub-NUMA-Clustering node (or a fragmented cpuset string) makes the *whole* slice rejected and that node publishes nothing. Those big nodes are the slurm-bridge target, so it may show up in real-system testing.
﻿
Could we fail loud instead of silently dropping the slice — e.g. cap what we emit + a `dra.cpu/topologyComplete=false` marker + a high-NUMA/fragmented-cpuset unit fixture?

ffromani:
thanks for pointing out @AutuSnow , that's a real concern. Let's see how we can handle this.

johnbelamaric:
> thanks for pointing out @AutuSnow , that's a real concern. Let's see how we can handle this.

We could always do this in grouped mode, regardless of the type of grouping. Then, for very large nodes we could publish individual device by socket or some other grouping. It's kind of awkward but at least it's workable.

AutuSnow:
I think that fallback is workable. My only concern is that once publication mode changes with topology size, discovery semantics become topology-dependent again, which makes grouped mode harder for consumers to reason about.
﻿
Before going there, I wonder if we should first try a more compact encoding on the machine device itself, for example list-style NUMA/socket topology attributes instead of one attribute per domain. That would keep `group-by=machine` as the actuation path while reducing pressure on the per-device attribute limits.
﻿
If that still does not scale on real large-node topologies, then I agree we probably need a different channel for topology discovery.

ffromani:
Let's summarize the current state. I think there are few directions to explore:
- more compact representation for topology in group-by=machine mode. Worth pursuing anyway, good research direction. Applies to all the following items too.
- keep group-by=machine and expose topology using a different channel
- integrate external scheduler opaque parameters into existing grouping modes (following the direction set by @johnbelamaric in a prior comment)

I think the last point, already briefly touched above, can be pretty defensible and actually reasonnably clean. It can work like this (assume no other changes to the existing behavior if not specified)
- add an explicit opt-in flag like `--external-scheduler` or `--disable-allocator`, final name TBD. Works only in --group-by={socket,numa}
- if the opt-in flag is set, add another attribute to expose the cpuids in each grouped device. This should be, I reckon, the only change the external scheduler integration requires, so the API shape is overall stable.
- the internal cpu allocator is now disabled, and the driver requires to be told which CPU to allocate using the opaque parameters (same shape as group-by=machine)
- if this variant is sufficient to integrate external schedulers (slurm-bridge) we can later keep deprecating individual mode
- we can keep existing group-by machine with the explicit marker @AutuSnow suggested

the only friction becomes the external scheduler has to select first the grouped device and then the cpuids inside that device, but considering the DRA device model and limitations it seems a fair and honestly reasonnably clean to me.

@killianmuldoon / @catblade would the above integration make sense for your usecases and help you?

I will sketch a quick and dirty demo to show how it can look in practice and polish later if you like)

killianmuldoon:
Interesting idea @ffromani. One question:

>the internal cpu allocator is now disabled, and the driver requires to be told which CPU to allocate using the opaque parameters (same shape as group-by=machine)

How would this integrate with CPUs allocated as native non-guaranteed CPUs rather than slurm? Would the base cgroup still be updated?

ffromani:
> Interesting idea @ffromani. One question:
> 
> > the internal cpu allocator is now disabled, and the driver requires to be told which CPU to allocate using the opaque parameters (same shape as group-by=machine)
> 
> How would this integrate with CPUs allocated as native non-guaranteed CPUs rather than slurm? Would the base cgroup still be updated?

I think we can make it work if we keep the device shape fixed but we get the cpu-ids from the api/slurm rather than figuring out using the internal allocator.

Shared cpus will work as intended. We will just have a "passthrough" allocator code. It may need some refactoring to have a clearn seam, but this can be a good shape because it's the best compromise we can have atm (probably also long term?)

machine mode would have been the cleanest, but we hit the attribute limits.

I'll sketch a demo PR ASAP so we can make progress.

pravk03:
Catching up on the discussion here. The WIP PR looks good. 

> Could we fail loud instead of silently dropping the slice — e.g. cap what we emit + a dra.cpu/topologyComplete=false marker + a high-NUMA/fragmented-cpuset unit fixture?

Allowing the validation throw an error because of `ResourceSliceMaxAttributesAndCapacitiesPerDevice` may be a better option to fail loudly. If `topologyComplete` marker is ignored, we would end up failing silently and also having partial info would anyway be unusable to the external scheduler.  

Regarding 
> add an explicit opt-in flag like --external-scheduler or --disable-allocator, final name TBD. Works only in --group-by={socket,numa}

+1 to the idea. But can this be be extended to all grouping modes ?. With this, grouping becomes independent of opaque config support. This has 2 caveats:
1. If the user is using opaque config, its an additional responsibility to match the CPUs being requested through opaque config with the device being requested. 
2. If the user does not care about the detailed topology info and just want to use machine mode, we may hit the number of attribute limitation again.

I don't think we need to immediately worry about this but some options:
1. Add another config to control the level of detail we want in attributes.
2. [Hacky] Metadata-only devices: Exposing just the topology info as attributes in devices with 0 consumable capacity. The external scheduler consolidates the topology info from all these metadata devices, but can only request capacity from the one device with non-zero capacity.
3. [Long term] Leverage [Shared Consumable Capacity](https://github.com/kubernetes/enhancements/pull/5942) to represent hierarchical topology so that the user does not have to decide the grouping mode at all.

AutuSnow:
Thanks @pravk03, good to have another set of eyes here.
On failing loud — I think we actually agree, and you've moved it forward: a topologyComplete=false marker that consumers can ignore is weak, and partial topology is useless to an external scheduler anyway, so I'm happy to drop the "cap + marker" half of my suggestion.
One nuance on "let the ResourceSliceMaxAttributesAndCapacitiesPerDevice validation reject the slice" though: that's loud to the driver operator (an error in logs / a rejected apply), but it's effectively silent to the scheduler and the user — the node simply publishes no ResourceSlice and looks like it has no CPU capacity, which is exactly the failure mode I was worried about on the big slurm-bridge nodes. So my ask is less "marker vs hard error" and more put the failure at the right layer: have the driver detect the over-limit condition itself and surface a precise diagnostic (node, which limit was hit, attribute/value count, offending cpuset) via log + metric + Event, rather than relying on a raw server-side Invalid. That keeps "loud" meaningful for whoever has to debug a node that went missing from schedulable inventory.
On extending the opt-in flag to all grouping modes — +1, decoupling grouping from opaque support is the right instinct, and it's partly what @ffromani's POC in #229 already does (socket/numa can expose the per-device cpuset behind an opt-in). Your caveat #2 (machine mode without detailed topology hits the attribute ceiling again) is really the crux. Of your three options I'd rank them: (1) an attribute-detail/verbosity knob is the pragmatic near-term lever; (3) Shared Consumable Capacity (KEP-5942) is the principled long-term fix since it removes the "pick a grouping mode up front" decision entirely — worth tracking even if we don't block on it; (2) metadata-only 0-capacity devices is clever but pushes consolidation complexity onto every external scheduler and feels fragile as a contract, so I'd keep it as a fallback rather than the plan.
None of this needs to block the WIP — the loud-failure diagnostic (driver-side pre-validation) is the one concrete item I'd like to see land alongside this direction.

ffromani:
keeping open because not completely superceded by #316

ffromani:
most of the changes we discussed are now implemented in #229 and #316 . To move further that point, we need either APIs to get introduced and/or stabilize (Shared Consumable Capacity) or to figure out new API (e.g. unified topology representation API).

PR #220, opened 2026-07-01
 Improve e2e reliability and reuse shared setup helpers

AutuSnow:
This PR improves the reliability and maintainability of the e2e test suite by:
﻿
- probing driver `/healthz` endpoints through the apiserver pod proxy instead of requiring the test runner to reach Pod IPs directly
- making the `ResourceSlice` suite wait for published slices and devices before asserting
- auditing contextual logs across every driver pod instead of sampling a single random worker pod
- skipping e2e suites when required prerequisites such as `KUBECONFIG` or `DRACPU_E2E_TEST_IMAGE` are missing
- reusing shared e2e setup helpers to reduce repeated daemonset/discovery/bootstrap logic across CPU-oriented suites

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,7 @@ PLATFORMS?=linux/amd64,linux/arm64
 LOCAL_PLATFORM?=linux/$(ARCH)
 
 # set convenient defaults for user variables
+DRACPU_E2E_TEST_IMAGE ?= $(IMAGE_TEST)
 DRACPU_E2E_CPU_DEVICE_MODE ?= grouped
 DRACPU_E2E_CPU_GROUP_BY ?= numanode
 DRACPU_E2E_RESERVED_CPUS ?= 0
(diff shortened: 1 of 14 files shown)


AutuSnow:
@ffromani To clarify the scope here: I directly reproduced the “suite gets skipped but still exits successfully” case when e2e prerequisites are missing, which is why I made `make test-e2e` fail fast. For `healthz`, I did not start from a live failure, but I confirmed the assertion had been weakened from explicit `HTTP 200` to implicit `DoRaw` success, so I restored the explicit status-code check and added coverage for 200/503. For the pod proxy RBAC point, I did not hit a denial with the current kubeconfig, but `pods/proxy` is a real prerequisite of this implementation path, so I documented it.

ffromani:
> @ffromani To clarify the scope here: I directly reproduced the “suite gets skipped but still exits successfully” case when e2e prerequisites are missing, which is why I made `make test-e2e` fail fast. For `healthz`, I did not start from a live failure, but I confirmed the assertion had been weakened from explicit `HTTP 200` to implicit `DoRaw` success, so I restored the explicit status-code check and added coverage for 200/503. For the pod proxy RBAC point, I did not hit a denial with the current kubeconfig, but `pods/proxy` is a real prerequisite of this implementation path, so I documented it.

thanks @AutuSnow , at glance they all look like good changes.

AutuSnow:
@ffromani Can you take a look again when you have time to see if the modifications made to this PR are in line with the improvements we will make to e2e in the future

ffromani:
> @ffromani Can you take a look again when you have time to see if the modifications made to this PR are in line with the improvements we will make to e2e in the future

yes, sorry, this PR is on my list but kube 1.37 code freeze is eating a lot of bandwidth. Apologies for the delay.

ffromani:
please rebase (perhaps let us just wait a bit more for #239 to land?) once you're ready to move forward with the review

PR #222, opened 2026-07-01
feat: add numa-mocking to helm chart for local testing

killianmuldoon:
#### What type of PR is this?

/kind feature

This PR adds useful helm options and an example values.yaml for testing on hardware that doesn't expose NUMA information. This is the case when running kind on Docker Desktop for MacOS, but may also be the case for other setups.

Modifications to the script can be used to test a wider range of NUMA / socket scenarios locally.

None of the options exposed in helm are test-only - each enables adding native pod spec options - args, env, volumes, volumeMounts, init containers - which are then used to mount a fake sysfs for testing.

#### Release note

```release-note
NONE
```

#### Documentation updates

- Helm chart docs

### DIFF
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -36,6 +36,10 @@ helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system -f m
 | args.hostnameOverride | string | `""` | Override the node name the driver registers under; omitted when empty |
 | args.logLevel | int | `4` | Log verbosity level passed as `--v` |
 | args.reservedCPUs | string | `""` | CPUs reserved for the OS and kubelet, excluded from DRA management (e.g. `"0-1"`); omitted when empty |
+| extraEnv | list | `[]` | Additional environment variables for the dracpu container |
+| extraInitContainers | list | `[]` | Additional init containers for the DaemonSet pod |
+| extraVolumeMounts | list | `[]` | Additional volume mounts for the dracpu container |
+| extraVolumes | list | `[]` | Additional volumes for the DaemonSet pod |
 | fullnameOverride | string | `""` | Override the full release name |
 | healthzPath | string | `"/healthz"` | Path for liveness and readiness probes |
 | healthzPort | int | `8080` | Port the HTTP server binds to; used for the container port and probes |
--- a/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
+++ b/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
@@ -56,6 +56,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.extraInitContainers }}
+      initContainers:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
       - name: dracpu
         args:
@@ -100,6 +104,10 @@ spec:
         securityContext:
           capabilities:
             add: ["NET_ADMIN", "SYS_ADMIN"]
+        {{- with .Values.extraEnv }}
+        env:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
         volumeMounts:
         - name: device-plugin
           mountPath: /var/lib/kubelet/plugins
@@ -109,6 +117,9 @@ spec:
           mountPath: /var/run/nri
         - name: cdi-dir
           mountPath: /var/run/cdi
+        {{- with .Values.extraVolumeMounts }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
       volumes:
       - name: device-plugin
         hostPath:
@@ -123,3 +134,6 @@ spec:
         hostPath:
           path: /var/run/cdi
           type: DirectoryOrCreate
+      {{- with .Values.extraVolumes }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
--- a/deployment/helm/dra-driver-cpu/values.schema.json
+++ b/deployment/helm/dra-driver-cpu/values.schema.json
@@ -51,6 +51,34 @@
       },
       "additionalProperties": false
     },
+    "extraEnv": {
+      "description": "Additional environment variables for the dracpu container",
+      "type": "array",
+      "items": {
+        "type": "object"
+      }
+    },
+    "extraInitContainers": {
+      "description": "Additional init containers for the DaemonSet pod",
+      "type": "array",
+      "items": {
+        "type": "object"
+      }
+    },
+    "extraVolumeMounts": {
+      "description": "Additional volume mounts for the dracpu container",
+      "type": "array",
+      "items": {
+        "type": "object"
+      }
+    },
+    "extraVolumes": {
+      "description": "Additional volumes for the DaemonSet pod",
+      "type": "array",
+      "items": {
+        "type": "object"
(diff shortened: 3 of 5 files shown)


killianmuldoon (commented):
I also considered adding a patch file to the repo under `hack` which could be applied after helm install. This is ~just as good to me - my main goal is to to have a solution to this problem in the repo for local-developement.

e.g. 

```
  kubectl patch daemonset dracpu --n kube-system --patch-file hack/examples/fake-sysfs-patch.yaml
```

Where the patch looks like:

```
  # hack/examples/fake-sysfs-patch.yaml
  spec:
    template:
      spec:
        initContainers:
          - name: create-fake-sysfs
            image: busybox:1.36
            command: ["/bin/sh", "-ceu"]
            args:
              - |
                root=/work
                discovered_cpus=/tmp/discovered-cpus
                online_cpus_file=/tmp/online-cpus

                : > "${discovered_cpus}"
                for cpu_dir in /sys/devices/system/cpu/cpu[0-9]*; do
                  [ -d "${cpu_dir}" ] || continue

                  if [ -f "${cpu_dir}/online" ] &&
                     [ "$(cat "${cpu_dir}/online")" != "1" ]; then
                    continue
                  fi

                  cpu=${cpu_dir##*/cpu}
                  printf '%s\n' "${cpu}" >> "${discovered_cpus}"
                done

                sort -n "${discovered_cpus}" > "${online_cpus_file}"
                num_cpus=$(wc -l < "${online_cpus_file}")

                if [ "${num_cpus}" -eq 0 ]; then
                  echo "no online CPUs discovered" >&2
                  exit 1
                fi

                node0_count=$(((num_cpus + 1) / 2))
                online_cpus=
                node0_cpus=
                node1_cpus=
                index=0

                while read -r cpu; do
                  if [ -n "${online_cpus}" ]; then
                    online_cpus="${online_cpus},${cpu}"
                  else
                    online_cpus=${cpu}
                  fi

                  if [ "${index}" -lt "${node0_count}" ]; then
                    if [ -n "${node0_cpus}" ]; then
                      node0_cpus="${node0_cpus},${cpu}"
                    else
                      node0_cpus=${cpu}
                    fi
                  else
                    if [ -n "${node1_cpus}" ]; then
                      node1_cpus="${node1_cpus},${cpu}"
                    else
                      node1_cpus=${cpu}
                    fi
                  fi

                  index=$((index + 1))
                done < "${online_cpus_file}"

                mkdir -p \
                  "${root}/devices/system/cpu/smt" \
                  "${root}/devices/system/node"

                printf '%s\n' "${online_cpus}" \
                  > "${root}/devices/system/cpu/online"
                printf 'off\n' \
                  > "${root}/devices/system/cpu/smt/control"

                index=0
                while read -r cpu; do
                  if [ "${index}" -lt "${node0_count}" ]; then
                    node=0
                    core=${index}
                    shared=${node0_cpus}
                  else
                    node=1
                    core=$((index - node0_count))
                    shared=${node1_cpus}
                  fi

                  cpu_root="${root}/devices/system/cpu/cpu${cpu}"

                  mkdir -p \
                    "${cpu_root}/topology" \
                    "${cpu_root}/node${node}" \
                    "${cpu_root}/cache/index3"

                  printf '%s\n' "${node}" \
                    > "${cpu_root}/topology/physical_package_id"
                  printf '%s\n' "${core}" \
                    > "${cpu_root}/topology/core_id"
                  printf '3\n' \
                    > "${cpu_root}/cache/index3/level"
                  printf '%s\n' "${node}" \
                    > "${cpu_root}/cache/index3/id"
                  printf '%s\n' "${shared}" \
                    > "${cpu_root}/cache/index3/shared_cpu_list"

                  index=$((index + 1))
                done < "${online_cpus_file}"

                mkdir -p "${root}/devices/system/node/node0"
                printf '%s\n' "${node0_cpus}" \
                  > "${root}/devices/system/node/node0/cpulist"

                if [ -n "${node1_cpus}" ]; then
                  mkdir -p "${root}/devices/system/node/node1"
                  printf '%s\n' "${node1_cpus}" \
                    > "${root}/devices/system/node/node1/cpulist"
                fi
            volumeMounts:
              - name: fake-sysfs
                mountPath: /work
        containers:
          - name: dracpu
            args:
              - /dracpu
              - --v=4
              - --cpu-device-mode=grouped
              - --group-by=numanode
              - --bind-address=:8080
            env:
              - name: HOST_ROOT
                value: "/"
            volumeMounts:
              - name: fake-sysfs
                mountPath: /sys
                readOnly: true
        volumes:
          - name: fake-sysfs
            emptyDir: {}
```

killianmuldoon:
We could also use this in e2e tests if we wanted to put this change in that path.

AutuSnow:
In my opinion, it is good and has indeed solved a real pain point in local development. For me, LGTM
/cc @ffromani @fmuyassarov

ffromani:
Conceptually +1 for me about mocking sysfs but I'd like to explore first if we can use the overlay concept we explored in #156 . I think the overlay concept will address this usecase in a cleaner way AND will have a very logical extension to feed sub-NUMA locality information to the driver, e.g. overriding pcieRoot or LLC locality.

ffromani:
...and I realized the overlay concept and PoC got lost in the force-pushes. Let me try to recover something from my local trees/notes.

ffromani:
overlayfs: a yaml document which represent an object. The keys are sysfs full paths. The values are a string which represent the content of the file that would be present at that path. The driver is optionally given the yaml file. The overlay is never hot-reloaded: it is processed once at startup. If the yaml file (aka the overlay) is given, every time the driver tries to access sysfs, it will try to resolve the full path against the overlay. If the path matches (fully), it will consume the data from the overlay and never hit the real sysfs.
The overlay can be partial: you can just wrap a single path or all the paths the driver needs, up to the user.
The driver would log at startup the data the overlay contains (at V>=4 probably).

[EDIT]
we should be able to implement with little refactoring into our codebase leveraging the recent `fs.FS` golang stdlib addition

[EDIT2] Actually partially wrong, we would need a refactoring/cleanup first to port `pkg/cpuinfo` to our `SysFS` interface which `pkg/device` is already using.

killianmuldoon:
>Let me try to recover something from my local trees/notes.

I'd definitely be happy to take a look and help out with it - at least with testing it for my local env

AutuSnow:
I agree the current fake-sysfs Helm path solves a real local-development pain point. I had forgotten the #156 overlay/merge concept, and I agree that it is a cleaner long-term direction if we can recover it. So I see this PR as a useful workaround, but not the same design as the overlay approach.😢

ffromani:
@killianmuldoon recovered the notes in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/222#issuecomment-4856649702 - happy to discuss further and clarify. If you like and agree about this direction feel free to work on this. Otherwise I'll post the `pkg/cpuinfo` cleanup/refactoring later this week.

killianmuldoon:
>@killianmuldoon recovered the notes in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/222#issuecomment-4856649702 - happy to discuss further and clarify.

I'll take a look tomorrow and see if I can spec something out.

ffromani:
if we decide to go forward with the `overlayfs` concept, as I hope, we would need to port the `cpuinfo` package to our little `SysFS` interface. https://github.com/kubernetes-sigs/dra-driver-cpu/pull/223 would enable that work.

killianmuldoon:
Will close this one in favor of #225

Issue #204, opened 2026-06-25
Improve NRI test assertions and error matching


AutuSnow:
Follow-up from #193.
﻿
The PR review raised two testing improvements for the NRI paths:
﻿
1. Avoid matching errors by string where possible.
2. Prefer stricter public-behavior assertions without inspecting internal data structures directly.
﻿
This issue tracks improving the NRI unit tests so they remain precise while testing through observable behavior.
﻿
Scope:
- Review NRI tests that match exact error strings.
- Replace string matching with typed errors, sentinel errors, helper predicates, or more stable assertions where appropriate.
- Look for places where tests were loosened to avoid checking internal state.
- Reintroduce stronger checks through public behavior when practical.
- Keep the tests focused; avoid turning unit tests into broad integration tests unless that is intentional.
﻿
This is not required for #193. It is follow-up test maintainability work, likely suitable for the v0.3.0 cycle.

PR #224, opened 2026-07-02
driver: improve NRI test assertions and error matching

AutuSnow:
Related to #204.

This PR improves NRI unit test maintainability by replacing fragile error-string matching with stable error-category assertions, and by tightening observable-behavior checks where practical.

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -18,6 +18,7 @@ package driver
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -27,6 +28,13 @@ import (
 	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/store"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/cpuset"
+)
+
+var (
+	errMalformedDRAEnv         = errors.New("malformed DRA env entry")
+	errParseCPUSet             = errors.New("failed to parse cpuset value")
+	errClaimNotPrepared        = errors.New("claim is not prepared by this driver")
+	errClaimAllocationMismatch = errors.New("claim cpuset does not match prepared allocation")
 )
 
 // Synchronize is called by the NRI to synchronize the state of the driver during bootstrap.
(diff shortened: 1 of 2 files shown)


AutuSnow:
@ffromani PTAL

pkg/driver/nri_hooks.go
```diff
@@ -109,7 +117,7 @@ func parseDRAEnvToClaimAllocations(logger logr.Logger, envs []string) (map[types
 		logger.V(4).Info("parsing DRA env entry", "env", env)
 		parts := strings.SplitN(env, "=", 2)
 		if len(parts) != 2 {
-			return nil, fmt.Errorf("malformed DRA env entry %q", env)
+			return nil, fmt.Errorf("%w: %q", errMalformedDRAEnv, env)
```
ffromani: the idea is good and I support it but we somehow use the wrapping backwards. Typically the wrapped error is the last item in the new error construction and each wrapping prepends data to enrich the context of the error. We may want to review a bit how we do handle error wrapping and construct the error chain. I don't have bright suggestions atm.

Apply for all the other changes into this file.

ffromani (commented):
Good idea, but the way we construct and wrap errors seems awkward. Comments inline.

pkg/driver/nri_hooks.go
AutuSnow: Updated.
﻿
This now uses a context error wrapper that keeps `errors.Is(...)` matching for the sentinel while making the underlying cause the main unwrap target.

thc1006:
Read through this on latest main, built and ran the `pkg/driver` tests. Moving to sentinel errors with `require.ErrorIs` (and matching the wrapped cause, e.g. `strconv.ErrSyntax`) is a solid robustness win over the old `err.Error()` string checks. I also like that `validatePreparedClaimAllocation` lands with both fail-closed paths covered, the unprepared-claim and the cpuset-mismatch cases, so the new validation is pinned down rather than just added.

The fail-closed choice reads right to me: rejecting a container whose env cpuset does not match what was prepared beats pinning an unvalidated set. Nothing blocking from me.

thc1006:
Re-checked on the latest push; the unused `containerIDsFromUpdates` cleanup reads fine.

One correction to my earlier comment, to keep the attribution accurate: `validatePreparedClaimAllocation` and both fail-closed cases (unprepared-claim and cpuset-mismatch) are already on `main`, so they are not new in this PR. What changes for those paths is the error style: `fmt.Errorf` strings become sentinel errors, and the tests match with `errors.Is` instead of `err.Error()` substrings. That is the real win here, and matching `strconv.ErrSyntax` through `Unwrap` is a good touch. Still nothing blocking from me.

ffromani:
before I move forward with another round of review, please check the commits metadata, it seems there are multiple accounts involved? 
image
Let's make sure the attributions are correct, both in general to give proper credit and to fulfill the CLA.

AutuSnow:
This is because the email used to sign the CLA is a personal email, but there is a conflict with the company email bound globally in Git. I will handle it

Issue #216, opened 2026-06-30
[admin] set up a design doc process


ffromani:
**What would you like to be added**:

We need to add a *extremely light weight* design doc process and knowledge base in order to discuss nontrivial enhancements and have a clear record of the decisions.
*extremely light weight* is paramount: we want to add the minimum possible red tape and rather consolidate the lessons learned and recurring concerns. People familiar with the KEP process should find some similarities but our process can and should be MUCH lighter.

**Why is this needed**:
Nowadays design, tradeoffs, API concerns, upgrade flow is all scattered across PR review comments, issue comments, some slack threads. The information is somehow availabile, just unnecessarily hard to find out.

The goal of this process is to make it easy to consolidate this information more than else.

AutuSnow:
Finally here, I have solved the trouble I couldn't find when reviewing and discussing in my daily life

ffromani:
Thanks for the feedback! It's very important for us to find the right balance and set the right tone. We have few contributors, this project _needs_ to keep velocity and momentum, and this project is exploratory in nature: we're (literally) working on the frontier of kubernetes capabilities. Yet we can use a better bookkeeping for our own sake!
I'll add some examples of information dispersed in PR iterations/issues and will sketch a template later. 

Notice: I will INTENTIONALLY err on the side of light process. For the aforementioned reasons we need to travel light and add process only if really really warranted. We won't go anywhere near the KEP full process anytime soon (maybe post 1.0.0, we'll see).

ffromani:
examples of issues whose design and key details emerged during review and comments. Worth reiterating: this is expected and healthy for this project at this maturity stage. The issue is not iterating on the design, let alone review comments which flow back in the design: that's actually the process working as intended! the issue is consolidating all the "why"s in a place easy to find.

- #163 was the final iteration of a design emerged during a review comment of #156. In turn #156 was the redoing of the draft sketched in #68! 
- in #83 we added helm chart, which is a great addition. Upgrade concern were worth documenting and we did as by product of review comments.
- #164 added machine mode. The API contract for scheduler input emerged in review; the missing topology data emerged as issue comment

In all these examples, the only missing chain is "consolidating information", everything else is either unavoidable or works as intended. And the "consolidating information" is exactly the piece we want to fix here.

kannon92:
[...]

PR #225, opened 2026-07-02
Add sysfs overlay

killianmuldoon:
#### What type of PR is this?

/kind feature

This PR adds an in-memory sysfs overlay which allows mocking system information for testing and local development. It allows users to configure a set of paths like:

```
sysfsOverlay:
  /sys/devices/system/cpu/online: "0-3\n"
  /sys/devices/system/cpu/smt/control: "off\n"
  /sys/devices/system/node/node0/cpulist: "0-1\n"
  /sys/devices/system/node/node1/cpulist: "2-3\n"
```

This is useful to test an array of node topologies without needing special hardware. It's also useful for testing on systems that don't expose topology as expected e.g. kind running on Docker Desktop for MacOS.

This supersedes #222 which served a similar purpose but used an init container to write files.

#### Release note

```release-note

```


pkg/cpuinfo/cpuinfo.go
```diff
@@ -464,19 +473,17 @@ func populateCpuSiblings(cpuInfos []CPUInfo) {
 	}
 }
 
-// ReadFile reads contents from a file.
-func ReadFile(filename string) (string, error) {
-	data, err := os.ReadFile(filename)
+func readFile(sysfs fs.FS, filename string) (string, error) {
```
killianmuldoon, line 470: I'm not sure if this repo has a go API compatibility. I made this function private when changing it as I didn't think it needed to be exported. 

If needed I can add back in the original exported ReadFile to maintain compatibility.

pkg/cpuinfo/cpuinfo.go
```diff
@@ -464,19 +473,17 @@ func populateCpuSiblings(cpuInfos []CPUInfo) {
...
 		return "", err
 	}
 
 	return string(data), nil
 }
 
-// ReadLines reads contents from a file and splits them by new lines.
-func ReadLines(filename string) ([]string, error) {
-	data, err := os.ReadFile(filename)
+func readLines(sysfs fs.FS, filename string) ([]string, error) {
```
killianmuldoon, line 479: Same as https://github.com/kubernetes-sigs/dra-driver-cpu/pull/225/changes#r3513013878

deployment/helm/dra-driver-cpu/README.md
```diff
@@ -57,6 +57,7 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
 | resources.requests.cpu | string | `"100m"` | CPU resource request |
 | resources.requests.memory | string | `"50Mi"` | Memory resource request |
 | serviceAccount.annotations | object | `{}` | Annotations to add to the ServiceAccount |
+| sysfsOverlay | object | `{}` | Sysfs file contents keyed by absolute /sys path; loaded once when the driver starts |
```
killianmuldoon: Not sure this is the best way to expose this in helm as it's really just a testing value.

Open to ideas here

killianmuldoon:
If we're happy with this approach I can take a stab at implementing it inside the e2e tests

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -34,9 +34,14 @@ spec:
         {{- with .Values.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.podAnnotations }}
+      {{- if or .Values.podAnnotations .Values.sysfsOverlay }}
```
killianmuldoon: This gives us auto-reload if the configMap changes - useful for local testing.

pkg/cpuinfo/cpuinfo.go
ffromani: short answer: what you did is fine.
long answer: we have no formal compatibility promise for packages, and even if we will add it (which is an option totally on the table, we just didn't discuss it yet). It will happen not before the project reaches beta level quality.

killianmuldoon: Sounds good - I can move ~everything into `/internal` as a follow up if you'd like?

ffromani: let's defer that to a sweeping change we can discuss later. We will do a cleanup before 0.3.0 for sure.
To explain a bit, the distinction between `internal` and `pkg` is also tentative and not well defined. Let us discuss jointly a policy/agreement (could just be a couple lines on some README) and then the move becomes trivial. Could actually be a great testbed for #216

deployment/helm/dra-driver-cpu/README.md
ffromani: I have no strong opinions here but I'd consider if we should expose in helm at all.
To be super-explicit: if we do, fine by me. I'm just saying let's not just assume we _must_ do it.

pkg/cpuinfo/cpuinfo.go
```diff
@@ -146,11 +148,19 @@ type CPUTopology struct {
...
-// NewSystemCPUInfo creates a new SystemCPUInfo.
+// NewSystemCPUInfo creates a new SystemCPUInfo backed by the host sysfs.
 func NewSystemCPUInfo() *SystemCPUInfo {
-	return &SystemCPUInfo{}
+	sfs := os.DirFS(hostSys()).(sysfs.FS)
+	return NewSystemCPUInfoFromFS(sfs)
+}
+
+// NewSystemCPUInfoFromFS creates a new SystemCPUInfo backed by sysfs.
+func NewSystemCPUInfoFromFS(sfs sysfs.FS) *SystemCPUInfo {
```
ffromani: we're not bound by backward compat premise, so we can always require a `sys.FS` argument, to be consistent with the `device` package is doing. Change should be mechanical, so little bother.
You can add a one-liner Host() helper function returning os.DirFS(Root).(FS) in `sysfs` package to push back the repetition.
In general, I'd be pushing back against shallow functions: not hard nack every time, but deeper scrutiny shallower the function. In this case this trivial oneliner embodies a concept and names a semantic intent, so it seems worth the price

ffromani: TL;DR: please rename NewSystemCPUInfoFromFS as NewSystemCPUInfo and drop the old variant.

pkg/cpuinfo/cpuinfo.go
```diff
@@ -265,7 +274,7 @@ func (s *SystemCPUInfo) GetCPUInfos(logger logr.Logger) ([]CPUInfo, error) {
 			cpuInfo.CoreType = CoreTypeStandard
 		}
 
-		if err := populateTopologyInfo(&cpuInfo, logger); err != nil {
+		if err := s.populateTopologyInfo(&cpuInfo, logger); err != nil {
```
ffromani: prefer free functions if possible - reduce coupling.
Then, in general, when the free functions start to take too many (tipically more than 3-4) or too different parameters, we can check which abstraction emerges and refactor accordingly.
In this case the old populateTopologyInfo can take a sysfs.FS parameter.

pkg/cpuinfo/cpuinfo.go
```diff
@@ -278,12 +287,12 @@ func (s *SystemCPUInfo) GetCPUInfos(logger logr.Logger) ([]CPUInfo, error) {
...
 }
 
-func populateTopologyInfo(cpuInfo *CPUInfo, logger logr.Logger) error {
+func (s *SystemCPUInfo) populateTopologyInfo(cpuInfo *CPUInfo, logger logr.Logger) error {
 	cpuID := cpuInfo.CpuID
 
 	// Get Socket ID from sysfs
-	socketPath := hostSys(fmt.Sprintf("devices/system/cpu/cpu%d/topology/physical_package_id", cpuID))
-	socketStr, err := ReadFile(socketPath)
+	socketPath := path.Join("devices", "system", "cpu", fmt.Sprintf("cpu%d", cpuID), "topology", "physical_package_id")
```
ffromani, line 288: good change, thanks

pkg/cpuinfo/sysfs_overlay_test.go
```diff
@@ -0,0 +1,59 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
ffromani: per newer (2026) kubernetes practices we should omit the copyright year

pkg/device/overlay_fs.go
```diff
@@ -0,0 +1,226 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
ffromani: This is more related than `sysfs` than else (and it is almost unrelated to `pkg/device`) so I'd move into `pkg/sysfs/overlay.go`. This also keeps the inter-pkg deps cleaner.

pkg/device/sysfs_overlay.go
```diff
@@ -0,0 +1,95 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
ffromani: likewise, please move this into `pkg/sysfs`. I don't mind if you keep this file separated from `pkg/sysinfo/overlay.go` but I'd see no problem in merging the content in just one file.

pkg/device/sysfs_overlay.go
```diff
@@ -0,0 +1,95 @@
...
+	"path"
+	"strings"
+
+	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/sysfs"
+	"sigs.k8s.io/yaml"
+)
+
+// ParseSysFSOverlay parses a YAML object whose keys are absolute sysfs paths
+// and whose values are the file contents to return for those paths.
+func ParseSysFSOverlay(data []byte) (map[string]string, error) {
```
ffromani: please evaluate if `FromYAMLString` (or some variation of) can be a good alternative name

pkg/device/sysfs_overlay.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		overlay[overlayPath] = contents
+	}
+	return overlay, nil
+}
+
+// NewOverlaySysFS returns a read-through sysfs which serves exact matches from
+// overlay and delegates all other operations to base. The overlay is copied
+// and is therefore immutable after this function returns.
+func NewOverlaySysFS(base sysfs.FS, overlay map[string]string) (sysfs.FS, error) {
```
ffromani: since our only current (and planned) way to add an overlay is using a YAML file, I won't mind having just a single function to create an overlay from a `base sysfs.FS` and a YAML byte slice. 
Something like: `NewOverlayFromYAML(base sysfs.FS, data []byte) (sysfs.FS, error)` would keep the API surface minimal.
We can trivially re-extract the generic function you have here later. Or do we plan another way to construct overlays?

ffromani (commented):
thanks, had a first pass. Need to do another pass about tests and doublecheck some implementation details but I'm already quite happy with the PR shape and I think we can converge quickly. About helm, I'll also review later, honestly I don't have strong opinions there.

pkg/cpuinfo/sysfs_overlay_test.go
killianmuldoon: First time adding a file in a while 😅 

ref for the change is here: https://github.com/kubernetes/kubernetes/pull/134835

pkg/device/sysfs_overlay.go
killianmuldoon: I think you're right - it's small enough that one file will work.

pkg/device/sysfs_overlay.go
killianmuldoon: Sounds good - I'll simplify it and make it yaml-only. This makes the surface area a lot smaller which is nice.

pkg/device/sysfs_overlay.go
killianmuldoon: I'll do NewOverlayFromYaml here so the call will be `sysfs.NewOverlayFromYaml`. What do you think?

cmd/dracpu/app.go
```diff
@@ -184,6 +192,7 @@ func run(logger logr.Logger) error {
 	}
 	driverProviders := driver.Providers{
 		K8SClient: clientset,
+		SysFS:     sfs,
```
AutuSnow, line 194: I might be missing something here — the overlay is wired into Providers.SysFS, but I don't see it reaching Providers.CPUInfo. Since CPUInfo is left unset, EnsureCPUInfo() falls back to cpuinfo.NewSystemCPUInfo(), which builds its own os.DirFS(hostSys()) — so GetCPUTopology() → populateTopologyInfo() reads the real host sysfs, bypassing the overlay entirely.
﻿
If I'm reading that right, only /sys/devices/system/cpu/online (read via OnlineCPUs(providers.SysFS)) actually gets overlaid — the socket/core/NUMA/cpulist/L3/smt entries in hack/examples/values_sysfs_overlay.yaml (29 of the 30 lines) would have no effect, which seems to defeat the topology-mocking purpose on a NUMA-less kind node.
﻿
It looks like NewSystemCPUInfoFromFS was added for exactly this, so I think it's just a missing line:
`
driverProviders := driver.Providers{
      K8SClient: clientset,
      SysFS:     sfs,
      CPUInfo:   cpuinfo.NewSystemCPUInfoFromFS(sfs),
}
`
Could you confirm whether the topology path is meant to go through the overlay? If so, an app-level test asserting "overlay set → topology comes from overlay" would lock this in — the existing unit test proves NewSystemCPUInfoFromFS works, but nothing currently covers the wiring.

cmd/dracpu/app.go
```diff
@@ -219,6 +228,38 @@ func run(logger logr.Logger) error {
 	return fatalErr
 }
 
+func newSysFS(logger logr.Logger, overlayPath string) (sysfs.FS, error) {
+	sfs := os.DirFS(sysfs.Root).(sysfs.FS)
```
AutuSnow: Small thing I wanted to check: the overlay base is os.DirFS(sysfs.Root) (i.e. /sys), whereas cpuinfo reads through hostSys(), which honors HOST_ROOT. When HOST_ROOT != "/" (driver in a container reading the host's /sys), wouldn't the overlay end up layered over the container's /sys rather than the host's? It's harmless when HOST_ROOT="/" (the kind/Docker-Desktop case), but once the overlay is also feeding CPUInfo (see the other comment), basing it on hostSys() too would keep the two consistent. Am I right that they can diverge here?

pkg/device/sysfs_overlay.go
ffromani: SGTM!

cmd/dracpu/app.go
killianmuldoon: This will change a bit with some updates - but in this version  EnsureCPUInfo()  falls back to cpuinfo.NewSystemCPUInfoFromFS(pr.EnsureSysFS()), not cpuinfo.NewSystemCPUInfo(). Then the topology discovery goes through the overlay through providers.SysFS

For the wiring / testing - I'll add some test demonstrating this.

cmd/dracpu/app.go
killianmuldoon: You're right here - this doesn't respect HOST_ROOT. I'll make sure I keep it working

cmd/dracpu/app.go
AutuSnow: You're completely right, and thanks for the pointer — I was reviewing against a stale local checkout and, worse, skimmed past the one-line driver.go change in the diff. Routing it through EnsureCPUInfo() → NewSystemCPUInfoFromFS(pr.EnsureSysFS()) is actually cleaner than injecting at the call site, since both providers share a single sysfs. So the topology path does go through the overlay — please disregard the "missing wiring" concern entirely, that was my mistake. The follow-up tests you mentioned would be a nice belt-and-suspenders, but the behavior itself is correct as-is. 👍

killianmuldoon:
Added a number of changes to address review comments and fix some other bugs I found along the way.

hack/examples/sysfs-overlay/values.yaml
```diff
@@ -0,0 +1,16 @@
+args:
+  exposePCIeRoots: true
+
+extraArgs:
+  - --sysfs-overlay=/etc/dra-driver-cpu/sysfs-overlay.yaml
+
+extraVolumeMounts:
```
killianmuldoon, line 7: Helm implementation now uses generic extension points which is overally much better, even if the UX is slightly worse.

deployment/helm/dra-driver-cpu/README.md
fmuyassarov: I don't see a problem with this either, although I don't expect production users to be consuming it. If developers are using local Helm charts then it makes sense to add this, unless we have another way to patch the templates during installation.

ffromani: We can keep the helm integration for now as noone seems to be bothered. But please add a clear mention in the README.md that this feature is meant for development and should not be used in production deployments.

fmuyassarov: and I guess we are reaching the point that we soon we will need a separate doc for developers of the project, probably that will included running tests, linting, doc/scheme re-generation, image building, uploading and consuming etc.

killianmuldoon: I updated the helm integration in the latest set of commits to be more generic - it uses extraVolumeMounts and extraVolumes instead of this sysfsOverlay value.

ffromani: > and I guess we are reaching the point that we soon we will need a separate doc for developers of the project, probably that will included running tests, linting, doc/scheme re-generation, image building, uploading and consuming etc.

Yes, this is one of the things we need to address for 0.3.0 (https://github.com/kubernetes-sigs/dra-driver-cpu/issues/226)

hack/examples/sysfs-overlay/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Create the overlay ConfigMap and deploy the chart:
+
+```bash
+kubectl apply -n kube-system -f hack/examples/sysfs-overlay/configmap.yaml
+
+helm upgrade --install dra-driver-cpu deployment/helm/dra-driver-cpu \
+  --namespace kube-system \
+  --set fullnameOverride=dracpu \
+  -f hack/examples/sysfs-overlay/values.yaml
```
fmuyassarov, line 28: One thing to note is that it will use the default image, which is currently v0.2.0 and doesn't include your changes. Until the next release, I think we should either add a quick note reminding users to rebuild the image or include a step for rebuilding/patching the templates during install.

killianmuldoon: Good catch - I've changed this flow to one that includes a new image build.

cmd/dracpu/app.go
killianmuldoon: I've added a test covering more of the e2e flow here https://github.com/kubernetes-sigs/dra-driver-cpu/pull/225/changes/7992bf67ab65f87dd51b4a0290ca4b99874af13d

cmd/dracpu/app_test.go
```diff
@@ -0,0 +1,210 @@
+/*
```
ffromani, line 1: tests in `cmd/dracpu` are always a smell to me. Code here should be doing orchestration only and tests should preferrably be done through e2e_local or at a different level. In this case, concerning code in `cmd`, the tests seem to only depend on `newSysFS` which is however trivial and I won't mind reimplementing it in the tests. About the behavior - the more interesting bit

TestSysFSOverlayFeedsCPUInfoProvider seems almost entirely a pure sysfs overlay + cpuinfo topology test, which I'd move in `pkg/cpuinfo`.

TestSysFSOverlayUsesHostRootAsBase checks HOST_ROOT being honored, and I'd move in `pkg/sysfs`.

TestSysFSOverlayUsesHostRootForTopology again concerns mostly about `pkg/cpuinfo` - kinda high level test, and I'd move it there.

We can do this in a followup PR.

internal/gatherinfo/gatherinfo.go
```diff
@@ -190,7 +191,17 @@ func createOutputFile(parentDir string, now time.Time, pid int) (*os.File, strin
 }
 
 func collectReport(logger logr.Logger, defaults driverconfig.Config, driverCmdlinePath string) (Report, error) {
-	sys := cpuinfo.NewSystemCPUInfo()
+	driverConfig := detectDriverConfig(defaults, driverCmdlinePath)
+	base := os.DirFS(cpuinfo.GetEnv("HOST_ROOT", "/", "sys")).(sysfs.FS)
```
ffromani, line 195: this part seems OK, but the changes we are doing highlight the need for a sweeping pass to make sure all the codebase honors `HOST_ROOT` consistently. We can probably use a couple helpers and some trivial targeted rewrite. E.g.:
```
diff --git a/pkg/sysfs/sysfs.go b/pkg/sysfs/sysfs.go
index 2bc52bd..41c9b9c 100644
--- a/pkg/sysfs/sysfs.go
+++ b/pkg/sysfs/sysfs.go
@@ -18,13 +18,17 @@ package sysfs
 
 import (
        "io/fs"
+       "os"
+       "path"
 )
 
-const (
-       Root = "/sys"
-)
+const Root = "/sys"
 
 type FS interface {
        fs.ReadLinkFS
        fs.ReadDirFS
 }
+
+func Host() FS {
+       return os.DirFS(path.Join(os.Getenv("HOST_ROOT"), Root)).(FS)
+}
```

Then use `sysfs.Host()` everywhere, and fix test/image/dracputester/app.go to consume it (note inside the tester image we should never set HOST_ROOT. Using Host() here is for consistency)

internal/gatherinfo/gatherinfo_test.go
```diff
@@ -94,6 +94,73 @@ func TestRunWritesReportToStdout(t *testing.T) {
...
+	writeTestFile(t, filepath.Join(driverRoot, strings.TrimPrefix(overlayPath, "/")), []byte(`
+/sys/devices/system/cpu/smt/control: "on\n"
+/sys/devices/system/cpu/cpu0/topology/physical_package_id: "7\n"
+`))
+	hostRoot := setupFakeHost(t, []byte("/dracpu\x00--sysfs-overlay="+overlayPath+"\x00"))
+	if err := os.Symlink(driverRoot, filepath.Join(hostRoot, "proc", "42", "root")); err != nil {
+		t.Fatalf("create driver root link: %v", err)
+	}
+
+	stdout, err := captureStdout(t, func() error {
```
ffromani, line 109: I think porting this test to `test/e2e_local` would allow to simplify the code considerably

ffromani (commented):
/lgtm


In the inline comments I identified some possible followup cleanups. Some of them have been highlighted by this PR changes, but are loosely related to this work so I think is best to save them for later. We can do them in future PRs (and I can take care as well).

I'm happy with the general shape of the PR and I think we can merge. Thanks for pivoting and implementing the overlay concept!

ffromani:
@AutuSnow / @fmuyassarov folks, any pending concern ? if you're happy we can merge

cmd/dracpu/app_test.go
killianmuldoon: Makes sense - I was looking for somewhere to plugin an e2e unit test to make sure everything plays nice together. Happy to move them in a follow-up.

internal/gatherinfo/gatherinfo.go
killianmuldoon: Combining it into a single `Host()` makes sense - I can take it as a follow-up

I was able to test this manually by combining this change with the approach from #222 - I think e2e might be the only way to have real confidence the flow works and stays working.

internal/gatherinfo/gatherinfo_test.go
killianmuldoon: Agreed - there's some outstanding issues with getting e2e testing working - but I'm happy to follow up on that.

fmuyassarov:
Pretty good in my opinion and thanks for the addition @killianmuldoon .
/lgtm
leaving to @AutuSnow to remove the hold if looks good or add other comments if there is any.

internal/gatherinfo/gatherinfo.go
ffromani: additional thoughts for the sake of going on record: the cross-dep with `cpuinfo` is a smell. The idea is:
- add `sysfs.Host()` - which should handle `HOST_ROOT` internally
- add `cpuinfo.ProcfsRoot()` - which will return the procfs root path, also handling `HOST_ROOT`
- with the 2 added and used consistently, all the codebase should now consistently honor `HOST_ROOT` (I think we have gaps now, but predating and loosely coupled to this PR)
- and we can now remove `hostSys` `hostRoot` `GetEnv` from `cpuinfo.go`

ffromani:
/approve

agree with above assessment

internal/gatherinfo/gatherinfo_test.go
AutuSnow: I can see the appeal of moving these into `test/e2e_local` for consistency.
﻿
I took a look at that option, but my impression was that it would still require most of the same setup: the gatherinfo-specific `/proc/<pid>/root` and `/proc/<pid>/cwd` resolution still needs the fake `HOST_ROOT/proc` layout, and we'd also need extra binary setup for the `dracpu-gatherinfo` entrypoint.
﻿
Since the lower-level overlay parsing and topology behavior is already covered in `pkg/sysfs`, `pkg/cpuinfo`, and `cmd/dracpu`, I was leaning toward keeping these gatherinfo-specific path-resolution tests here for this PR, and treating any `e2e_local` smoke coverage as a follow-up if we want that additional layer.

ffromani: Fair point, that's a reasonnable take. I'm not pushing for that port. If makes sense, we can port. New test can exercise that option. If it's a close call, no need to change for the sake of change. The only important point is that `e2e_local` tests are a real option now.

PR #227, opened 2026-07-06
fix: refactor sysfs HOST_ROOT handling and tests

killianmuldoon:
#### What type of PR is this?

/kind cleanup

This is a follow-up to https://github.com/kubernetes-sigs/dra-driver-cpu/pull/225. Two separate commits cover two asks:

1) Ensuring HOST_ROOT is handled consistently across the codebase 
2) Moving tests from app_test to lower level packages

#### Release note

```release-note

```


killianmuldoon (commented):
There's a couple of usages of HOST_ROOT that I'm not sure about:

1) https://github.com/kubernetes-sigs/dra-driver-cpu/blob/a24db9180ba6c11202d774da9995bb6e7c073266/pkg/driver/driver.go#L59
2) https://github.com/kubernetes-sigs/dra-driver-cpu/blob/a24db9180ba6c11202d774da9995bb6e7c073266/pkg/driver/cdi.go#L31

I'm not sure about the usage of those paths - but they're related to talking to CDI / kubelet rather than reading device information.

Should we also be using HOST_ROOT for those? If we did we'd also have to change how they're mounted in the container - so I think ignoring them is the right approach.

pkg/sysfs/sysfs.go
```diff
@@ -18,13 +18,18 @@ package sysfs
...
-)
+const Root = "/sys"
 
 type FS interface {
 	fs.ReadLinkFS
 	fs.ReadDirFS
 }
+
+// Host returns the host sysfs, honoring HOST_ROOT when set.
+func Host() FS {
```
AutuSnow, line 33: This is a nice cleanup overall — much better than the old scattered GetEnv calls. One small thing I found myself wondering about: Host() and ProcfsRoot() each read os.Getenv("HOST_ROOT") independently, so the "host root" concept now lives in two packages instead of one.
Not a blocker at all — probably a reasonable trade-off given the package boundaries. Was a single shared source (e.g. a small hostfs that exposes both the sys and proc roots) considered and rejected, or is splitting them just the pragmatic choice here? Happy either way, just curious about the intent.

cmd/dracpu/app.go
```diff
@@ -228,8 +227,7 @@ func run(logger logr.Logger) error {
 }
 
 func newSysFS(logger logr.Logger, overlayPath string) (sysfs.FS, error) {
```
AutuSnow, line 229: Moving these tests down into pkg/sysfs and pkg/cpuinfo is the right call — they exercise the real logic without the cmd wrapper, agreed. The one thing I noticed is that newSysFS itself no longer has any test touching it.
It's admittedly a one-liner now (NewOverlayFromFile(sysfs.Host(), path)), so I don't feel strongly — just checking whether dropping the last regression net at the cmd layer was deliberate. If you'd rather keep a tiny smoke test there it'd be cheap, but I'm equally happy to leave it as is.
﻿

tools/gen-pcie-testdata/main.go
```diff
@@ -42,10 +43,10 @@ const (
 )
 
 func main() {
-	sysfsRoot := sysfs.Root
+	var sysfsRoot string
 	gen := newMapFSGen()
 
-	flag.StringVar(&sysfsRoot, "sysfs-root", sysfsRoot, "sysfs root path")
+	flag.StringVar(&sysfsRoot, "sysfs-root", sysfsRoot, "sysfs root path (defaults to HOST_ROOT/sys)")
```
AutuSnow, line 49: just want to make sure I'm reading this right: these two used to read /sys and /proc literally, and now they go through sysfs.Host() / ProcfsRoot(), which means the default path changes when HOST_ROOT is set. For the in-container case (HOST_ROOT unset) it's identical, so I think we're fine.
The only scenario I'm slightly unsure about is someone running gen-pcie-testdata in an environment that has HOST_ROOT exported, without passing -sysfs-root — they'd now silently read a different root than before. The flag help update covers it, so this looks intentional; mostly flagging it so it's a conscious choice rather than a surprise. WDYT?

pkg/cpuinfo/cpuinfo.go
```diff
@@ -32,6 +32,13 @@ import (
 	"k8s.io/utils/cpuset"
 )
 
+const procfsRoot = "/proc"
```
AutuSnow, line 35: Here too

pkg/sysfs/sysfs.go
killianmuldoon: Mostly because of the order in which it happened to be honest - the sysfs changes came first and proc was bolted on. There's a well ordered sysfs package here that does a few things - mostly handling the overlay. We get nice handles like `sysfs.NewOverlayFromFile` this way. 

The `ProcfsRoot()` is just a single function so it seemed to slot nicely into cpuinfo. I don't feel too strongly about this though - so I can change the structure if you'd prefer.

cmd/dracpu/app.go
killianmuldoon: I think I'm fine with not including a test for `newSysFS` as it's a small helper for logging really.

The idea in general was to remove testing from the main package. We could have something like sysfs.New() in the sysfs package?

tools/gen-pcie-testdata/main.go
killianmuldoon: Yeah - I should have mentioned this one here too: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/227#pullrequestreview-4635507021

I'm not 100 percent sure this is the behaviour we want - would like some guidance if you have a strong instinct.

pkg/cpuinfo/cpuinfo.go
killianmuldoon: Not sure what this refers to?

pkg/sysfs/sysfs.go
ffromani: a bit historical context :)

procfs went first in the project history (EDIT: but Killian is correct in the sense that we improved sysfs first, and now indeed we're improving procfs after we fixed sysfs); sysfs came later, and we spent more polish on it. Now we're trying to apply the same or comparable polish level to proc, which we need less overall. Arguably, sysfs code was more polished because came second and we had a reference point and some lessons.

Personally, rather than polish more procfs support, or introduce another abstraction like hostfs, albeit tiny and probably useful, I'd rather try to _remove it_ and see if we can depend only on sysfs. That, of course, would be another followup. For this PR, I think it's more than enough to have a consistent handling of HOST_ROOT; I'm personally fine with this little duplication.

ffromani (commented):
I'm generally fine with the direction and the shape of this PR. I think as followup cleanup is fine, therefore provisional LGTM. I'll comment on the inline threads later.

cmd/dracpu/app.go
AutuSnow: Agreed — for a one-liner that's really just NewOverlayFromFile(sysfs.Host(), path) I don't think a cmd-layer test earns its keep, and moving the coverage down into pkg/sysfs / pkg/cpuinfo is the improvement I wanted to see.
On sysfs.New(): I like it as a design signal (a named constructor documents intent and gives the overlay-wiring a home in the package it belongs to), but only if it removes duplication or gets exercised by a real test — not solely to host a smoke test for newSysFS. I'd leave it out of this PR and consider it if a second caller shows up. No change needed here.

pkg/cpuinfo/cpuinfo.go
AutuSnow: Sorry, that was a terse drive-by from me — "here too" just meant the same observation I raised on sysfs.go:33 (the independent os.Getenv("HOST_ROOT") read) also applies to ProcfsRoot(). It's fully answered by the thread over there (ffromani's "we'd rather remove procfs than abstract it"), so nothing to do here. Resolving.

tools/gen-pcie-testdata/main.go
AutuSnow: Here's my instinct. gen-pcie-testdata is a developer codegen tool, normally run on a workstation, not inside the driver container — so HOST_ROOT semantics don't really "belong" to it the way they do to the driver. Honoring it by default is the mildly surprising bit.
That said, I don't think it's worth diverging from the rest of the codebase. My recommendation: keep sysfs.Host() / ProcfsRoot() as the default (one mental model everywhere), keep -sysfs-root as the authoritative override, and just remove the "silent" part — log the resolved sysfs/proc root once at startup (you already silence stdout on the happy path, so send it to stderr). That way a dev who happens to have HOST_ROOT exported sees exactly which tree was read instead of getting a surprising fixture.
If you'd rather not honor the env at all here, the alternative is to have the tool default to literal /sys and /proc and only consult HOST_ROOT when -sysfs-root/-proc-root is explicitly passed — but I think that's over-thinking it for a dev tool. I'd go with the "keep consistent + log the resolved root" option.

killianmuldoon: I added a log here to make this clearer to the user - let me know if you think it's good enough.

ffromani (commented):
/lgtm

@AutuSnow are you happy with the current PR shape? When you are, I think we can merge

test/image/dracputester/app.go
```diff
@@ -88,7 +90,7 @@ func affinityScanBoundFromTopology(topo *cpuinfo.CPUTopology) int {
 func main() {
 	logger := stdr.New(log.Default())
 	for {
-		cpus, err := cpuSet("/sys")
+		cpus, err := cpuSet(sysfs.Host())
```
AutuSnow: Small thing I wanted to double-check here — should the tester's own cpuset really be read through sysfs.Host()?
﻿
My understanding is that cpuSet() reads fs/cgroup/cpuset.cpus.effective to find out which CPUs this container was actually allocated, so it needs the container-local /sys/fs/cgroup view. But sysfs.Host() honors HOST_ROOT, which is meant for reading the host's sysfs.
﻿
Today this is harmless — nothing sets HOST_ROOT on the tester pod, so Host() resolves to /sys and everything works (and the e2e jobs confirm it). But I'm a little unsure about the long term: if HOST_ROOT ever gets set on this pod, wouldn't we end up reading the host root cgroup instead of the container's own cpuset, and silently get the wrong allocation in the assertions?
﻿
Would it make sense to keep this one explicitly container-local (e.g. the previous os.DirFS("/sys"), or a dedicated helper) with a short comment noting it intentionally does not honor HOST_ROOT? Happy to be corrected if I'm misreading how the tester gets deployed.

killianmuldoon: You're dead right here - I updated this so cputester only sees the real cpuset and ignore HOST_ROOT.

ffromani: Good catch @AutuSnow !
Taking this chance to reconsider, the solution as-is in the PR is probably the best approach indeed.

EDIT: the following was meant to record the design direction, but feels misplaced here. Let me move it to a top-level comment. Leaving it here for historical reference.
~my 2c about the general matter, now that is settled: I thought about this and I was on the fence. Using HOST_ROOT everywhere is easy and consistent. Avoiding HOST_ROOT in dracputester and hardcoding (= avoid possible override) of the sysfs path is more correct - there's not a usecase for using it HOST_ROOT here.
But then we have a (arguably more correct) split between what the production code in the driver uses and what the external tooling uses. Is not clear how to explain and enforce who uses what except the docs.
Clearing up the situation fully could likely require more iteration, which doesn't belong in this PR.~

ffromani:
Note for future (xref: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/227#discussion_r3548923576): we now intentionally split HOST_ROOT usage between production code (honors it) and test/dev tooling (ignores it).

non-production driver code we have is support tooling code. Important, but there is no obvious need or reason for it to support sysfs relocation/non-standard sysfs mounts; therefore we prefer to hardcode full paths/do direct access (= not mediated by golang's `fs` or our `pkg/sysfs`)

test/image/dracputester/app.go
AutuSnow: Thanks @killianmuldoon — the latest commit (explicitly avoid HOST_ROOT in cputester) covers the last thing I was unsure about. Reading the container's own cgroup cpuset through an explicit os.DirFS("/sys") with the intent comment is exactly what I was hoping for; keeping that path independent of HOST_ROOT avoids a future foot-gun where a stray HOST_ROOT could silently point the tester at the host's cgroup view. The rest of the consolidation reads cleanly and all the e2e modes are green.

AutuSnow:
/lgtm
Thanks ！！

ffromani (approved):
/approve

thanks for the contribution and for the accurate review

PR #228, opened 2026-07-06
test: Add e2e testing based on sysfs overlay

killianmuldoon:
#### What type of PR is this?

/kind cleanup 

#225  added a way to overlay files on top of local sysfs, allowing device topology and other attributes to be mocked. 

This set of changes leverages that in an e2e test - giving developers a clear way to test the behaviour of the DRA driver with mocked hardware under the hood.

There are also a couple of cleanup commits that enable the e2e test suite to work on Docker Desktop for macOS, and removes some unneeded e2e specific code.

#### Release note

```release-note

```


test/image/dracpuinfo/app.go
```diff
@@ -1,49 +0,0 @@
-/*
-Copyright 2025 The Kubernetes Authors.
```
killianmuldoon, line 2: I can move this change to a separate PR if that's better. It became clear while I was working on the test that dropping this would be ~easy.

Makefile
```diff
@@ -26,7 +26,7 @@ YQ_VERSION ?= 4.47.1
 GOLANGCI_LINT_VERSION ?= 2.12.2
 HELM_DOCS_VERSION ?= 1.14.2
 HELM_SCHEMA_VERSION ?= 2.3.1
-KIND_K8S_VERSION ?= v1.36.0
+KIND_K8S_VERSION ?= v1.36.1
```
killianmuldoon, line 29: This enables us to pull from docker hub. Previously - at least for me - I had to build kindest/node for 1.36.0 as it's not available on docker hub.

test/e2e/nri_reconciliation_test.go
```diff
@@ -207,11 +193,14 @@ var _ = ginkgo.Describe("NRI Reconciliation on Restart", ginkgo.Serial, ginkgo.O
...
 
-		ginkgo.By("Verifying Pod 2 CPU mask is NOT restricted to shared pool (should be default/all)")
+		ginkgo.By("Verifying Pod 2 CPU mask is NOT restricted to the shared pool")
 		alloc2 := getTesterPodCPUAllocation(fxt.K8SClientset, ctx, createdPod2)
 		fxt.Log.Info("Pod 2 CPU allocation (without NRI)", "cpuAssigned", alloc2.CPUAssigned.String())
-		// Since NRI is down, pod2 is not restricted to shared pool CPUs and gets all online CPUs.
-		gomega.Expect(alloc2.CPUAffinity).To(cpusetmatchers.Equal(allocatableCPUs), "Pod 2 CPU mask not equal to all CPUs")
+		// Since NRI is down, pod2 is not restricted to shared pool CPUs. Its native affinity may
+		// include CPUs outside the driver's overlaid topology, so require the driver CPUs and the
+		// exclusively allocated CPUs to be subsets instead of requiring exact equality.
```
killianmuldoon: This was a bit tricky - but I think the result is sufficient. If checking the sets doesn't seem clear enough I can have another go at tackling this.

Makefile
```diff
@@ -211,14 +218,20 @@ build-test-image: ## build tests image
...
+		DRACPU_E2E_EXPECTED_NUM_CPUS=$(DRACPU_E2E_EXPECTED_NUM_CPUS) \
+		DRACPU_E2E_EXPECTED_NUMA_NODES=$(DRACPU_E2E_EXPECTED_NUMA_NODES) \
+		go test -v ./test/e2e/ --ginkgo.v
 
 test-e2e-kind: ci-kind-setup test-e2e ## run e2e test against a purpose-built kind cluster
 
+test-e2e-kind-overlay: DRACPU_E2E_SYSFS_OVERLAY=true
+test-e2e-kind-overlay: DRACPU_E2E_EXPECTED_NUM_CPUS=4
+test-e2e-kind-overlay: DRACPU_E2E_EXPECTED_NUMA_NODES=2
+test-e2e-kind-overlay: ci-kind-setup test-e2e ## run e2e test against a kind cluster using the example sysfs overlay
```
killianmuldoon, line 233: This isn't invoked by CI in this MR - It would be good to understand if that's desirable at this point. I'm not sure if this flow is really just dev-only or whether it will be useful to test different hardware configs going forward.

ffromani:
DISCLOSURE: sharing thoughts based on the PR description only.
NOTE: I wanted to record these thoughts since quite long time, so I'm taking the chance here. Some or most of the following _may not apply_ to the work actually done here. I'll post a PR to consolidate this direction in the README.md later.

I'm all to improve the devex, enable testing, allow us to test more of the codebase and more often and enable users on MacOS/other combos than linux/amd64. The line I'm careful not to cross, however, is to _depend_ on faking.

I believe the highest value coming from e2e tests is the ability to run them on real hardware and real clusters, and, in general, to be anchored to real, existing HW. It's way too easy to introduce good-lokoking fakes which slowly make our code drift from the real hardware; and if we always (or most often) test on fake hardware, that drift will only materialize when we try things in production, too late.

There's real value in faking complex hw topologies or hw in our tests. That coverage would be very valuable and would make us more robust: the only alternative would be borrowing HW (unlikely) or best-effort runs on HW us devs we have available (unreliable) or adding expensive HW to CI (very unlikely).

Faking, however, should be clearly insulated and must be opt-out somehow. We should have e2e tests which can run anytime on real HW, we should keep running as many tests as possible on real HW, and we we need to have a clear on/off mechanism to fake HW we don't have readily available.

This way the e2e suite stays the same, and I can run it anytime on a mix of real/fake HW (ranging from 0 to 100 in both directions) depending on what I have handy.

[EDIT] wanted to take the chance to write down these thoughts - something which I was ruminating about and postponing for too long. I'll do a proper review ASAP, referring to these principles. Thanks for your PR, this is a very important area we need to start addressing sooner than later.

killianmuldoon:
>I'm all to improve the devex, enable testing, allow us to test more of the codebase and more often and enable users on MacOS/other combos than linux/amd64. The line I'm careful not to cross, however, is to depend on faking.

Completely agree - this PR adds a separate make target which leverages the overlay and doesn't touch the existing e2e testing. It's a starting off point - potentially at least - for being able to test against mocked hardware both locally and in CI if that's desirable.

test/e2e_local/sysfs_overlay_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"sigs.k8s.io/yaml"
+)
+
+type gatherInfoReport struct {
```
AutuSnow: Just curious about the choice here: the cluster-side test (test/e2e/sysfs_overlay_test.go) unmarshals straight into gatherinfo.Report, but this local test hand-rolls its own gatherInfoReport mirror. Since both live in the same module, was there a reason not to reuse the real type?
My worry is drift — if the Report JSON tags ever change, this copy won't fail to compile, it'll just silently decode into zero values and the assertions might pass for the wrong reason. If it's intentional (e.g. pinning the JSON contract from an "external consumer" viewpoint), a one-line comment saying so would save the next reader from thinking it's an oversight. Otherwise, reusing gatherinfo.Report seems safer. WDYT?

test/e2e/healthz_test.go
```diff
@@ -74,42 +44,26 @@ var _ = ginkgo.Describe("dra-driver-cpu HTTP health endpoints", ginkgo.Ordered,
...
+				ginkgo.By(fmt.Sprintf("GET %s through the pod proxy (pod %s, node %s)",
+					healthzPath, pod.Name, pod.Spec.NodeName))
 
-				var lastStatus int
-				var lastBody string
 				gomega.Eventually(func(g gomega.Gomega) {
-					resp, err := httpClient.Get(url) //nolint:noctx // httpClient.Timeout covers this
+					_, err := fxt.K8SClientset.CoreV1().Pods(daemonSetNamespace).
+						ProxyGet("http", pod.Name, strconv.Itoa(driverHTTPPort), healthzPath, nil).
+						DoRaw(ctx)
```
AutuSnow, line 60: One thing I wanted to double-check: the old version asserted resp.StatusCode == http.StatusOK and surfaced the body on failure, whereas the proxy version only checks that err is nil. I think DoRaw turns any status ≥ 400 into an error, so a 500 would still be caught — is that the assumption you're relying on?
If so, no objection to keeping it. The only thing we lose is the "exactly 200" precision and the raw-body diagnostics when something goes sideways. Might be worth keeping the body in the failure message just to make future debugging less painful — but I'll leave that to your judgement.

test/e2e/sysfs_overlay_test.go
```diff
@@ -0,0 +1,109 @@
...
+		gomega.Expect(overlayPath).NotTo(gomega.BeEmpty(),
+			"overlay topology expectations are set, but the driver has no %s argument", argSysFSOverlay)
+
+		restConfig, err = e2eclient.NewK8SConfig()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot create Kubernetes config")
+	})
+
+	ginkgo.It("should expose the expected topology through each driver pod", func(ctx context.Context) {
+		for _, pod := range waitForRunningDriverPods(ctx, fxt.K8SClientset) {
+			readyPod := waitForReadyDriverPod(ctx, fxt.K8SClientset, pod.Spec.NodeName)
```
AutuSnow, line 83: nit: take it or leave it: inside the loop we already have the pod from waitForRunningDriverPods, then waitForReadyDriverPod(pod.Spec.NodeName) goes and re-fetches it via GetDRACPUPod. Functionally fine, just an extra round-trip per pod. Not worth churning the code over unless you're touching this block anyway.

test/e2e/nri_reconciliation_test.go
AutuSnow: The switch from Equal(allocatableCPUs) to the subset checks makes sense to me for the overlay case (host has more physical CPUs than the overlaid topology, so exact equality can't hold). The comment explains it well.
The one thing I'm mildly wondering about: for the non-overlay run we've now given up the strict "equals all CPUs" guarantee, so a regression where pod2 ends up with an unexpectedly wider affinity wouldn't be caught anymore. Probably an acceptable trade-off for overlay compatibility — just flagging it so it's a conscious decision rather than a side effect. No change needed if you're comfortable with that.

test/e2e/healthz_test.go
killianmuldoon: AFAICT `DoRaw` accepts only 200-206 as successes

https://github.com/kubernetes/client-go/blob/c0f38c0636ddecab3c6b32ef4ecdcf1cf884df3f/rest/request.go#L1161 

That behaviour seems alright - but I'll add the body in the failure message. It could definitely come in useful.

test/e2e_local/sysfs_overlay_test.go
killianmuldoon: This was a mistake I made when moving things around 😅  - thanks for catching it. I'll import `internal/gatherinfo`

test/e2e/sysfs_overlay_test.go
killianmuldoon: It's a tricky one - when we get the pods initially we only check that they're Running, not that they're Ready. IMO it's better just to get them again as this is test code.

One optimisation - if we wanted - would be to use a controller runtime caching client in these tests - that would save the round trip in ~all cases, but I'm not sure if it's worth it.

test/e2e/nri_reconciliation_test.go
killianmuldoon: Yeah - it's not nice to lose exactness here. An easy but ugly option would be to check conditionally based on whether the overlay is being used - I really don't like that though. 

I'll play around with it and see if I can come up with something better.

killianmuldoon: One potential solution would be to grab the set of "real" CPUs in the cpu tester at start time and then use that as the comparison in this check. It sort of muddles what we're looking for here - but it keeps the test as accurate as previously.

test/e2e/healthz_test.go
AutuSnow: Perfect, that link settles it — 200–206 is the right success window and /healthz returns 200, so we're covered. Thanks for adding the body to the failure message too; that'll make the next flake much easier to diagnose.

test/e2e/sysfs_overlay_test.go
AutuSnow: Ah, the Running-vs-Ready distinction is a fair reason — the re-fetch is buying correctness, not just repetition, so I'm happy to leave it. Agree the caching client is more machinery than this is worth right now; not worth churning for a test-only round trip. Resolving.

test/e2e/nri_reconciliation_test.go
AutuSnow: I'd vote for the second idea (capturing the tester's real CPU set at start time), and agree the overlay-conditional branch is best avoided — that path tends to rot because the two branches drift.
One reframing that might make (b) feel less like it muddles the intent: what this assertion really wants to prove is "with NRI down, pod2 is fully unrestricted — it sees the node's native cpuset". The node's real online CPUs are the correct reference for "unrestricted"; comparing against the driver's allocatableCPUs was actually slightly wrong under overlay (the driver's view is a subset of the real machine). So switching the comparison to the tester's real CPUs restores exactness and arguably states the invariant more honestly than the old Equal(allocatableCPUs) did.
If it helps keep it readable: the subset checks answer "did pod2 lose access to anything it should have?", while Equal(realCPUs) answers "is pod2 truly unrestricted?" — keeping the latter as the primary assertion (minus reserved CPUs, if those apply here) looks like the cleaner spot to land. Happy to review whatever you come up with.

killianmuldoon: My latest commit adds this. It's not too confusing - but I added a few comments to try to explain it to myself.

PR #229, opened 2026-07-07
feat: external allocator in grouped device mode

ffromani:
#### What type of PR is this?

/kind feature

Support external allocator (SLURM) in grouped device mode alongside in group-by=machine

N/A

gen AI usage disclosure: AI was used for self-review and to assist with rebases and commit reshaping.

#### Release note

```release-note
The grouped device mode now support external allocator hints controlled by a driver flag
```

#### Documentation updates

- README
- e2e README
- upgrade instructions


ffromani:
xref: #164 #217

ffromani:
depends on a still-on-polish cleanup PR which I'll post independently later on

ffromani:
/cc @killianmuldoon re: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/217#issuecomment-4901549809

please don't spend too much time on this rough sketch. The last 3 commits are a quite crude demo of what I have in mind. Need to finalize and post my refactoring PR to clean the board before I can polish that concept.

ffromani:
I believe this approach has legs. Problem: it greatly improves on the problem we had on #217 with `DeviceAttributeMaxValueLength` limit, but it doesn't address it. 

But we have options without resorting to bump the limit, or fall back to individual mode:
a quick but realistic scan of the existing and planned general available hardware on the market  makes us reasonnably confident we won't hit this problem in practice (this argument also fuels back in #217)

How the worst case would look like? A full non-contiguous enumeration: firmware assigns even CPU IDs to one socket and odd to the other. "0,2,4,...,510" for a 256-thread socket is ~1400 chars, so goodbye to range compression.
(NUMA grouping has the exact identical concern). How likely this is on shipped or upcoming hardware I can't say; it seems very weird and I've only seen myself on old nehalem-old hardware, but the sample size is too tiny to gain any confidence.

So, if this becomes a real problem, we can *extend* the linux cpuset notation with more compact encodings.
Obviously, our extended cpuset MUST be a proper superset of the standard existing and universal cpuset notation.
Going down this path, we can add a stride concept.  The standard N-M notation means stride=1 (contiguous). A new form N>M>S would express arbitrary strides, e.g. 0>510>2

The two forms compose freely in a single string: 0-15,32>510>2,512-527.
This would be published as a separate attribute (`dra.cpu/cpusetExt or similar`) since it's not standard cpuset syntax.

killianmuldoon:
Thanks for this - I'll put time into reviewing and testing this approach tomorrow - but from a high level it really looks promising. I think it's not feasible to cover all possible CPU/NUMA combinations at this point - so as long as the representation isn't close to the limits that seems reasonable.

TIL about strided ranges - thanks! 


Maybe we should use the [taskset convention](https://github.com/util-linux/util-linux/blob/89f905b0e36570a927bf002f6b7e04b0cf2e72ac/schedutils/taskset.1.adoc?plain=1#L65) - e.g. `0-510:2`?

ffromani:
> Maybe we should use the [taskset convention](https://github.com/util-linux/util-linux/blob/89f905b0e36570a927bf002f6b7e04b0cf2e72ac/schedutils/taskset.1.adoc?plain=1#L65) - e.g. `0-510:2`?

Even better! TIL about taskset new syntax, will have a look

ffromani:
> Thanks for this - I'll put time into reviewing and testing this approach tomorrow - but from a high level it really looks promising. I think it's not feasible to cover all possible CPU/NUMA combinations at this point - so as long as the representation isn't close to the limits that seems reasonable.

Note about the code: basically it compiles and passes the tests, expect rough edges. If you think the direction is promising I'll surely polish this up.

Note: I don't think we should remove machine grouping - not yet at least, and as part of the PR finalization I plan to bring it up to speed. ATM I think we need both.

pkg/cpumanager/shim.go
```diff
@@ -0,0 +1,59 @@
...
+		if config.Opaque == nil || config.Opaque.Driver != alc.driverName {
+			continue
+		}
+		return cpuset.New(), fmt.Errorf("opaque device configuration is not supported by the %q allocator; use the external allocator to provide custom cpuset assignments", alc.driverName)
+	}
+	return cpuset.New(), nil
+}
+
+func (alc *Allocator) Validate(allocatedCPUs, opaqueCPUSet, cpuAssignment cpuset.CPUSet, count int) error {
+	return nil // we trust the cpumanager code
```
AutuSnow: Really like where this POC is going. One thing I want to make sure I'm reading correctly, though: in the GROUP_BY_MACHINE branch we always pull the cpuset via extalloc.GetOpaqueCPUSet(...) and set cur = opaqueCPUSet, but the validation now goes through cp.cpuAllocator.Validate(...) — and since --allocator defaults to cpumanager, that Validate is return nil // we trust the cpumanager code.
If I'm not missing something, that means the online / reserved / size / overlap checks the old validateOpaqueCPUSet used to enforce get skipped for group-by=machine whenever --allocator isn't explicitly external — i.e. on the default path. That would let an out-of-range opaque cpuset (offline, reserved, wrong count, or conflicting) land on the container cgroup, which feels like a regression from current main. Is machine mode meant to imply the external allocator, or is the intent for both allocators to still share those safety checks?

ffromani: Thanks for the early review and for pointing this out. In machine mode we essentially imply allocator=external. Failing to validate like we do now in main would be a regression indeed, and that must be avoided. The options I see now are: 1. force validation like before using an explicit path 2. force `allocator=external` if `group-by=machine`. I tend to prefer the former, but haven't made my mind yet.
Note: obviously we will need to make sure we have test coverage to catch this flow.

killianmuldoon:
First look looks good - here's the slice from a GB200:

```
apiVersion: resource.k8s.io/v1
kind: ResourceSlice
metadata:
  creationTimestamp: "2026-07-08T09:35:52Z"
  generateName: 00000-dra.cpu-validation-k8s-gpu-03-
  generation: 1
  name: 00000-dra.cpu-validation-k8s-gpu-03-s256l
  ownerReferences:
  - apiVersion: v1
    controller: true
    kind: Node
    name: validation-k8s-gpu-03
    uid: bf733fdd-91a5-4636-bd3c-ad42c61f843f
  resourceVersion: "37864370"
  uid: b3474dfd-86e6-4435-9e8c-16ce666e4139
spec:
  devices:
  - attributes:
      dra.cpu/cpuIDs:
        string: 0-71
      dra.cpu/numCPUs:
        int: 72
      dra.cpu/numaNodeID:
        int: 0
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 268435456
      dra.net/numaNode:
        int: 0
    capacity:
      dra.cpu/cpu:
        value: "72"
    name: cpudevnuma000
  - attributes:
      dra.cpu/cpuIDs:
        string: 72-143
      dra.cpu/numCPUs:
        int: 72
      dra.cpu/numaNodeID:
        int: 1
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 285212672
      dra.net/numaNode:
        int: 1
    capacity:
      dra.cpu/cpu:
        value: "72"
    name: cpudevnuma001
  driver: dra.cpu
  nodeName: validation-k8s-gpu-03
  pool:
    generation: 1
    name: validation-k8s-gpu-03
    resourceSliceCount: 1
```

ffromani:
> First look looks good - here's the slice from a GB200:
> 
> ```
>       dra.cpu/socketID:
>         int: 268435456
>       dra.net/numaNode:
>         int: 0
>       dra.cpu/socketID:
>         int: 285212672
>       dra.net/numaNode:
>         int: 1
> ```

just curious about the unexpected socketIDs, will try to investigate (with low priority) later on

killianmuldoon:
> just curious about the unexpected socketIDs, will try to investigate (with low priority) later on

These are correct in the sense that they're the physical_package_id on my system:

```
 cat /sys/devices/system/cpu/cpu0/topology/physical_package_id
268435456
```

ffromani:
**NOTE** rebased naively on top of the config file support. Yet to be decided if we want to expose this tunable as flag or if it should be the first config-file only flag. Will decide as part as the finalization PR.

pkg/cpumanager/shim.go
AutuSnow: Following up now that this landed on forcing external for machine mode: the cpumanager + machine combo gets silently swapped instead of rejected. Now that Config.Validate() exists, wouldn't erroring there on the incompatible pair be cleaner? Quietly changing the allocator under the user feels worse than a clear config error. Or are you still leaning toward option 1 (keep both allocators and force the validation path)?

AutuSnow: One gap I don't see covered: a bad opaque cpuset in machine mode (offline, reserved, wrong count, overlap) actually getting rejected. The happy path is there, but since the safety all sits in external's Validate() now, a negative test would catch a future regression. Can add one if it helps.

pkg/driver/driver.go
```diff
@@ -72,6 +75,12 @@ type CPUEnumerator interface {
 	CreateDevices(logger logr.Logger) []resourceapi.Device
 }
 
+type CPUAllocator interface {
```
AutuSnow, line 88: Minor: could we spell out the safety contract on this interface? cpumanager does the checks in Allocate and leaves Validate a no-op, extalloc is the reverse, and nothing here says Validate is mandatory at the call site. Pretty easy to break by accident later.

ffromani:
@AutuSnow thanks for the early feedback. This PR is still WIP and we have many rough edges or unfinished parts. I'm waiting for @killianmuldoon / @catblade  's feedback to verify it serves the purposes to fill the missing gaps for external scheduler (slinky) integration before I will continue the work and finish the impl.

I'll make sure to address your points first!

the rebase was meant to signal the POC is still active and we have plans for it, it was not just a one-off experimet (fire and forget). In hindsight, I should probably have made that explicit earlier. Better late than never I guess?

AutuSnow:
No worries, the rebase read as "still alive" to me. Waiting on killian/catblade before finishing makes sense, that's the right order. My notes were just early flags, nothing blocking, so hit them whenever suits you.

killianmuldoon:
From the slurm-bridge side I vote we go ahead and pursue this. It really seems to fit the slurm model well. I've been playing with this version +/- a couple of fixes and it works well. @catblade is OOO for a few days - but agrees with the direction.

Let me know if there's anything I can help with to push this forward.

ffromani:
thanks @killianmuldoon . I can go ahead and finish this PR, then you and @catblade can have the final ack/nack if it fits Slurm needs.  I'll resume this work in the next few days, when the kube 1.37 frenzy winds down (because the code freeze deadline!)

killianmuldoon:
Sounds great - I'll review this once it's out of draft. Thanks again for working on it!

ffromani:
reviewers this is highly WIP (and likely broken now) while I find the final internal shape, please ignore till the WIP is lifted

pkg/driver/driver.go
ffromani: good point, addressing as code comments.

pkg/cpumanager/shim.go
ffromani: Thanks for raising these 2 good points. Will add stricter validation and the tests before to un-draft.

ffromani:
I have resolutions for all the outstanding issues. I'm heading for some OOO time (~1 week), I'll finalize and un-draft this work once back. Rebase is partial and there are known issues (hence WIP)

ffromani:
I finally found an approach I'm happy with regarding the string attributes potentially exceeding limits.

The "strided range" cpuset support is not yet implemented, intentionally. But if we ever need it, there's some code here: https://github.com/ffromani/cpuset/commit/c781d7b7d84a1a967ec002c8d32ee735d066e042

pkg/cpumanager/shim.go
ffromani: I think I addressed the points in my latest commit. If CI agrees, will undraft soon.

ffromani:
ready for review!

killianmuldoon:
Nice! I'm OOO for the next couple of days but will put it top of my list for next week.

ffromani:
let me narrow down this PR to its core. I'll post the other improvements (e.g. device.Builder improvements, or attempts to) as followups.

[EDIT]
followup 1 (for logistical reasons): `pkg/device/builder.go` cleanup/refactoring
followup 2: expose cpuID attribute in grouped mode when `allocator=external`

both however will need to wait for this PR to land, to keep each PR focused. Nothe both of them were previously bundled in this PR

ffromani:
self-review notes:
- extalloc.go can use some unit tests, will work on them in parallel
- the branch is not really commit-by-commit safe. The commit `feat: allocator: add and wire external allocator` introduces a validation regression, which *is fixed* right afterwards by the following commit, but still.

killianmuldoon:
I've been playing around with this and it looks basically good. I'm working on a PoC of an integration with slurm-bridge so I can test our use case more clearly.

One error I've run into so far. With a ResourceClaim that would require multiple results per request I'm seeing a failure.

The pod gets scheduled but fails at ContainerCreating with `Failed to prepare dynamic resources: prepare dynamic resources: NodePrepareResources failed for ResourceClaim opaque-two-numa-cpus: preferred CPUs cpuset size 2 does not match requested capacity 1`

I think this is fine, but it would be good to document it somewhere, or maybe return a more specific error. 

Here's my pod spec, resourceclaims and the resourceslice I'm using:

<details>

```yaml
apiVersion: resource.k8s.io/v1
kind: ResourceSlice
metadata:
  creationTimestamp: "2026-08-17T09:48:06Z"
  generateName: 00000-dra.cpu-dra-pr229-live-worker-
  generation: 1
  name: 00000-dra.cpu-dra-pr229-live-worker-sjrzq
  ownerReferences:
  - apiVersion: v1
    controller: true
    kind: Node
    name: dra-pr229-live-worker
    uid: 63d52662-6b66-4fcf-b8d4-d21b7f8211b7
  resourceVersion: "656"
  uid: bada9c88-0e9f-4008-9d71-32fd34380fca
spec:
  devices:
  - allowMultipleAllocations: true
    attributes:
      dra.cpu/numCPUs:
        int: 2
      dra.cpu/numaNodeID:
        int: 0
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 0
      dra.net/numaNode:
        int: 0
    capacity:
      dra.cpu/cpu:
        value: "2"
    name: cpudevnuma000
  - allowMultipleAllocations: true
    attributes:
      dra.cpu/numCPUs:
        int: 2
      dra.cpu/numaNodeID:
        int: 1
      dra.cpu/smtEnabled:
        bool: false
      dra.cpu/socketID:
        int: 1
      dra.net/numaNode:
        int: 1
    capacity:
      dra.cpu/cpu:
        value: "2"
    name: cpudevnuma001
  driver: dra.cpu
  nodeName: dra-pr229-live-worker
  pool:
    generation: 1
    name: dra-pr229-live-worker
    resourceSliceCount: 1

---

apiVersion: resource.k8s.io/v1
kind: ResourceClaim
metadata:
  name: opaque-two-numa-cpus
  namespace: pr229-repro
spec:
  devices:
    requests:
      - name: request-cpus
        exactly:
          deviceClassName: dra.cpu
          count: 2
          capacity:
            requests:
              dra.cpu/cpu: "1"
    config:
      - requests:
          - request-cpus
        opaque:
          driver: dra.cpu
          parameters:
            apiVersion: v1alpha1
            cpuConfig:
              cpuset: "0,2"
---
apiVersion: v1
kind: Pod
metadata:
  name: opaque-two-numa-test
  namespace: pr229-repro
spec:
  restartPolicy: Never
  nodeSelector:
    kubernetes.io/hostname: dra-pr229-live-worker2
  resourceClaims:
    - name: cpus
      resourceClaimName: opaque-two-numa-cpus
  containers:
    - name: workload
      image: busybox:1.36.1
      command:
        - /bin/sh
        - -c
      args:
        - "grep '^Cpus_allowed_list' /proc/self/status; sleep 3600"
      resources:
        requests:
          cpu: "2"
          memory: 32Mi
        limits:
          cpu: "2"
          memory: 32Mi
        claims:
          - name: cpus
```

</details>

ffromani:
thanks for the feedback @killianmuldoon I'll take care (documenting, improving, both) in the next upload alongside with more test coverage

ffromani:
> thanks for the feedback @killianmuldoon I'll take care (documenting, improving, both) in the next upload alongside with more test coverage

restoring WIP meanwhile

ffromani:
> I've been playing around with this and it looks basically good. I'm working on a PoC of an integration with slurm-bridge so I can test our use case more clearly.
> 
> One error I've run into so far. With a ResourceClaim that would require multiple results per request I'm seeing a failure.
> 
> The pod gets scheduled but fails at ContainerCreating with `Failed to prepare dynamic resources: prepare dynamic resources: NodePrepareResources failed for ResourceClaim opaque-two-numa-cpus: preferred CPUs cpuset size 2 does not match requested capacity 1`
> 
> I think this is fine, but it would be good to document it somewhere, or maybe return a more specific error.

This behavior stems from the fact the current code was a direct and intentional generalization of the basic machine mode logic. By "intentional" I mean I reshaped the old code without challenging the machine mode assumptions yet.

So we inherited a limitation which makes sense on machine mode: 1 request = 1 result.
Your ResourceClaim has `count: 2` with `capacity.requests.dra.cpu/cpu: "1"`, and the opaque config `cpuset: "0,2"` In DRA, count: 2 expands into two DeviceRequestAllocationResults, each with `ConsumedCapacity[dra.cpu/cpu] = 1`. Therefore a single request can lead to multiple allocation results.

The builtin/canonical mode enables this path: a claim can span across  the cross-NUMA/cross-socket boundary within a single request. A `count:N` request can expand into one result per grouped device, and because grouped devices are `allowMultipleAllocations=true`, the scheduler can satisfy one request by drawing from several group devices.

The opaque/external allocator mode should be (or start as) symmetric and equally powerful than canonical mode, and `group-by=machine` enabled this path. Rather than document the asymmetry, I'd rather lift it.

ffromani:
realized we need better docs for the grouped device mode w/ external allocator. Took some time to investigate and address some gaps in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/303 - now we have a better base to finish this work.

pkg/cpuallocator/kubelet.go
```diff
@@ -0,0 +1,38 @@
...
+package cpuallocator
+
+import (
+	"github.com/go-logr/logr"
+	topology "github.com/kubernetes-sigs/dra-driver-cpu/pkg/cpuinfo"
+	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/cpumanager"
+	"k8s.io/utils/cpuset"
+)
+
+type Kubelet struct {
```
pravk03: Maybe just `Kubelet` is under-specifying here ?

how about  `KubeletCPUManager` or `KubeletStaticCPUPolicy` ?

pkg/cpuallocator/external.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Verify cores do not overlap with other claims prepared in this same batch
+	currentClaimCPUs := opaqueCPUSet.Intersection(cpuAssignment)
+	if currentClaimCPUs.Size() > 0 {
+		return fmt.Errorf("requested CPUs %s from opaque config are already assigned to another device in this claim", opaqueCPUSet.String())
+	}
+
+	// Verify cores do not overlap with other active claims on this node
+	existingClaimCPUs := allocatedCPUs
```
pravk03: nit: we can avoid this copy ?

pkg/cpuallocator/kubelet.go
```diff
@@ -17,22 +17,63 @@ limitations under the License.
...
+	if err != nil {
+		return cpuset.New(), fmt.Errorf("allocating %d CPUs from the priority set %v: %w", priorityCPUs.Size(), priorityCPUs, err)
+	}
+	remainingCPUs := availableCPUs.Difference(alloc)
+	restCount := count - alloc.Size()
+	rest, err := cpumanager.TakeByTopologyNUMAPacked(logger, alc.topo, remainingCPUs, restCount, cpumanager.CPUSortingStrategyPacked, true)
+	if err != nil {
+		return cpuset.New(), fmt.Errorf("allocating %d CPUs from the remaining set %v: %w", restCount, remainingCPUs, err)
+	}
+	return alloc.Union(rest), nil
```
pravk03: Since `GetPreferredCPUs()` always returns empty cpuset (or ErrUnsupportedPreferredCPUs),  the else block (preferredCPUs.Size() != 0) is unreachable. Should we add support for preferredCPUs in both these functions simultaneously ?

pkg/driver/driver.go
```diff
@@ -461,3 +471,15 @@ func generateShortID(length int) string {
...
 }
+
+func findAllocatorMode(logger logr.Logger, config *Config) string {
+	if config.Allocator == driverconfig.AllocatorExternal {
+		// easy case: explicit user preference
+		return config.Allocator
+	}
+	if config.CPUDeviceMode == device.CPU_DEVICE_MODE_GROUPED && config.CPUDeviceGroupBy == device.GROUP_BY_MACHINE {
+		logger.Info("machine grouping in grouped device mode requires external allocator, forcing")
+		return driverconfig.AllocatorExternal
```
pravk03: Should we fail loudly if Allocator and CPUDeviceGroupBy are not compatible ?

pkg/driver/dra_hooks.go
```diff
@@ -145,12 +144,14 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		topo := cp.topology.cpuTopology
 		// TODO: what if `claimCPUCount==0`?
 
-		var cur cpuset.CPUSet
+		// The preferred hint comes from the request's opaque config (if any); the
```
pravk03, line 165: nit: I think this can be moved before this for loop.

pkg/cpuallocator/kubelet.go
ffromani: Thanks for the early (because PR still WIP!) review. Yes, struggling with naming. I'm probably reverting the name to `CPUManager`. I'm open to suggestion here, my only preference would be to prefer concise names.

pkg/cpuallocator/external.go
ffromani: most likely yes. Let me try harder, or clearly document why not.

pkg/cpuallocator/kubelet.go
ffromani: Thanks for raising this point. I was wrestling myself with this idea because the code we had before was much simpler but technically not compliant to the interface. I wanted to see how it looked but went overboard. I'll simplify the code back, and add comments.

pkg/driver/driver.go
ffromani: Indeed we can probably start failing loudly if the configuration is not compatible. Good point.

ffromani:
addressed and incorporated the multi-result case. I did some more cleanup and added tests. There it was some churn but, except for the changes necessary to handle multi-result, no real substantial difference.

ffromani:
requires https://github.com/kubernetes/test-infra/pull/37758 to fix the machine mode lane

ffromani:
@killianmuldoon thanks for your patience, updating took longer than I hoped. Ready for another review round!

killianmuldoon:
> @killianmuldoon thanks for your patience, updating took longer than I hoped. Ready for another review round!

Great - thanks for the ping. I'll dedicate some time tomorrow to review.

killianmuldoon:
I've been playing around with this a lot today and it looks really good - no real review findings. I'm still trying to figure out how to adapt it to the slurm-bridge hybrid cluster case. I want to make sure that's not blocked off by this approach.  I'll come back and give it another review next week.

Thanks again for working on this!

ffromani:
> I've been playing around with this a lot today and it looks really good - no real review findings. I'm still trying to figure out how to adapt it to the slurm-bridge hybrid cluster case. I want to make sure that's not blocked off by this approach. I'll come back and give it another review next week.
> 
> Thanks again for working on this!

Great news, thanks! be advised there are missing parts which I will implement as followup PRs, outlined in #304 . Most notably, the code does NOT report the `cpuIDs` with external allocator, a key feature requested and which I'll add, but depends on this PR. Likewise, machine mode will gain `cpuIDs` reporting.

Other than that looking forward to hear more feedback. Unblocking external schedulers and be able to solve the individual mode conundrum is really a key feature for 0.3.0

pkg/driver/dra_hooks.go
```diff
@@ -144,74 +142,77 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
...
+			cur, err = cp.cpuAllocator.Allocate(logger, availableCPUs, preferredCPUs, claimCPUCount)
 			logger.V(2).Info("using opaque config CPU assignment", "device", alloc.Device, "assigned", cur.String())
 		}
 
 		if err != nil {
 			return kubeletplugin.PrepareResult{Err: err}
 		}
-		cpuAssignment = cpuAssignment.Union(cur)
-		logger.V(2).Info("CPU assignment for device", "device", alloc.Device, "assigned", cur.String(), "allAssigned", cpuAssignment.String())
+		if err := cp.cpuAllocator.Validate(cur, assignedCPUs, cp.cpuAllocationStore.GetPreparedCPUs()); err != nil {
```
pravk03, line 203: We can have opaque config at request scope, so having count > 1 would fail validation IIUC. 

Eg:
```
requests:
- name: cpu-req-1
  exactly:
    deviceClassName: dra.cpu
    count: 2
    capacity:
      requests:
        dra.cpu/cpu: "2" 
config:
- opaque:
    driver: dra.cpu
    parameters:
      cpuConfig: {cpuset: "0,1,4,5"}  
```

Should we just disallow count != 1 in machine mode ?. I don't see a reason to do it since we're setting exact CPUs.

pkg/driver/dra_hooks.go
ffromani: yes, it can because ultimately we use `alloc` only to access `Request`, which is indeed a loop invariant.
The real cost to pay, though, is to roll out our own version of `resourceclaim.ConfigForResult()` called in `getOpaqueCPUSet`. I think is worth it. Would it be fine for you?

pkg/driver/dra_hooks.go
ffromani: Thanks for raising this point!

TL;DR: yes, I think we should restrict and document the restriction for the short/mid term and revisit later depending on how the external scheduler integration progresses.

Details:

This is one of the areas I was tip-tapping around. It basically it boils down to:
1. you are correct: the code in machine mode for count > 1 is currently buggy, and most notably
2. does it make sense to support count > 1 in machine mode at all?

I reviewed (quickly!) docs/user/opaque-cpouset-overrides.md and while it does mention limitations and restrictions, it doesn't _explicitly_ prescribes count=1. So I'm not sure it actually fails validation, and this deserves tests anyway.

The tension here is that machine mode would be the odd one which would limit the DRA API semantics a bit more, restricting count to exactly 1.
This restriction feels unnecessary to me, but OTOH machine mode itself is a bit of more experimental than other grouping modes, so a further restriction is defensible and probably the best solution in the short term while we keep machine mode "experimental" ("staging" is probably better description?)

Longer term I think that we should decide if we keep machine mode or not.
1. keep it: then make machine mode should  work as closely as possible (ideally identically) to other grouping modes, unless there's a clear strong and *intrinsic* reason
2. deprecate it: then not worth spending more energy.

internal/driverconfig/validate.go
```diff
@@ -0,0 +1,65 @@
...
+		if c.GroupBy != device.GROUP_BY_SOCKET && c.GroupBy != device.GROUP_BY_NUMA_NODE && c.GroupBy != device.GROUP_BY_MACHINE {
+			return fmt.Errorf("invalid groupBy %q: must be %q, %q, or %q",
+				c.GroupBy, device.GROUP_BY_SOCKET, device.GROUP_BY_NUMA_NODE, device.GROUP_BY_MACHINE)
+		}
+	}
+	if c.Allocator != AllocatorExternal && c.Allocator != AllocatorCPUManager {
+		return fmt.Errorf("invalid allocator %q: must be %q or %q",
+			c.Allocator, AllocatorExternal, AllocatorCPUManager)
+	}
+	if c.CPUDeviceMode == device.CPU_DEVICE_MODE_GROUPED && c.GroupBy == device.GROUP_BY_MACHINE && c.Allocator != AllocatorExternal {
```
AutuSnow, line 45: I am thinking that we included groupBy=machine in the release v0.2.0, and the allocator key is not displayed in the chart. Therefore, will the existing machine mode deployment fail during the upgrade due to "invalid cpumanager allocator for groupBy machine" and Resolve()? Should machine mode default to cpumanager allocator or should documentation and release instructions be used for migration?

pkg/driver/dra_hooks.go
AutuSnow: On the count=2 case you were unsure about: opaque cpuset 0,1,4,5 split into two 2-CPU results fails deterministically here with "requested CPUs 0-1 ... already assigned to another device in this claim". If machine mode restricts to count=1, could that be rejected explicitly instead?

internal/driverconfig/config.go
```diff
@@ -43,7 +48,8 @@ type Config struct {
 	// PublishNodeAllocatableResourceMapping publishes KEP-5517 nodeAllocatableResources
 	// mappings in ResourceSlice devices. Requires the DRANodeAllocatableResources
 	// feature gate in the cluster. Defaults to false.
-	PublishNodeAllocatableResourceMapping bool `json:"publishNodeAllocatableResourceMapping,omitempty"`
+	PublishNodeAllocatableResourceMapping bool   `json:"publishNodeAllocatableResourceMapping,omitempty"`
+	Allocator                             string `json:"allocator,omitempty"`
```
AutuSnow, line 52: Could allocator be added to schemaEnums so the generated values.schema.json carries an enum like cpuDeviceMode and groupBy? Right now a typo such as allocator=extenral renders fine and only fails once the driver starts.

ffromani:
rebased only, comments not yet addressed

internal/driverconfig/validate.go
ffromani: Thanks for raising. A prior version of this PR forced allocator mode to external when machine mode was selected. Then we decided to tighten the validation, but I forgot to add a clear documentation note (action required on upgrades). Let me address this.

internal/driverconfig/config.go
ffromani: absolutely, forgot to do that, thanks for the callout

pkg/driver/dra_hooks.go
ffromani: after some experiment, extracting the code was the easy part. The hard part is to figure out which `RequestName` to feed it, and the best/only authoritative way is through `claim.Status.Allocation.Devices.Results[*].Request`. With this constraint, I can't think of a safe way to hoist out of the loop `GetPreferredCPUs`; even forcing `count==1` I'm not sure it helps or is everything we need, and for non-machine grouping it seems an unnecessary constraint.

For all these reasons I'm parking this change for now; IMO it's a possible optimization anyway, not a correctness concern, so postponing seems fair to me.

ffromani:
addressed all comments. Intentionally deferring updates and rebases to #316 until we reach consensus on the general direction; ofc I'll keep update this PR as needed.

pkg/driver/dra_hooks.go
ffromani: I *think* I addressed this concern in my last commit. PTAL folks.

pkg/driver/dra_hooks.go
```diff
@@ -392,75 +401,16 @@ func (cp *CPUDriver) HandleError(ctx context.Context, err error, msg string) {
...
-
-	for i := range configs {
-		config := &configs[i]
-		if config.Opaque.Driver != cp.driverName {
-			continue
-		}
-		if config.Source != resourceapi.AllocationConfigSourceClaim {
-			return cpuset.CPUSet{}, false, fmt.Errorf("opaque config: configuration from DeviceClass is not supported by this driver, custom cpusets must be defined per ResourceClaim request")
+func checkDeviceRequestCountUpTo(claim *resourceapi.ResourceClaim, limit int64, groupBy string) error {
+	for _, req := range claim.Spec.Devices.Requests {
```
AutuSnow, line 417: Should this only look at requests this driver actually serves? A claim mixing a machine-mode CPU request with another driver's multi-device request would be rejected here, even though the rest of prepareGroupedResourceClaim filters results by alloc.Driver.
`
requests:
- name: cpus
  exactly: { deviceClassName: dra.cpu, capacity: { requests: { dra.cpu/cpu: "64" } } }
- name: gpus
  exactly: { deviceClassName: nvidia.com/gpu, count: 8 }
`


This request will be rejected by the CPU driver in "machine" mode, triggering an error stating that driver grouping mode "machine" supports exact device requests of up to 1; the CPU request itself is perfectly valid, and the `count=8` applies to the GPUs. Multi-GPU nodes are the norm in SLURM clusters.

ffromani: YES! I should have guarded that applies only to the request managed by this driver. Thanks for catching. Working on better tests to catch these issues.

killianmuldoon:
/lgtm

This is looking in really good shape now. I've been working with it +/- #316  for the last ~week and it's pretty rock solid. Thanks for the great work @ffromani - this is a major improvement over individual mode.

ffromani:
@pravk03 @AutuSnow just to REALLY turn even the tiniest stone, are you good with moving forward with this approach? any pending items? please let me know if you want to see a followup (or revert!) before we move forward with #316

pravk03:
@ffromani This looks great. No pending items or follow-ups from me and good to move forward with #316. Thanks for all your work on this !!

AutuSnow:
@ffromani My comments hava been processed and now they look gread. thank you for your patience in editing them

ffromani:
thanks all folks. Moving forward with #316

PR #230, opened 2026-07-07
codeorg: cleanup: introduce the cpuenumerator abstraction

ffromani:
#### What type of PR is this?

/kind cleanup

This is the promised followup of #183 which completes the cleanup introducing the `cpuenumerator` object.
The main goal is to push back the gravitational pull of the CPUDriver god object and to decompose it before it's too late and becomes unwieldly.

An obvious split is device enumeration, so we start with that.

N/A

This work suggests another possible split and slim down of CPUDriver at the cpu allocation boundary, explored in #229 - but this PR is fully independent

#### Release note

```release-note
NONE
```


ffromani:
/cc @AutuSnow @pravk03 

the promised and final part of the cleanup started in #183

AutuSnow:
Thanks！！
/lgtm

PR #239, opened 2026-07-17
Preserve prepared CPU allocations across container restarts

### LINKED ISSUE #235: StopContainer early release breaks guaranteed container restarts within an existing Pod sandbox
`StopContainer` removes a guaranteed container's ResourceClaim CPU allocation from `cpuAllocationStore`.
﻿
This breaks a normal container restart when the Pod sandbox remains running, because kubelet does not call `NodePrepareResources` again for that restart. The replacement container still receives the existing CDI-injected `DRA_CPUSET_<claimUID>` environment variable, but the NRI `CreateContainer` hook rejects it because the claim is no longer present in the driver's prepared allocation store.

AutuSnow:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Fixes guaranteed CPU containers failing to restart when kubelet creates a replacement container without calling `NodePrepareResources` again.

Previously, `StopContainer` removed the allocation and ownership state used to validate the CDI-injected `DRA_CPUSET_<claimUID>` value. The replacement container still received that value, but `CreateContainer` rejected it because the trusted prepared state was gone.

This change separates prepared reservations from allocations enforced by running containers:

1. `NodePrepareResources` reserves the CPUs for the claim.
2. `CreateContainer` validates the CDI allocation and marks it enforced.
3. `StopContainer` marks it pending, returning the CPUs to the shared pool while retaining the reservation and owner.
4. A replacement container can enforce the same allocation without another prepare call.
5. `NodeUnprepareResources` removes the allocation and owner.

It also makes reservation and multi-claim enforcement atomic, rolls back reservations when CDI preparation fails, and adds unit and end-to-end coverage for the restart flow.

Prepared state remains in memory. Persisting it across driver restarts is tracked separately in #<checkpoint-issue>.

#### Which issue(s) this PR fixes:

Fixes #235

#### Special notes for your reviewer:

The implementation still assumes that a ResourceClaim is owned by one container. Claim sharing, preemption, and mixed shared/exclusive use remain out of scope.

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -374,6 +374,9 @@ func (cp *CPUDriver) unprepareResourceClaim(logger logr.Logger, claim kubeletplu
 	}
 	cp.cpuAllocationStore.RemoveResourceClaimAllocation(logger, claim.UID)
 	cp.claimTracker.Cleanup(claim.UID)
+	// Existing shared containers pick up the expanded cpuset on their next CreateContainer
+	// or Synchronize. Updating them here requires NRI UpdateContainers support across all
+	// supported runtime versions.
 	return nil
 }
 
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -123,7 +123,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 	}
 
 	var cpuAssignment cpuset.CPUSet
-	sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
+	allocatableCPUs := cp.cpuAllocationStore.GetAllocatableCPUs()
 	for _, alloc := range claim.Status.Allocation.Devices.Results {
 		if alloc.Driver != cp.driverName {
 			continue
@@ -153,7 +153,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 				return kubeletplugin.PrepareResult{Err: fmt.Errorf("no valid socket ID found for device %s", alloc.Device)}
 			}
 			socketCPUs := topo.CPUDetails.CPUsInSockets(socketID)
-			availableCPUsForDevice := sharedCPUs.Difference(cpuAssignment).Intersection(socketCPUs)
+			availableCPUsForDevice := allocatableCPUs.Difference(cpuAssignment).Intersection(socketCPUs)
 			logger.V(4).Info("socket CPU availability", "socketID", socketID, "socketCPUs", socketCPUs.String(), "availableCPUs", availableCPUsForDevice.String())
 			cur, err = cpumanager.TakeByTopologyNUMAPacked(logger, topo, availableCPUsForDevice, int(claimCPUCount), cpumanager.CPUSortingStrategyPacked, true)
 		case device.GROUP_BY_NUMA_NODE:
@@ -162,7 +162,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 				return kubeletplugin.PrepareResult{Err: fmt.Errorf("no valid NUMA node ID found for device %s", alloc.Device)}
 			}
 			numaCPUs := topo.CPUDetails.CPUsInNUMANodes(numaNodeID)
-			availableCPUsForDevice := sharedCPUs.Difference(cpuAssignment).Intersection(numaCPUs)
+			availableCPUsForDevice := allocatableCPUs.Difference(cpuAssignment).Intersection(numaCPUs)
 			logger.V(4).Info("NUMA node CPU availability", "numaNodeID", numaNodeID, "numaCPUs", numaCPUs.String(), "availableCPUs", availableCPUsForDevice.String())
 			cur, err = cpumanager.TakeByTopologyNUMAPacked(logger, topo, availableCPUsForDevice, int(claimCPUCount), cpumanager.CPUSortingStrategyPacked, true)
 		case device.GROUP_BY_MACHINE:
@@ -193,11 +193,14 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		return kubeletplugin.PrepareResult{}
 	}
 
+	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, cpuAssignment); err != nil {
+		return kubeletplugin.PrepareResult{Err: err}
+	}
 	result := cp.prepareDevices(logger, claim, cpuAssignment)
 	if result.Err != nil {
+		cp.cpuAllocationStore.RemoveResourceClaimAllocation(logger, claim.UID)
 		return result
 	}
-	cp.cpuAllocationStore.AddResourceClaimAllocation(logger, claim.UID, cpuAssignment)
 	cp.metricsRecorder().RecordClaimAllocatedCPUs(cpuAssignment.Size())
 	cp.refreshAllocationMetrics()
 	return result
@@ -243,19 +246,22 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 		return cp.prepareDevices(logger, claim, existingCPUs)
 	}
 
-	// All the CPUs allocated to a claim should currently be in the shared pool.
-	sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
-	if !claimCPUSet.IsSubsetOf(sharedCPUs) {
+	// All the CPUs allocated to a claim must not be prepared for another claim.
+	allocatableCPUs := cp.cpuAllocationStore.GetAllocatableCPUs()
+	if !claimCPUSet.IsSubsetOf(allocatableCPUs) {
 		return kubeletplugin.PrepareResult{
 			Err: fmt.Errorf("claim %s/%s has overlapping device assignment with other claims", claim.Namespace, claim.Name),
 		}
 	}
 
+	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, claimCPUSet); err != nil {
+		return kubeletplugin.PrepareResult{Err: err}
+	}
 	result := cp.prepareDevices(logger, claim, claimCPUSet)
 	if result.Err != nil {
+		cp.cpuAllocationStore.RemoveResourceClaimAllocation(logger, claim.UID)
 		return result
 	}
-	cp.cpuAllocationStore.AddResourceClaimAllocation(logger, claim.UID, claimCPUSet)
 	cp.metricsRecorder().RecordClaimAllocatedCPUs(claimCPUSet.Size())
 	cp.refreshAllocationMetrics()
 	return result
@@ -365,6 +371,7 @@ func (cp *CPUDriver) unprepareResourceClaim(logger logr.Logger, claim kubeletplu
 		return err
 	}
 	cp.cpuAllocationStore.RemoveResourceClaimAllocation(logger, claim.UID)
+	cp.claimTracker.Cleanup(claim.UID)
 	return nil
 }
 
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -193,6 +193,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		return kubeletplugin.PrepareResult{}
 	}
 
+	// Reserve before CDI I/O so concurrent Prepare calls cannot select the same CPUs.
 	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, cpuAssignment); err != nil {
 		return kubeletplugin.PrepareResult{Err: err}
 	}
@@ -254,6 +255,7 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 		}
 	}
 
+	// Reserve before CDI I/O so concurrent Prepare calls cannot select the same CPUs.
 	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, claimCPUSet); err != nil {
 		return kubeletplugin.PrepareResult{Err: err}
 	}
@@ -461,7 +463,7 @@ func (cp *CPUDriver) validateOpaqueCPUSet(opaqueCPUSet cpuset.CPUSet, onlineCPUs
 	}
 
 	// Verify cores do not overlap with other active claims on this node
-	existingClaimCPUs := cp.cpuAllocationStore.GetAllocatedCPUs()
+	existingClaimCPUs := cp.cpuAllocationStore.GetPreparedCPUs()
 	if opaqueCPUSet.Intersection(existingClaimCPUs).Size() > 0 {
 		return fmt.Errorf("requested CPUs %s from opaque config conflict with already allocated claims", opaqueCPUSet.String())
 	}
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -123,7 +123,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 	}
 
 	var cpuAssignment cpuset.CPUSet
-	allocatableCPUs := cp.cpuAllocationStore.GetAllocatableCPUs()
+	allocatableCPUs := cp.cpuAllocationStore.GetSharedCPUs()
 	for _, alloc := range claim.Status.Allocation.Devices.Results {
 		if alloc.Driver != cp.driverName {
 			continue
@@ -248,7 +248,7 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 	}
 
 	// All the CPUs allocated to a claim must not be prepared for another claim.
-	allocatableCPUs := cp.cpuAllocationStore.GetAllocatableCPUs()
+	allocatableCPUs := cp.cpuAllocationStore.GetSharedCPUs()
 	if !claimCPUSet.IsSubsetOf(allocatableCPUs) {
 		return kubeletplugin.PrepareResult{
 			Err: fmt.Errorf("claim %s/%s has overlapping device assignment with other claims", claim.Namespace, claim.Name),
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -374,9 +374,9 @@ func (cp *CPUDriver) unprepareResourceClaim(logger logr.Logger, claim kubeletplu
 	}
 	cp.cpuAllocationStore.RemoveResourceClaimAllocation(logger, claim.UID)
 	cp.claimTracker.Cleanup(claim.UID)
-	// Existing shared containers pick up the expanded cpuset on their next CreateContainer
-	// or Synchronize. Updating them here requires NRI UpdateContainers support across all
-	// supported runtime versions.
+	// TODO(#279): Update existing shared containers here once all supported runtimes can
+	// safely process unsolicited NRI UpdateContainers calls. Until then, each container
+	// picks up the expanded cpuset on its next CreateContainer or Synchronize.
 	return nil
 }
 
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -95,7 +95,7 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 			if len(claimUIDs) == 0 {
 				state = store.NewContainerState(container.GetName(), containerUID)
 			} else {
-				if _, err := claimTracker.SetOwners(cLogger, claimUIDs, types.UID(pod.Uid), container.Name); err != nil {
+				if _, err := claimTracker.SetOwner(cLogger, types.UID(pod.Uid), container.Name, claimUIDs...); err != nil {
 					return nil, err
 				}
 				if err := cpuAllocationStore.ValidateResourceClaimAllocations(validatedClaimAllocations); err != nil {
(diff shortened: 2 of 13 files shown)


ffromani (commented):
I can only glance at the code and I see you are adding a checkpoint file
I'm not against checkpoint, and we probably need it. I will reserve the final call once I can have time to properly review.

If we go the checkpoint path, there are lessons we must learn from the kubelet experience
- the checkpoint path must be fully configurable. Hardcoded paths have been a huge pain in the kubelet experience. We need to enable admis to have private data (checkpoint, state) in one path and public data (config files, etc.) in possibly complete separated path
- the checkpoint format must be private and opaque. We can use json internally, but the content and the format are implementation details. User must be actively discouraged to consume internal checkpointed data
- the format must be forward and backward compatible to avoid migration path. 1 version skew is usually sufficient.

I recommend to take these points into account but NOT act upon them prior to the proper review to save you effort and to make efficient use of your time.

AutuSnow:
Thanks, this is helpful and lines up with the kubelet history. Where it stands today: the path is hardcoded under the plugin dir, and there's a version field but it hard-rejects any mismatch rather than tolerating a skew. Agree on all three. I'll leave it until your proper review like you said, then make the path configurable and loosen the version check to allow 1 skew. When you get there: separate flag for the private/public path split, or fold it into the config-file support that just landed?

pravk03:
Thanks for catching this. 

+1 to what @ffromani mentioned above. I'd suggest splitting the checkpointing topic into a separate issue for further discussion/exploration.

For this bug, the fix may be simple. A claim's  lifetime is tied to the pod and not the container (since we do not allow sharing across pods currently or even with KEP-5517), so we can release it on unprepare / pod sandbox deletion instead of on `StopContianer` ?

AutuSnow:
Thanks, I agree that checkpointing should move to a separate issue/PR.
For this bug, the prepared claim must survive StopContainer and be released by NodeUnprepareResources. I’d avoid using PodSandbox deletion as the release point, since kubelet can recreate a sandbox while the Pod and its claim are still alive.
One trade-off is that keeping the allocation until Unprepare leaves those CPUs out of the shared pool during the stop/restart window. The current PR keeps a pending/enforced distinction to preserve early shared-pool release while retaining the prepared claim. I can either keep that small lifecycle change here or simplify it and accept the temporary utilization regression.

pkg/store/cpu_allocation.go
```diff
@@ -108,16 +197,16 @@ func (s *CPUAllocation) GetReservedCPUs() cpuset.CPUSet {
 func (s *CPUAllocation) GetAllocatedCPUs() cpuset.CPUSet {
```
pravk03: nit: GetPreparedCPUs() ?

Its better to keep the function consistent with what they are returning.

pkg/store/cpu_allocation.go
```diff
@@ -54,47 +61,129 @@ func NewCPUAllocation(cpuTopology *cpuinfo.CPUTopology, reservedCPUs cpuset.CPUS
...
+		}
+		allocation.enforced = true
+		s.resourceClaimAllocations[claimUID] = allocation
+		s.enforcedCPUs = s.enforcedCPUs.Union(allocation.cpus)
+	}
+	return nil
+}
+
+// MarkResourceClaimsPending stops enforcing claims while retaining their prepared reservations.
+func (s *CPUAllocation) MarkResourceClaimsPending(claimUIDs ...types.UID) {
```
pravk03: nit: Calling this `Pending` is confusion, because pending generally refers to claim not being allocated. 

Alternate naming option:
`SetResourceClaimsEnforcementState(claimUIDs ...types.UID, enforced bool)`

pkg/driver/dra_hooks.go
```diff
@@ -242,19 +245,22 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
...
-	if !claimCPUSet.IsSubsetOf(sharedCPUs) {
+	// All the CPUs allocated to a claim must not be prepared for another claim.
+	allocatableCPUs := cp.cpuAllocationStore.GetAllocatableCPUs()
+	if !claimCPUSet.IsSubsetOf(allocatableCPUs) {
 		return kubeletplugin.PrepareResult{
 			Err: fmt.Errorf("claim %s/%s has overlapping device assignment with other claims", claim.Namespace, claim.Name),
 		}
 	}
 
+	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, claimCPUSet); err != nil {
```
pravk03, line 259: Any reason to move this up ?

pkg/driver/nri_hooks.go
pravk03, line 290: I think we will hit a problem here. Remove is called async. In the restart scenario, the old container can be removed after the new container has started and we end up removing the state here. 

This was a preexisiting bug, but since we are fixing the restart scenario in this PR, I think we should address this too.

pkg/driver/nri_hooks.go
```diff
@@ -207,6 +207,9 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
 			guaranteedCPUs = guaranteedCPUs.Union(cpus)
 			claimUIDs = append(claimUIDs, uid)
 		}
+		if err := cp.cpuAllocationStore.EnforceResourceClaims(claimAllocations); err != nil {
```
pravk03: This function introduces some duplication with `validatePreparedClaimAllocation()` which is called above.

test/image/dracputester/app.go
```diff
@@ -87,6 +88,9 @@ func affinityScanBoundFromTopology(topo *cpuinfo.CPUTopology) int {
 }
 
 func main() {
+	exitAfter := flag.Duration("exit-after", 0, "exit after this duration to exercise container restart handling")
```
pravk03: nit: test code

pkg/store/cpu_allocation.go
```diff
@@ -54,47 +61,129 @@ func NewCPUAllocation(cpuTopology *cpuinfo.CPUTopology, reservedCPUs cpuset.CPUS
...
+		resourceClaimAllocations: make(map[types.UID]claimAllocation),
+		preparedCPUs:             cpuset.New(),
+		enforcedCPUs:             cpuset.New(),
 	}
 }
 
-// AddResourceClaimAllocation adds a new resource claim allocation to the store.
+// AddResourceClaimAllocation adds an allocation already enforced by a running container.
+// It is used while reconstructing state from the NRI runtime.
 func (s *CPUAllocation) AddResourceClaimAllocation(logger logr.Logger, claimUID types.UID, cpus cpuset.CPUSet) {
```
pravk03: nit: This function is almost identical to `ReserveResourceClaimAllocation` ,  we can remove the duplication by passing additional input for enforcement to `ReserveResourceClaimAllocation` . Or, calling both `ReserveResourceClaimAllocation` and  `EnforceResourceClaims` both in Synchronize()

pravk03 (commented):
Thanks. I like the direction of having `prepared` and `enforced` CPUs in allocation state. Added some initial review comments, I will do a more thorough pass soon.

pkg/store/cpu_allocation.go
AutuSnow: Done

pkg/store/cpu_allocation.go
AutuSnow: Agreed. Renamed it to SetResourceClaimsEnforcementState. I kept the claim IDs variadic, so the enforced flag comes first in the Go signature.

pkg/driver/dra_hooks.go
AutuSnow: Yes. The reservation needs to happen before the CDI update. Otherwise another concurrent Prepare could select the same CPUs while the CDI operation is still in progress. Added a comment for this.

pkg/driver/nri_hooks.go
AutuSnow: Good catch. Container state is now removed only when the runtime container ID still matches. Delayed StopContainer and RemoveContainer events for the old container are ignored, with a regression test covering both cases.

test/image/dracputester/app.go
AutuSnow: Yep, this flag is only used by dracputester for the restart e2e test.

pkg/store/cpu_allocation.go
AutuSnow: Agreed. Removed AddResourceClaimAllocation. Synchronize now calls ReserveResourceClaimAllocation followed by EnforceResourceClaims.

pkg/driver/nri_hooks.go
AutuSnow: Right. I removed validatePreparedClaimAllocation and kept the validation in EnforceResourceClaims. Newly added claim owners are rolled back if that validation fails.

pravk03:
I was thinking more about this problem.

The current implementation correctly fixes the restart issue, but in doing so it can affect every other workload on the node. With this PR:

* StopContainer - releases CPU back to the shared pool 
* CreateContainer - removes CPUs from the shared pool 

So a guaranteed container in a crashloop grows and shrinks the cgroups of all shared containers once per restart cycle indefinitely (bounded by the kubelet's backoff).  Correctness-wise this is fine, but each shrink forces the kernel to migrate threads off the removed CPUs. This causes periodic jitter for every colocated pod (and may other scalability concerns with N_pods*C_containers_per_pod cgroup updates per restart). Alternative to this is, hold the CPUs out of the shared pool from ResourceClaim `Prepare` until `Unprepare`. This causes no churn during the restart scenario, but the obvious drawback is that we do not update shared CPUs until the next container create event. @AutuSnow @ffromani Do you think this is a valid concern ? 
[Update] How about using [UpdateContainers](https://github.com/containerd/nri/blob/84180b63351d03b54a317a265ba01e3e9db19f24/pkg/api/api.proto#L31) RPC ?. The comment indicates that this can be triggered adhoc. Maybe we can call this when we unprepare resource claims. If this works, may simplify state management greatly.

Sorry for not posting this question sooner which may have lead to some added code churn. I only realized in this during the latest round of review. Let's make sure we align on this before making further code changes. 

(PS: Worth noting that this is not a regression introduced by this PR. Before the PR, restarts were simply broken.)

AutuSnow:
The churn concern is valid. I agree that the allocation should remain excluded from the shared pool from Prepare until Unprepare, so Stop/Create during a restart does not change shared cpusets.
UpdateContainers is the right primitive for immediate expansion after Unprepare, but there is one runtime compatibility concern: plugin-initiated updates can deadlock containerd when its runtime-side NRI predates containerd/nri#301 (v0.12.1). Current stable containerd releases supported by this project still vendor v0.8.0-v0.12.0, and updating this driver's NRI dependency alone does not fix that runtime-side lock inversion.
I suggest simplifying this PR to keep CPUs reserved until Unprepare and accepting reconciliation on the next Create/Synchronize. We can add unsolicited updates once the runtime fix is available in our supported versions, or use it here only if we intentionally raise the minimum runtime requirement.

pravk03:
> I suggest simplifying this PR to keep CPUs reserved until Unprepare and accepting reconciliation on the next Create/Synchronize

Still thinking if there is a way around it, but the options seem limited.

pravk03:
Given the options, I agree with the plan in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/239#issuecomment-5113034921. 
I think we should document the gap - Reconciliation happens on the next Create/Synchronize. Lets also create an issue and leave a TODO that this will be handled using `UpdateContainers` in the future.

ffromani:
This PR is waiting in my queue so I'm not fully up to speed with the review (the dreaded "postponing because deserves more mental bandwidth" reviewer deadlock), but I just want to mention that I'll get to this but I don't want to stall.

In any case, my general stance is that correctness almost always prevails over everything, so if we need to wait a little longer in order to release CPUs but this delay makes the behavior correct, so be it. I hope to have bandwidth to have a pass before end of the week (before my availability further declines). If you reach consensus please move forward; we can always optimize later and correctness tests actually make for a better base for optimizations.

[EDIT] likewise IIUC I left a comment about checkpoints, feel free to move forward restoring correctness, we can iterate later and revisit the checkpointing matter, surely before 0.3.0.

pravk03:
Thanks @ffromani. I think more input would be valuable here.
My main goal was to make sure we don't have @AutuSnow go through the effort of implementing and testing something we might need to modify/revert. IMO, taking a little extra time to align on the best path is well worth it.

AutuSnow:
@ffromani @pravk03  I think the best path for this PR is to keep the claim allocation reserved from Prepare until Unprepare, and to accept shared-pool reconciliation on the next Create/Synchronize for now.
﻿
That keeps restart behavior correct and avoids restart-driven shared-pool churn for all colocated workloads. I don't see a safer immediate-release mechanism in our current support matrix: PodSandbox deletion is too early/ambiguous for claim lifetime, and unsolicited UpdateContainers looks like a follow-up optimization once the runtime-side NRI fix is available in supported containerd versions.
﻿
Given that, I would also consider simplifying the implementation accordingly and keeping any UpdateContainers-based improvement as a separate follow-up issue/PR.

ffromani:
fair enough. I'll try to get to this work as soon as I can with full context. Thanks for your patience.

AutuSnow:
> fair enough. I'll try to get to this work as soon as I can with full context. Thanks for your patience.

Don't worry, I won't change them until we reach an agreement

pkg/store/pod_config.go
```diff
@@ -92,19 +92,20 @@ func (s *PodConfig) GetContainerState(podUID types.UID, containerName string) *C
 	return nil
 }
 
-// RemoveContainerState removes a container's state from the store.
-func (s *PodConfig) RemoveContainerState(podUID types.UID, containerName string) []types.UID {
+// RemoveContainerState removes a container's state if its runtime ID still matches.
+// A replacement container can have the same name while an old Remove event is in flight.
```
ffromani, line 96: let's document the return value. Why do we need an extra bool return? Is it sufficient to check `len(claimUIDs) > 0` ?

ffromani:
Thanks everyone again for the patience and thanks for taking my opinion in high esteem to delay this PR.
I'm taking the liberty to summarize the current design trend and direction, then I'll check the implementation.
So, designwise/directionwise:
- The core idea #26 wanted to embody was not "release early" but rather "do not depend on other lifecycle events to release container resources" IOW, the release should be bound to the parent container lifecycle, not to other container lifecycle; in even simpler terms, each container should cleanup after themself. The idea of release early kinda folded together with this concept (largely because my bad!) but correctness trumps over efficiency, so "early" for us actually means **at Unprepare time**.
- For this driver, the reservation is a DRA concern, the actuation is a NRI concern. Thinking carefully, releasing resources is a reservation responsability, which steers us again towards "at Unprepare time" - a reservation lifecycle event, not towards NRI (aka actuation) events.
- This was further reinforced by the good research done by @AutuSnow in #235, which I'm paraphrasing here. I asked to compare with the closest relative, the NRI topology-aware plugin; topology-aware does NOT have a reservation mechanism - it's a pure NRI driver, so it can re-allocate at each restart. Doing so is _not_ a breach of the reservation contract, but in our case we do have, the DRA claim
- The churn concern is valid. It's a second-order effect which is going to bite us later; I do think that the existing isolation mechanism shield the exclusively-assigned workloads from this churn, but still is not something we should do intentionally, so better to avoid it. Avoiding it also align with the (very solid) fail-close principle. All in all, I'd call this form of early release an optimization, which we can re-evaluate in the future, once correctness is achieved (and tested with our tests!)
- In any case, Unprepare is expected to happen "soon", so the extra-reservation aka suboptimal behavior should be tolerable.
- I concur calling UpdateContainers at Unprepare time is the right approach, and also embodies and preserve the good part of #26, the part which I still think is very relevant. Too bad we need to wait for it. I think we would need to block 0.3.0 on the availability of this API and use it.
[EDIT] - this is my own current stance and not a decision - obviously - but something we would need to discuss and agree, and surely would be non-blocking for this PR and followup work.
- Lastly, I thought about batching UpdateContainers call using something like a reconcile loop. But thinking twice the async release seems way simpler and totally good enough. I'd keep this option open but don't bother anytime soon.

pkg/store/claim_tracker.go
```diff
@@ -57,26 +57,42 @@ func NewClaimTracker() *ClaimTracker {
 }
 
 func (ctk *ClaimTracker) SetOwner(logger logr.Logger, claimUID, podUID k8stypes.UID, containerName string) error {
+	_, err := ctk.SetOwners(logger, []k8stypes.UID{claimUID}, podUID, containerName)
+	return err
+}
```
ffromani: I think this is used only in test code after your PR, so it can be dropped.
Actually, I'd keep this `SetOwner` singular because it wants to mean "set the owner of these claims" and I'd change the API to read:
```
// SetOwner atomically binds claims to a single container. It returns the claims
// which were newly bound so callers can roll them back if a later operation fails.
func (ctk *ClaimTracker) SetOwner(logger logr.Logger, podUID k8stypes.UID, containerName string, claimUIDs ...k8stypes.UID) ([]k8stypes.UID, error) {
        if len(claimUIDs) == 0 { // this should never happen
                return nil, errors.New("no claims to bind")
        }
         curIdent := OwnerIdent{
                 PodUID:        podUID,
                ContainerName: containerName,
        }
        ctk.mu.Lock()
        defer ctk.mu.Unlock()

       ...
```

The variadic form makes for a little nicer code; everything else looks fine to me

pkg/store/claim_tracker_test.go
```diff
@@ -149,6 +149,39 @@ func TestSetOwner(t *testing.T) {
...
+	t.Run("returns only newly bound claims", func(t *testing.T) {
+		tracker := NewClaimTracker()
+		require.NoError(t, tracker.SetOwner(logger, "claim-1", owner.PodUID, owner.ContainerName))
+
+		newlyBound, err := tracker.SetOwners(logger, []k8stypes.UID{"claim-1", "claim-2"}, owner.PodUID, owner.ContainerName)
+		require.NoError(t, err)
+		require.Equal(t, []k8stypes.UID{"claim-2"}, newlyBound)
+		require.Equal(t, 2, tracker.Len())
+	})
+}
```
ffromani: minor: I'd prefer tabular tests if the test share some conditions, or each test on its own. While I'd appreciate changing like this, is not a blocking requirement. But please consider if you resubmit.

test/image/dracputester/app.go
ffromani: I understand why you did this, and this is a pragmatic solution. But adding this option is something I'd do as last resort, and I'd be trying a different approach first: a sidecar ephemeral container.
When the tests whats to restart the driver, it would do something like
```
  ec := v1.EphemeralContainer{
      EphemeralContainerCommon: v1.EphemeralContainerCommon{
          Name:    "trigger-restart",
          Image:   "busybox",
          Command: []string{"kill", "-9", "1"},
      },
      TargetContainerName: "app",
  }
  client.CoreV1().Pods(ns).UpdateEphemeralContainers(ctx, name, pod, metav1.UpdateOptions{})
```

and opbserve the pod restart counter.
Furthermore, I think we should take this chance to create a new file `test/e2e/restart_test.go` to include all the tests which involve any form of restarts; these (important) tests are the most likely to flake, per kube experience, so we would need to label and isolate them clearly.

I acknowledge this is a big ask and increase the scope of the PR, so I'd be fine deferring the e2e test(s) to a followup; what matters is e2e coverage before 0.3.0 is cut (I think this is uncontroversial) and we can iterate to find the best e2e test shape; ofc if you are up to bundle everything here that's completely fine.

pkg/store/cpu_allocation.go
```diff
@@ -54,47 +61,127 @@ func NewCPUAllocation(cpuTopology *cpuinfo.CPUTopology, reservedCPUs cpuset.CPUS
...
+
+	for claimUID := range expected {
+		claimUIDs = append(claimUIDs, claimUID)
+	}
+	s.setResourceClaimsEnforcementStateLocked(true, claimUIDs...)
+	return nil
+}
+
+// SetResourceClaimsEnforcementState updates whether prepared claims are enforced by a running container.
+func (s *CPUAllocation) SetResourceClaimsEnforcementState(enforced bool, claimUIDs ...types.UID) error {
```
ffromani: probably no longer relevant since we agree to move forward and simplify - and release CPUs at `UnprepareResource` stage, but this seems to be called with `enforced=false`, therefore it should lose the `enforced` parameter and be renamed `ReleaseResourceClaims` or so.

ffromani (commented):
I think we're all aligned about the design of the fix (the bug was obvious once reported, I believe, and never challenged) and about the major items. I'm commenting about the points which caught my eye most about the implementation.

I'd be personally fine about landing the core fix and iterating later - iterations would be about refinements, not about pivoting/redoing work. Note I'd be mostly unavailable (offline) august 4-9, both included

pkg/store/cpu_allocation.go
```diff
@@ -55,46 +56,86 @@ func NewCPUAllocation(cpuTopology *cpuinfo.CPUTopology, reservedCPUs cpuset.CPUS
...
+// GetSharedCPUs returns CPUs available to shared containers.
 func (s *CPUAllocation) GetSharedCPUs() cpuset.CPUSet {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
-	return s.availableCPUs.Difference(s.allocatedCPUs)
+	return s.availableCPUs.Difference(s.preparedCPUs)
+}
+
+// GetAllocatableCPUs returns CPUs available for a new exclusive claim.
+func (s *CPUAllocation) GetAllocatableCPUs() cpuset.CPUSet {
```
pravk03: This function is exactly the same as GetSharedCPUs()

pravk03:
Sorry for the delay. I missed the last few updates on this PR. 
Needs a rebase and 1 cleanup comment I just added. Looks good to go after that.

ffromani:
/lgtm

all my comments have been addressed, and it seems to me also Praveen's have been.
I'd be happy to help writing the e2e test with ephemeral containers later on (this week hopefully).
I'm leaving the final approval to @pravk03

pravk03:
Thanks a lot.
 
/lgtm
/approve

PR #240, opened 2026-07-17
Fix allocation metrics after NRI state changes

AutuSnow:
Refreshes the Prometheus allocation gauges after NRI changes the CPU allocation store.
﻿
Previously:

- `Synchronize` rebuilt and replaced the allocation store without refreshing the gauges.
- `StopContainer` released claim allocations without refreshing the gauges.
﻿
As a result, allocation metrics could remain at zero after driver restart or report stale values between `StopContainer` and `UnprepareResourceClaims`.
﻿
﻿
﻿
﻿

### DIFF
--- a/pkg/driver/metrics_test.go
+++ b/pkg/driver/metrics_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/containerd/nri/pkg/api"
 	"github.com/go-logr/logr/testr"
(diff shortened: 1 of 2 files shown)


ffromani (commented):
/approve
/lgtm

thanks for the fix

ffromani:
we can probably think (low priority, unrelated to this PR) a followup e2e test similar in spirit to https://github.com/kubernetes-sigs/dra-driver-cpu/blob/main/test/e2e/contextual_logging_test.go#L42 which checks the basic properties of the metrics.

PR #241, opened 2026-07-20
test/e2e: cover metrics reconciliation after NRI restart

AutuSnow:
## Summary
- add a metrics-focused `ReportAfterSuite` that scrapes the `dracpu` pod `/metrics` endpoint through the pod proxy
- verify the allocation gauges reflect a real exclusive CPU claim, survive driver restart, and return to baseline after workload deletion
- add a small pod helper for fetching `/metrics` from the driver pod during e2e checks

Follow-up to the post-merge test suggestion on #240.

### DIFF
--- a/test/e2e/metrics_test.go
+++ b/test/e2e/metrics_test.go
@@ -1,0 +1,422 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"math"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/discovery"
+	"github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/fixture"
+	e2enode "github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/node"
+	e2epod "github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/pod"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/cpuset"
+)
+
+const (
+	driverDaemonSetName = "dracpu"
+	metricsPath         = "/metrics"
+
+	allocatedCPUsMetric        = "dra_cpu_allocated_cpus"
+	availableCPUsMetric        = "dra_cpu_available_cpus"
+	reservedCPUsMetric         = "dra_cpu_reserved_cpus"
+	activeResourceClaimsMetric = "dra_cpu_resource_claims_active"
+)
+
+type allocationMetricSnapshot struct {
+	AllocatedCPUs        int
+	AvailableCPUs        int
+	ReservedCPUs         int
+	ActiveResourceClaims int
+}
+
+func (s allocationMetricSnapshot) TotalCPUs() int {
+	return s.AllocatedCPUs + s.AvailableCPUs + s.ReservedCPUs
+}
+
+func (s allocationMetricSnapshot) String() string {
+	return fmt.Sprintf("allocated=%d available=%d reserved=%d activeClaims=%d total=%d",
+		s.AllocatedCPUs, s.AvailableCPUs, s.ReservedCPUs, s.ActiveResourceClaims, s.TotalCPUs())
+}
+
+// This serial spec complements the unit coverage in pkg/driver/metrics_test.go
+// with a real-cluster check that gauge state survives NRI restart and is released
+// again when the workload stops.
+var _ = ginkgo.Describe("Metrics", ginkgo.Serial, func() {
+	ginkgo.It("reconciles allocation metrics across driver restart and workload deletion", func(ctx context.Context) {
+		dracpuTesterImage := os.Getenv("DRACPU_E2E_TEST_IMAGE")
+		gomega.Expect(dracpuTesterImage).ToNot(gomega.BeEmpty(), "missing environment variable DRACPU_E2E_TEST_IMAGE")
+
+		rootFxt, err := fixture.ForGinkgo()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot create root fixture: %v", err)
+		metricsFxt := rootFxt.WithPrefix("metrics")
+		gomega.Expect(metricsFxt.Setup(ctx)).To(gomega.Succeed())
+		ginkgo.DeferCleanup(metricsFxt.Teardown)
+
+		targetNode, err := e2enode.PickWorker(ctx, metricsFxt.K8SClientset, 5*time.Second, 1*time.Minute, metricsFxt.Log)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		metricsFxt.Log.Info("using worker node", "nodeName", targetNode.Name)
+
+		daemonSet, err := rootFxt.K8SClientset.AppsV1().DaemonSets(daemonSetNamespace).Get(ctx, driverDaemonSetName, metav1.GetOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot get dracpu daemonset")
+		gomega.Expect(daemonSet.Spec.Template.Spec.Containers).ToNot(gomega.BeEmpty(), "no containers in dracpu daemonset")
+		orgDaemonSet := daemonSet.DeepCopy()
+
+		container := &daemonSet.Spec.Template.Spec.Containers[0]
+		cpuDeviceMode, _ := findArgInContainer(container, argCPUDeviceMode)
+		groupBy, _ := findArgInContainer(container, argGroupBy)
+
+		var claimSpec resourcev1.ResourceClaimSpec
+		if cpuDeviceMode == "grouped" && groupBy == "machine" {
+			reservedCPUs, err := reservedCPUsFromContainer(container)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			availableCPUs, err := discoverAvailableCPUs(ctx, metricsFxt, targetNode.Name, dracpuTesterImage, reservedCPUs)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			if availableCPUs.Size() == 0 {
+				ginkgo.Skip(fmt.Sprintf("no allocatable CPUs left on node %q", targetNode.Name))
+			}
+
+			cpuID := availableCPUs.UnsortedList()[0]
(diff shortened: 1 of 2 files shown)


test/e2e/metrics_test.go
```diff
@@ -221,13 +223,40 @@ func getDriverAllocationMetrics(ctx context.Context, cs kubernetes.Interface, no
 		return allocationMetricSnapshot{}, err
 	}
 
-	rawMetrics, err := e2epod.GetMetrics(ctx, cs, dracpuPod, driverMetricsPort)
+	podIP, err := waitForPodIP(ctx, cs, dracpuPod.Name)
```
ffromani: I see no problem with this approach, but why? could you please elaborate in the commit message?
In addition, this commit and the previous one which just fixes the go.mod seems only fixes or legit pivots mid-PR when we realize an approach doesn't work as intended, so consider squashing all 3 in the first commit.

test/e2e/metrics_test.go
```diff
@@ -258,7 +259,7 @@ func getMetricsFromPodIP(podIP string) (string, error) {
 }
 
 func parseAllocationMetricSnapshot(rawMetrics string) (allocationMetricSnapshot, error) {
-	parser := expfmt.TextParser{}
+	parser := expfmt.NewTextParser(model.UTF8Validation)
```
ffromani, line 257: likewise this looks like a legit fixup of the base commit; in this case the followup commits don't tell a story but they look rather like necessary adjustments of the core idea, so please squash all in the first commit.

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+	gomega.Expect(daemonSet.Spec.Template.Spec.Containers).ToNot(gomega.BeEmpty(), "no containers in dracpu daemonset")
+	orgDaemonSet := daemonSet.DeepCopy()
+
+	container := &daemonSet.Spec.Template.Spec.Containers[0]
+	cpuDeviceMode, _ := findArgInContainer(container, argCPUDeviceMode)
+	groupBy, _ := findArgInContainer(container, argGroupBy)
+
+	var claimSpec resourcev1.ResourceClaimSpec
+	switch {
+	case cpuDeviceMode == "grouped" && groupBy == "machine":
```
ffromani: nit: do we ever see any more cases? if not, let's just use a plain `if`

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+		reservedCPUs, err := reservedCPUsFromContainer(container)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		availableCPUs, err := discoverAvailableCPUs(ctx, metricsFxt, targetNode.Name, dracpuTesterImage, reservedCPUs)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		if availableCPUs.Size() == 0 {
+			fmt.Fprintf(ginkgo.GinkgoWriter, "skipping metrics after-suite: no allocatable CPUs left on node %q\n", targetNode.Name)
```
ffromani: for multi-line unstructured data, we can and should use raw `GinkgoWriter`. For genuine logs or something that we can meaningfully put into logs (e.g. to make troubleshootiung easier) we should use `metricsFxt.Log`

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+		claimSpec = makeResourceClaimSpecWithOpaqueConfig(1, true, cpuset.New(cpuID).String())
+	default:
+		claimSpec = makeResourceClaimSpec(1, cpuDeviceMode == "grouped")
+	}
+
+	baseline, err := getDriverAllocationMetrics(ctx, metricsFxt.K8SClientset, targetNode.Name)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot get baseline allocation metrics")
+	expectBasicMetricPropertiesNow(baseline)
+	if baseline.AvailableCPUs < 1 {
+		fmt.Fprintf(ginkgo.GinkgoWriter, "skipping metrics after-suite: no available CPUs reported on node %q baseline=%s\n", targetNode.Name, baseline.String())
```
ffromani: likewise, this seems better served by `metricsFxt.Log`. IIRC we can't use `ginkgo.Skip` (best choice) here.

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+		Spec: resourcev1.ResourceClaimTemplateSpec{
+			Spec: claimSpec,
+		},
+	}
+	createdClaimTemplate, err := metricsFxt.K8SClientset.ResourceV1().ResourceClaimTemplates(metricsFxt.Namespace.Name).Create(ctx, &claimTemplate, metav1.CreateOptions{})
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+	testerPod := makeTesterPodWithExclusiveCPUClaim(metricsFxt.Namespace.Name, dracpuTesterImage, createdClaimTemplate.Name, 1, targetNode.Name)
+	createdPod, err := e2epod.CreateSync(ctx, metricsFxt.K8SClientset, testerPod)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
ffromani: my original suggestion was built on the idea that we can run a `ReportAfterSuite` to check the metrics which other tests make change. Creating objects here is legit for the purpose of testing, but it is borderline bending the rules.
IOW, I think that using ReportAfterSuite to check what _other_ tests did is OK if we only read data from the cluster, like `contextual_logging_test.go` wants to do (it never mutates cluster state). If we need, or if it's just better because more comprehensive/reliable to mutate cluster state (e.g. create objects), it's probably better to turn this code into a regular e2e test.

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+
+func expectBasicMetricProperties(g gomega.Gomega, snapshot allocationMetricSnapshot) {
+	g.Expect(snapshot.AllocatedCPUs).To(gomega.BeNumerically(">=", 0), "allocated CPUs must be non-negative")
+	g.Expect(snapshot.AvailableCPUs).To(gomega.BeNumerically(">=", 0), "available CPUs must be non-negative")
+	g.Expect(snapshot.ReservedCPUs).To(gomega.BeNumerically(">=", 0), "reserved CPUs must be non-negative")
+	g.Expect(snapshot.ActiveResourceClaims).To(gomega.BeNumerically(">=", 0), "active resource claims must be non-negative")
+	g.Expect(snapshot.TotalCPUs()).To(gomega.BeNumerically(">", 0), "total CPUs derived from metrics must be positive")
+}
+
+func expectBasicMetricPropertiesNow(snapshot allocationMetricSnapshot) {
```
ffromani, line 320: maybe add `GinkgoHelper()` here?
probably unnecessary if we pass a `g gomega.Gomega` parameter.
I'd prefer us to be consistent in the same file

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+	gomega.Expect(snapshot.AllocatedCPUs).To(gomega.BeNumerically(">=", 0), "allocated CPUs must be non-negative")
+	gomega.Expect(snapshot.AvailableCPUs).To(gomega.BeNumerically(">=", 0), "available CPUs must be non-negative")
+	gomega.Expect(snapshot.ReservedCPUs).To(gomega.BeNumerically(">=", 0), "reserved CPUs must be non-negative")
+	gomega.Expect(snapshot.ActiveResourceClaims).To(gomega.BeNumerically(">=", 0), "active resource claims must be non-negative")
+	gomega.Expect(snapshot.TotalCPUs()).To(gomega.BeNumerically(">", 0), "total CPUs derived from metrics must be positive")
+}
+
+func excludeNodeFromDriverDaemonSet(ctx context.Context, cs kubernetes.Interface, nodeName string) {
```
ffromani, line 330: same comment as per line 325

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+				terms[i].MatchExpressions = append(terms[i].MatchExpressions, req)
+			}
+		}
+
+		_, err = cs.AppsV1().DaemonSets(daemonSetNamespace).Update(ctx, ds, metav1.UpdateOptions{})
+		g.Expect(err).ToNot(gomega.HaveOccurred())
+	}, driverPodPollTimeout, driverPodPollInterval).Should(gomega.Succeed(), "failed to update DaemonSet affinity")
+}
+
+func restoreDriverDaemonSet(ctx context.Context, cs kubernetes.Interface, orgDaemonSet *appsv1.DaemonSet) {
```
ffromani, line 367: same comment per line 325

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+		if err != nil {
+			return
+		}
+		ds.Spec = orgDaemonSet.Spec
+		_, err = cs.AppsV1().DaemonSets(daemonSetNamespace).Update(ctx, ds, metav1.UpdateOptions{})
+		g.Expect(err).ToNot(gomega.HaveOccurred())
+	}, driverPodPollTimeout, driverPodPollInterval).Should(gomega.Succeed())
+}
+
+func waitForDriverPodTerminationOnNode(ctx context.Context, cs kubernetes.Interface, nodeName string) {
```
ffromani, line 380: same comment per line 325

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+			if pod.Spec.NodeName == nodeName && pod.Status.Phase != v1.PodFailed && pod.Status.Phase != v1.PodSucceeded {
+				terminated = false
+				break
+			}
+		}
+		g.Expect(terminated).To(gomega.BeTrue(), "driver pod on node %q is still running", nodeName)
+	}, driverPodPollTimeout, driverPodPollInterval).Should(gomega.Succeed(), "timed out waiting for pod to terminate")
+}
+
+func waitForDriverPodReadyOnNode(ctx context.Context, cs kubernetes.Interface, nodeName string) {
```
ffromani, line 398: same comment as per line 325

test/e2e/metrics_test.go
```diff
@@ -0,0 +1,418 @@
...
+					ready = true
+					break
+				}
+			}
+			if ready {
+				break
+			}
+		}
+		g.Expect(ready).To(gomega.BeTrue(), "driver pod on node %q is not ready", nodeName)
+	}, driverPodPollTimeout, driverPodPollInterval).Should(gomega.Succeed(), "timed out waiting for pod to become ready")
```
ffromani: please use WithTimeout and WithPolling with new code

ffromani (commented):
thanks for adding this coverage. The base concept LGTM, comments about test code implementation and discussion about how to best integrate this code in the testsuite.

test/e2e/metrics_test.go
AutuSnow: Thanks for the suggestion. The direct PodIP path is intentional: it probes the exact driver pod and avoids the proxy path used in the earlier version. I’ll document this rationale in the commit message and squash the follow-up commits.

ffromani (commented):
In addition, please don't neglect the commit messages. The commit messages should describe the why and the tradeoffs of a commit, acting as micro-design doc to give us context in the future to reprocess code state if conditions changed. We don't need to write essays every time, sometimes a couple line suffice, sometimes empty bodies are still OK, sometimes [we may need more text to explain the change](https://github.com/kubernetes-sigs/dra-driver-cpu/commit/9d3ed5a936ce81823faa337d01e1c4bd68cd8e35)

test/e2e/metrics_test.go
AutuSnow: Thanks for clarifying. Since this check creates workload objects and mutates cluster state, I agree it should be a regular e2e test rather than a `ReportAfterSuite` check. I’ll rework it accordingly.

test/e2e/metrics_test.go
AutuSnow: Thanks, agreed. Since this runs from `ReportAfterSuite`, `ginkgo.Skip` is not available here, so the early return is intentional. I’ll switch the diagnostic output to `metricsFxt.Log`.

test/e2e/metrics_test.go
AutuSnow: Good point. I’ll update the new `Eventually` calls to use `WithTimeout` and `WithPolling`, then squash the follow-up commits into the initial implementation commit.

pravk03:
Looks like all the previous comments are addressed, and I do not have anything to add. 

/lgtm

ffromani (commented):
/approve
/lgtm

thanks for addressing all the comments

PR #242, opened 2026-07-20
feat: generate JSON schema from the Go struct 

### LINKED ISSUE #233: Generate driverConfig JSON schema from the Go struct
Currently driverconfig.Config's fields are manually mirrored in:

- flagToJSONKey map(internal/driverconfig/load.go)
- LogValues() (internal/driverconfig/config.go)
- deployment/helm/dra-driver-cpu/driverconfig.schema.json
- values.schema.json

What we should do as suggested by @ffromani [here](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/213#discussion_r3557202304) is to generate these (or at least the JSON schema) directly from the Config struct's Go types and JSON tags instead of hand-maintaining each copy.

**Why is this needed**:
This drift has already taken place and caused multiple bugs that were caught during #213 review, where `showMetrics` & `sysfsOverlay` were missing from `flagToJSONKey`, from the JSON schema and from LogValues(). Every time a field is added/removed from Config, a contributor must remember to update all locations by hand, with no test enforcement (aside from the internal reflection based tests that's being added in #213). A single source of truth would eliminate this entire bug class.

Requires or should be implemented on top of #213

fmuyassarov:
#### What type of PR is this?
/kind feature

Today and temporarily the driver can be configured two ways: CLI flags and via config file since #213. Both are backed by one Go struct. Since we want the config file's shape to be validated (both by Helm's values.schema.json at install time and conceptually by the driver itself), someone has to describe what the config file accepts as a JSON schema. Before this patch, describing the config file's shape as a JSON schema (`values.schema.json`'s `driverConfig` section) meant manually writing it. That wasn't generated from the Config struct at all. Two other places had the same problem/limitation, `flagToJSONKey` and `LogValues()`. That is a drift trap because you add a field to the struct and can easily (as I did) forget one of the three copies and end up with a silent, untested inconsistency. This exact issue was caught during #213's review and resulted in the need of this improvement. 

With this patch we now reflect the schema from the Go struct with the help of https://github.com/google/jsonschema-go. It's good enough to generate a schema from the `Config` struct, but it can't capture everything:

- Go's `string` type has no notion of "only these N values are legal",  that constraint just isn't part of the type system, so no reflection-based approach could infer it automatically. Some JSON schema libraries allows declaring this via tags but not this one. So `cpuDeviceMode` and `groupBy` need their enum added manually after reflection.
- More importantly, reflection can never tell "this field exists on config struct but must never be settable via a config file". And as you might remember, we already have such fields like `bindAddress`, `exposePCIeRoots` and `showMetrics`. These are not meant to be config-file inputs. For this part, we first reflect the full schema from `Config` and then apply two small override maps (`schemaEnums` and `schemaExcludedFields`) to patch in what reflection can't express and finally stamp `$schema`, `$id`, `$comment` metadata and re-add `apiVersion` (which lives outside `Config` entirely).

You will notice `tools/gen-driverconfig-schema/main.go` and what it does is basically writes driver config JSON schema (generated by `GenerateDriverConfigSchema()`) to a disk so that `driverconfig-schema` makefile targer can consume it when running on CI for example.

Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/233

For a quick test to this in action:
1. I added a fake field in the Config struct 
    ```diff
    diff --git a/internal/driverconfig/config.go b/internal/driverconfig/config.go
    index cb817a3..f1a416b 100644
    --- a/internal/driverconfig/config.go
    +++ b/internal/driverconfig/config.go
    @@ -30,6 +30,7 @@ type Config struct {
            ExposePCIeRoots  bool   `json:"exposePCIeRoots,omitempty"`
            ShowMetrics      bool   `json:"showMetrics,omitempty"`
            SysFSOverlay     string `json:"sysfsOverlay,omitempty"`
    +       FooBar           string `json:"fooBar,omitempty"`
    }

    // LogValues returns key-value pairs for structured logging of the config.
    ```
2.  Then ran the test to double check that we catch such issues
    ```go
    go test ./internal/driverconfig/... -v
    === RUN   TestFlagToJSONKey_CoversAllFlags
    --- PASS: TestFlagToJSONKey_CoversAllFlags (0.00s)
    === RUN   TestFlagToJSONKey_NoStaleEntries
    --- PASS: TestFlagToJSONKey_NoStaleEntries (0.00s)
    === RUN   TestConfigLogValues_CoversAllFields
        logvalues_internal_test.go:43: Config field "FooBar" (json key "fooBar") is missing from LogValues
    --- FAIL: TestConfigLogValues_CoversAllFields (0.00s)
    === RUN   TestGenerateDriverConfigSchema_CoversAllFields
    --- PASS: TestGenerateDriverConfigSchema_CoversAllFields (0.00s)
    === RUN   TestLoad_NoFile
    --- PASS: TestLoad_NoFile (0.00s)
    === RUN   TestLoad_FileOverridesDefaults
    --- PASS: TestLoad_FileOverridesDefaults (0.00s)
    === RUN   TestLoad_CLIFlagWinsOverFile
    --- PASS: TestLoad_CLIFlagWinsOverFile (0.00s)
    === RUN   TestLoad_PartialFile
    --- PASS: TestLoad_PartialFile (0.00s)
    === RUN   TestLoad_FileWithoutAPIVersion
    --- PASS: TestLoad_FileWithoutAPIVersion (0.00s)
    === RUN   TestLoad_UnknownAPIVersionIsError
    --- PASS: TestLoad_UnknownAPIVersionIsError (0.00s)
    === RUN   TestLoad_MissingFileIsError
    --- PASS: TestLoad_MissingFileIsError (0.00s)
    === RUN   TestLoad_UnmappedFlagLogsError
    --- PASS: TestLoad_UnmappedFlagLogsError (0.00s)
    === RUN   TestDefault
    --- PASS: TestDefault (0.00s)
    === RUN   TestLoad_InvalidCPUDeviceModeIsError
    --- PASS: TestLoad_InvalidCPUDeviceModeIsError (0.00s)
    === RUN   TestLoad_InvalidGroupByIsError
    --- PASS: TestLoad_InvalidGroupByIsError (0.00s)
    === RUN   TestLoad_ExcludedFieldInFileIsError
    === RUN   TestLoad_ExcludedFieldInFileIsError/bindAddress
    === RUN   TestLoad_ExcludedFieldInFileIsError/exposePCIeRoots
    === RUN   TestLoad_ExcludedFieldInFileIsError/showMetrics
    --- PASS: TestLoad_ExcludedFieldInFileIsError (0.00s)
        --- PASS: TestLoad_ExcludedFieldInFileIsError/bindAddress (0.00s)
        --- PASS: TestLoad_ExcludedFieldInFileIsError/exposePCIeRoots (0.00s)
        --- PASS: TestLoad_ExcludedFieldInFileIsError/showMetrics (0.00s)
    FAIL
    FAIL	github.com/kubernetes-sigs/dra-driver-cpu/internal/driverconfig	0.652s
    FAIL
    ```

3. Then the next radar is scheme checker target

    ```
    make helm-schema-check
    go run ./tools/gen-driverconfig-schema --out deployment/helm/dra-driver-cpu/driverconfig.schema.json
    go run github.com/losisin/helm-values-schema-json/v2@v2.3.1 \
            -f deployment/helm/dra-driver-cpu/values.yaml \
            -o deployment/helm/dra-driver-cpu/values.schema.json \
            --use-helm-docs \
            --bundle \
            --bundle-root deployment/helm/dra-driver-cpu \
            --indent 2
    Loading file driverconfig.schema.json
    => got 1101B
    JSON schema successfully generated
    diff --git a/deployment/helm/dra-driver-cpu/values.schema.json b/deployment/helm/dra-driver-cpu/values.schema.json
    index 0fae901..94aeeef 100644
    --- a/deployment/helm/dra-driver-cpu/values.schema.json
    +++ b/deployment/helm/dra-driver-cpu/values.schema.json
    @@ -231,6 +231,9 @@
                "individual"
            ]
            },
    +        "fooBar": {
    +          "type": "string"
    +        },
            "groupBy": {
            "type": "string",
            "enum": [
    ERROR: values.schema.json is out of date. Run 'make helm-schema' to update it.
    make: *** [helm-schema-check] Error 1
    ```

4. and finally the docs checker
    ```
    make helm-docs-check
    go run github.com/norwoodj/helm-docs/cmd/helm-docs@v1.14.2 --chart-search-root=deployment/helm
    INFO[2026-07-20T17:30:59+03:00] Found Chart directories [dra-driver-cpu]
    INFO[2026-07-20T17:30:59+03:00] Generating README Documentation for chart deployment/helm/dra-driver-cpu
    diff --git a/deployment/helm/dra-driver-cpu/values.schema.json b/deployment/helm/dra-driver-cpu/values.schema.json
    index 0fae901..94aeeef 100644
    --- a/deployment/helm/dra-driver-cpu/values.schema.json
    +++ b/deployment/helm/dra-driver-cpu/values.schema.json
    @@ -231,6 +231,9 @@
                "individual"
            ]
            },
    +        "fooBar": {
    +          "type": "string"
    +        },
            "groupBy": {
            "type": "string",
            "enum": [
    ERROR: Helm chart README.md is out of date. Run 'make helm-docs' to update it.
    make: *** [helm-docs-check] Error 1
    ```
   

#### Release note

```release-note
NONE
```

#### Documentation updates
- README updates


fmuyassarov:
@ffromani, @pravk03  as promised [here](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/213#discussion_r3557202304), this should hopefully simplify our life in the future when adding/removing fields.

tools/gen-driverconfig-schema/main.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func main() {
+	out := flag.String("out", "deployment/helm/dra-driver-cpu/driverconfig.schema.json", "path to write the generated schema to")
+	flag.Parse()
+
+	if err := run(*out); err != nil {
+		fmt.Fprintln(os.Stderr, "gen-driverconfig-schema:", err)
+		os.Exit(1)
+	}
```
ffromani, line 37: nit/tip: I tend to default to dump to stdout and use shell redirection to create files. Obviously not worth any change because of this comment. Let's take into account just dumping on stdout (or using stdio in general) in the future.

ffromani (commented):
/approve
/lgtm

thanks!

tools/gen-driverconfig-schema/main.go
fmuyassarov: Roger that. I will make sure to update this part too when the next chance comes in the future.

PR #243, opened 2026-07-21
docs: re-structure and split user & dev parts

fmuyassarov:
#### What type of PR is this?
/kind documentation

#### What this PR does / why we need it
Re-structure the growing documentation for easier consumption and division between user & developer. 
#### Which issue(s) this PR is related to
Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/226
#### Release note

```release-note
NONE
```

#### Documentation updates
- README updates
- Helm chart docs

PR #244, opened 2026-07-21
Make the kubelet root directory configurable

### LINKED ISSUE #231: Helm chart hardcodes /var/lib/kubelet; driver fails to register on a relocated kubelet --root-dir
### What happened

On a node whose kubelet runs with a non-default `--root-dir` (e.g. `/mnt/fast/k8s/kubelet`), the `dra-driver-cpu` DaemonSet pod crash-loops:

```
draplugin.go: "Starting" driver="dra.cpu"
nonblockinggrpcserver.go: "GRPC server started" logger="registrar" endpoint="/var/lib/kubelet/plugins_registry/dra.cpu-reg.sock"
app.go: "failed to run" err="driver failed to start: context deadline exceeded"
```

-> `CrashLoopBackOff`. No `dra.cpu` ResourceSlice is ever published, and the kubelet logs show no registration attempt for `dra.cpu`. The NVIDIA GPU DRA driver on the same node registers fine.

### Root cause

The kubelet plugin-watcher only watches `<kubelet-root-dir>/plugins_registry/`. With a relocated `--root-dir` that is **not** `/var/lib/kubelet/plugins_registry/`. But the driver creates its registrar socket under a hardcoded `/var/lib/kubelet/plugins_registry/` (the `kubeletplugin` default), and the chart mounts that path as a literal `hostPath`, so the socket lands in a directory the kubelet never watches, registration never completes, and `kubeletplugin.Start` times out (~30s) -> CrashLoop.

Confirmed on `main` (VERSION `v0.2.0`):
- `deployment/helm/dra-driver-cpu/templates/daemonset.yaml` hardcodes both `/var/lib/kubelet/plugins` and `/var/lib/kubelet/plugins_registry` (literals, not templated).
- `deployment/helm/dra-driver-cpu/values.yaml` exposes no override.
- `cmd/dracpu` defines no kubelet-path flag.

### Why NVIDIA/k8s-dra-driver-gpu is unaffected

It makes these paths configurable via `--kubelet-registrar-directory-path` / `--kubelet-plugins-directory-path` (env `KUBELET_REGISTRAR_DIRECTORY_PATH` / `KUBELET_PLUGINS_DIRECTORY_PATH`, defaulting to `kubeletplugin.KubeletRegistryDir` / `KubeletPluginsDir`), so operators can point them at the real root-dir. `dra-driver-cpu` has no equivalent.

### Environment

- Kubernetes v1.36.1 (kubeadm), kubelet `--root-dir=/mnt/fast/k8s/kubelet`
- dra-driver-cpu v0.2.0 (`oci://registry.k8s.io/dra-driver-cpu/charts/dra-driver-cpu`)
- containerd 2.2.1 (NRI enabled), CPUManager policy `none`

### Reproduction

1. Run kubelet with a non-default `--root-dir` (e.g. `/custom/kubelet`).
2. `helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driver-cpu -n kube-system`
3. The pod CrashLoops with `context deadline exceeded`; `dra.cpu-reg.sock` appears under `/var/lib/kubelet/plugins_registry` instead of `<root-dir>/plugins_registry`.

### Workaround (confirmed)

Repoint just the `plugin-registry` hostPath at the real root-dir (the kubelet dials the advertised `/var/lib/kubelet/plugins/dra.cpu/dra.sock` directly, so the plugins mount can stay):

```
kubectl patch ds dra-driver-cpu -n kube-system --type=strategic -p \
  '{"spec":{"template":{"spec":{"volumes":[{"name":"plugin-registry","hostPath":{"path":"/custom/kubelet/plugins_registry"}}]}}}}'
```

After this the driver registers and publishes a `dra.cpu` ResourceSlice.

### Proposed fix

Mirror the NVIDIA driver: add `--kubelet-registrar-directory-path` / `--kubelet-plugins-directory-path` flags (defaulting to the `kubeletplugin` constants, so current behavior is unchanged) and surface them as Helm values that template the two hostPaths, their mountPaths, and the driver args. Happy to open a PR if that direction sounds good.

thc1006:
#### What type of PR is this?

/kind bug

The driver hardcoded `/var/lib/kubelet`, so it failed to register on a cluster whose kubelet `--root-dir` is relocated: the registration socket landed in a directory the kubelet does not watch.

Following the review, this adds a single `kubeletRootDir` config-file field (via the `driverConfig` mechanism from #213 / #242) rather than CLI flags. The driver derives the registrar and plugins directories from it as `<root>/plugins_registry` and `<root>/plugins`, and keeps the DRA socket in the per-driver `<root>/plugins/<driver name>` subdirectory, matching the kubeletplugin contract that this directory not be shared with other plugins. The default is `/var/lib/kubelet`, so behavior is unchanged unless the root is relocated.

The Helm chart derives the two `hostPath` volumes and their container `mountPath`s from the same `driverConfig.kubeletRootDir` value, cleaned so they match `filepath.Join` on the driver side. An empty value is treated as the default on both sides, so they stay consistent.

Fixes #231

Default behavior is unchanged: with no `driverConfig` set, `helm template` produces the same DaemonSet as before, except that every chart-generated argument is now rendered through `quote` (same argv values), and the binary derives the same `/var/lib/kubelet` paths it used before.

A relative `kubeletRootDir` is rejected at load time; an empty value is defaulted (matching the chart). The kubelet-path derivation and the `extraArgs` `--config` guard live in `_helpers.tpl`; the driver rejects unexpected positional arguments; and config loading no longer logs a spurious error for the `--config` and klog `--v` flags that share the process FlagSet but are not Config fields. The `internal/driverconfig` tests now run under `make test-unit`; `values.schema.json` and the chart `README.md` are regenerated via `make helm-schema` / `make helm-docs`; and `make helm-render-check` asserts the rendered args and kubelet paths.

The Helm E2E workflow now runs two matrix variants: the default-root deployment, and a relocated-root kind cluster where every kubelet uses `--root-dir=/var/lib/custom-kubelet` and the chart is installed with a matching `driverConfig.kubeletRootDir`. The relocated variant verifies ResourceSlice publication and DaemonSet readiness (which only happen if the kubelet watches the relocated `plugins_registry`), the sockets under the relocated root, the absence of the driver socket under the default root, and that no spurious `flagToJSONKey` error appears in the driver logs. This exercises the #231 scenario end to end rather than only the default `/var/lib/kubelet` path.

**Behavior change:** the chart now owns `--config`. Passing `--config` (or `-config`) through `extraArgs` is rejected at template time, because an externally mounted config file could set `kubeletRootDir` independently of the chart's hostPath and mountPath and let them diverge. Users who previously supplied a raw config file through `extraArgs` + `extraVolumes` should move those settings into `driverConfig`.

#### Release note

```release-note
dra-driver-cpu: add the driverConfig.kubeletRootDir configuration field for clusters whose kubelet uses a non-default --root-dir. The registrar, plugins, and per-driver plugin data directories are derived from this root; it defaults to /var/lib/kubelet. The Helm chart now rejects --config in extraArgs; configure the driver through driverConfig instead.
```

#### Documentation updates

- Helm chart `README.md` (regenerated), driverConfig `kubeletRootDir` documented in values.yaml
- `docs/user/configuration.md`: added `kubeletRootDir` (and `sysfsOverlay`) to the config-file schema table, dropped `exposePCIeRoots` (a flag-only field the config file rejects), and noted the per-node heterogeneous-root deployment pattern


thc1006:
Sorry wrong double presse the button

internal/driverconfig/load.go
```diff
@@ -26,56 +26,60 @@ import (
...
-	if err := applyMap(&result, confMap); err != nil {
-		return Config{}, fmt.Errorf("applying config file %q: %w", filePath, err)
+		logger.V(6).Info("config: after file", result.LogValues()...)
 	}
 
-	logger.V(6).Info("config: after file", result.LogValues()...)
-
+	// Always validate, so a bad value from a CLI flag (no config file) is
+	// rejected too, not only a bad value from the file.
 	if err := result.Validate(); err != nil {
```
AutuSnow, line 96: This also changes behavior for existing flags: an invalid value like a bad --cpu-device-mode used to slip through when no config file was set, now it fails at startup. I agree it's better, but maybe worth a line in the PR description since it goes beyond the kubelet paths?

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -94,6 +94,13 @@ healthzPath: /healthz
...
 healthzPort: 8080 # @schema type:integer;minimum:1;maximum:65535
 
+# -- Kubelet root directory. Set this to the kubelet's `--root-dir` when it is
+# not the default `/var/lib/kubelet`. The driver's plugin-registration and
+# plugins host paths, their container mount paths, and the matching
+# `--kubelet-registrar-directory-path` / `--kubelet-plugins-directory-path`
+# flags are all derived from it, so they stay in sync.
+kubeletRootDir: /var/lib/kubelet # @schema pattern:^/.+
+
 # -- Driver config file contents. When non-empty, a ConfigMap is created and
```
AutuSnow, line 100: If someone tries to set kubeletRegistrarDirectoryPath here, all they get is the generic additionalProperties schema error. Would it help to mention in this comment that the kubelet paths are only settable via kubeletRootDir?

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -1,3 +1,9 @@
+{{- $kubeletRoot := .Values.kubeletRootDir | trimSuffix "/" -}}
```
AutuSnow: trimSuffix only strips a single trailing slash, so `/var/lib/kubelet//` would render double slashes in the paths. Harmless on Linux and all six spots stay consistent, so probably fine as is — just checking it's intentional?

internal/driverconfig/load.go
thc1006: Good catch, thanks. Added a note to the PR description. You are right that it validates the whole merged config now on the flags-only path, not only the kubelet paths (the enum flags like --cpu-device-mode are already rejected by their flag.Value at parse time, so this mainly matters for the kubelet paths and for a programmatic caller).

deployment/helm/dra-driver-cpu/values.yaml
thc1006: Done, added a line to the comment noting that the two directory paths are set via kubeletRootDir, not individually under driverConfig.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
thc1006: Fair point, thanks for checking. I switched trimSuffix to regexReplaceAll so all trailing slashes are removed, so /var/lib/kubelet// renders cleanly now too.

AutuSnow:
/lgtm
/hold


The code looks good to me. Just a heads-up on timing: we're in the middle of the 1.37 release cycle, so @ffromani  and @pravk03  have limited bandwidth right now. No need to rush anything on your side — let's give the maintainers plenty of time to get to this. Thanks for your patience!

internal/driverconfig/flags.go
```diff
@@ -36,6 +37,8 @@ func (c *Config) AddFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&c.ExposePCIeRoots, "expose-pcie-roots", c.ExposePCIeRoots, "Discover and expose PCIe roots as device attributes. Requires the DRAListTypeAttributes=true Feature Gate in the cluster.")
 	fs.BoolVar(&c.ShowMetrics, "show-metrics", c.ShowMetrics, "Print custom driver metrics metadata as JSON and exit.")
 	fs.StringVar(&c.SysFSOverlay, "sysfs-overlay", c.SysFSOverlay, "Path to a YAML file containing sysfs file overlays.")
+	fs.StringVar(&c.KubeletRegistrarDirectoryPath, "kubelet-registrar-directory-path", c.KubeletRegistrarDirectoryPath, "Directory where the kubelet plugin registration socket is created. Set this to <kubelet --root-dir>/plugins_registry when the kubelet root directory is not the default /var/lib/kubelet. Must be identical inside and outside the driver container.")
+	fs.StringVar(&c.KubeletPluginsDirectoryPath, "kubelet-plugins-directory-path", c.KubeletPluginsDirectoryPath, "Kubelet plugins directory where the driver's DRA socket is created. Set this to <kubelet --root-dir>/plugins when the kubelet root directory is not the default /var/lib/kubelet. Must be identical inside and outside the driver container.")
```
ffromani: Thanks @AutuSnow for giving a review pass. I'm indeed still bandwidth-constrained (and I'm aware of all the work piling up, and I'll be super happy to address it as soon as possible). I want to comment that we need to set a project policy about when adding flags and when not. We implemented config file support exactly to keep flags size under control, and keep adding flags defies the point. @thc1006 did nothing wrong, just following the existing pattern; we need to change the pattern. Other PRs like #238 and #239 and #217 may be impacted by this decision, so it is becoming urgent, like #216. I'll propose a clear direction and I'll post for community review as soon as things calm down. Thanks for the patience and the continued contributions.

Makefile
```diff
@@ -67,7 +67,7 @@ clean: ## clean
 	rm -rf "$(OUT_DIR)/"
 
 test-unit: ## run tests
-	CGO_ENABLED=1 GOTOOLCHAIN=${TOOOLCHAIN_MODE} go test -v -race -count 1 -coverprofile=coverage.out ./pkg/...
+	CGO_ENABLED=1 GOTOOLCHAIN=${TOOOLCHAIN_MODE} go test -v -race -count 1 -coverprofile=coverage.out ./pkg/... ./internal/... ./cmd/...
```
ffromani: testing internal makes sense (but is borderline better in a cleanup/separate PR rather than intermixed here, albeit still acceptable). Testing `cmd` should be not warranted: we should have no logic in `cmd`

internal/driverconfig/flags.go
thc1006: Thanks @ffromani, no rush at all. This makes sense to me: #213 added config file support to keep the flag surface small, and two more flags here work against that. Once you post a direction I am glad to reshape this to match. One option is dropping the CLI flags and keeping the kubelet paths as config-file fields, with the chart deriving them from kubeletRootDir, but I will follow whatever the community review lands on. Happy to keep it consistent with the other PRs when you get to it.

Makefile
thc1006: Agreed on cmd, I will drop `./cmd/...` since the logic should not live there. I added `./internal/...` only so this PR's new driverconfig tests run in CI, and I am happy to split that into a separate cleanup PR if you would rather keep this one focused. I will fold both into the reshape.

internal/driverconfig/flags.go
thc1006: I see #242 landed the schema generation and the flag-only field policy on main, so I will rebase onto that. Before I redo this, I want to make sure I follow the direction you have in mind for the two kubelet paths rather than guess.

They feel like the same class as `bindAddress` / `exposePCIeRoots` / `showMetrics`: deployment concerns the chart always supplies, not something a user should set per-field in the config file. So one option is to drop the two CLI flags, add the paths to `schemaExcludedFields`, and have the chart keep deriving them from `kubeletRootDir`. That keeps the flag surface from growing and fits the new pattern.

Does that match what you were thinking, or would you rather fold the kubelet paths into whatever broader change you are planning for #216 / #217 / #238 / #239 so it all lands together? No rush from my side, happy to wait until after 1.37 and then rework in one clean pass.

pravk03:
@thc1006 Thanks for raising this PR. Sorry for the delay in getting to this. Since we are past the code freeze now, I should have more time to review.

One question - do we need 2 configs `kubelet-registrar-directory-path` and `kubelet-plugins-directory-path`?. I believe they are always under the `kubelet-root-dir`,  so we can do with one config and derive the sub-paths internally?

thc1006:
Thanks @pravk03, I really appreciate you taking the time to look at this, and no worries at all about the timing.

You are right, and it is a nice simplification. The registrar and plugins directories are always `<kubelet-root-dir>/plugins_registry` and `<kubelet-root-dir>/plugins`, so I can take just the root and derive both sub-paths, plus the per-driver socket directory, inside the driver rather than exposing two options. That is cleaner than what I had, and it also keeps the flag surface from growing, which was the concern @ffromani kindly raised earlier in the review.

For the rework I plan to:
- collapse to a single kubelet root dir and derive the sub-paths in Go
- expose it as a config-file field through the #242 schema mechanism instead of a new flag, so it does not add to the flag surface
- rebase onto main, since #242 changed how the driverconfig schema is generated
- keep the chart's existing single `kubeletRootDir` value, which already derives everything

If you would prefer a `--kubelet-root-dir` flag over the config-file field, just let me know and I am happy to adjust. Otherwise I will push the reworked version along these lines. Thanks again to you and @ffromani for the thoughtful reviews.

pravk03:
@thc1006 I missed reading existing discussions before adding my comment. I just saw that you had already proposed single root dir approach in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/244#discussion_r3636496487 before my comment. 

On config file vs flag, I believe config file is the preference now with newer feaures. I see it's already being discussed in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/244#discussion_r3622654691.

ffromani:
> On config file vs flag, I believe config file is the preference now with newer feaures. I see it's already being discussed in [#244 (comment)](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/244#discussion_r3622654691).

Yes, I think we should default to config file now and add flags on top only in exceptional cases. I wanted to land https://github.com/kubernetes-sigs/dra-driver-cpu/pull/243 + add clear (albeit concise) dev guidelines on top, but I guess we can move in parallel and update this work pre-emptively.

thc1006:
Thanks @pravk03, no problem at all, and thanks @ffromani for confirming the direction. I will go ahead and rework this in parallel then: a single kubelet root as a config-file field through the #242 mechanism, with the sub-paths derived internally and rebased onto main. Glad to align with the dev guidelines from #243 once they land.

thc1006:
Pushed the reworked version, rebased onto main. It now uses a single `kubeletRootDir` config-file field instead of the two flags: the driver derives `<root>/plugins_registry` and `<root>/plugins` internally, and keeps the DRA socket in the per-driver `<root>/plugins/<driver>` subdirectory. The chart derives the host paths and mounts from the same `driverConfig.kubeletRootDir` value, so they match what the driver uses. Default behavior is unchanged at `/var/lib/kubelet`. Thanks again @pravk03 and @ffromani for the direction.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -1,3 +1,9 @@
+{{- /* The driver derives the registrar and plugins directories from the kubelet root, so the host paths and mounts below must match. Both come from driverConfig.kubeletRootDir (default /var/lib/kubelet). Each path is cleaned so it matches filepath.Join on the driver side. A relative value is rejected here, mirroring the driver's own validation, so it fails at template time rather than at startup. */ -}}
+{{- $rawKubeletRoot := (.Values.driverConfig).kubeletRootDir | default "" -}}
+{{- if and $rawKubeletRoot (not (isAbs $rawKubeletRoot)) -}}
+{{- fail "driverConfig.kubeletRootDir must be an absolute path" -}}
+{{- end -}}
+{{- $kubeletRoot := $rawKubeletRoot | default "/var/lib/kubelet" -}}
```
pravk03: Misplaced ?. Copyright should be at the top.

pravk03: Also daemonset.yaml may not be the right place for this check

`deployment/helm/dra-driver-cpu/templates/_helpers.tpl` perhaps ?

cc @fmuyassarov

docs/user/configuration.md
```diff
@@ -32,15 +32,16 @@ driver container, and passes `--config` automatically.
...
+| `sysfsOverlay`     | string | *(none)*           | Path to a YAML file with sysfs file overlays (testing/advanced use).                                                                                                                                                                                                                                                                                   |
+| `kubeconfig`       | string | *(none)*           | Path to a kubeconfig file (for out-of-cluster use).                                                                                                                                                                                                                                                                                                    |
+| `kubeletRootDir`   | string | `/var/lib/kubelet` | Kubelet root directory. Must match the node kubelet's `--root-dir`. The driver derives `<root>/plugins_registry` and `<root>/plugins/<driver-name>` from it. Must be an absolute path. Every node the DaemonSet selects uses this one value; for nodes with different kubelet roots, deploy separate releases with mutually exclusive `nodeSelector`s. |
```
pravk03: We can add a link to https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/. Either here, or in README

fmuyassarov:
Sorry for late. I had it on my plan to go through this change but didn't get chance earlier. 
But I'm on this now and will leave my review very soon.

pkg/driver/driver.go
```diff
@@ -42,7 +42,11 @@ import (
 )
 
 const (
-	kubeletPluginPath = "/var/lib/kubelet/plugins"
+	// defaultKubeletRootDir defaults an empty root for direct callers of New
+	// (driverconfig.Load already normalizes empty to the same value for the
+	// production path). Kept equal to driverconfig.DefaultKubeletRootDir, guarded
+	// by TestDefaultKubeletRootDirMatchesConfig.
+	defaultKubeletRootDir = "/var/lib/kubelet"
```
fmuyassarov: nit, nobody actually consumes it as the real default outside its driver package. Can we mayne use `driverconfig.DefaultKubeletRootDir` directly instead?

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -62,30 +62,36 @@ spec:
...
           {{- end }}
+          {{- range $arg := .Values.extraArgs }}
+          {{- if regexMatch "^--?config(=|$)" ($arg | trim) }}
+          {{- fail "extraArgs must not set --config; configure the driver through driverConfig so the chart and driver stay in sync" }}
+          {{- end }}
+          {{- end }}
+          {{- with .Values.extraArgs }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
         image: {{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}
```
fmuyassarov: For the sake of consistency, I would rather express this in JSON schema and keep daemonset.yaml focused on structure rather than embedding validation logic. For example, we could do something like this in the scheme file

```JSON
"extraArgs": {
  "type": "array",
  "items": {
    "type": "string",
    "not": { "pattern": "^--?config(=|$)" }
  }
}
```

Also, let's move the loop into `_helpers.tpl`.

deployment/helm/dra-driver-cpu/README.md
```diff
@@ -41,8 +41,8 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
...
-| driverConfig | object | `{}` | Driver config file contents. When non-empty, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. Fields in `args.*` that are non-empty always take priority over values set here. Example:   driverConfig:     cpuDeviceMode: individual     groupBy: socket     reservedCPUs: "0-3" |
-| extraArgs | list | `[]` | Extra command-line arguments appended to the driver arguments |
+| driverConfig | object | `{}` | Driver config file contents. When non-empty, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. Fields in args.* that are non-empty always take priority over values set here. Set kubeletRootDir under driverConfig to point the driver at a relocated kubelet root directory. It must match the node kubelet root, applies to every node the DaemonSet selects, and the host paths and mounts are derived from it. It defaults to /var/lib/kubelet. Example:   driverConfig:     cpuDeviceMode: individual     groupBy: socket     reservedCPUs: "0-3"     kubeletRootDir: /var/lib/kubelet |
```
fmuyassarov: I think the part that described the kubeletRootDir is good and useful. However, it is one of the many options under config. Meaning, it would be better to include this info somewhere else.

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -96,10 +96,15 @@ healthzPort: 8080 # @schema type:integer;minimum:1;maximum:65535
 
 # -- Driver config file contents. When non-empty, a ConfigMap is created and
 # mounted into the driver container as /etc/dracpu/config.yaml. Fields in
-# `args.*` that are non-empty always take priority over values set here.
+# args.* that are non-empty always take priority over values set here. Set
+# kubeletRootDir under driverConfig to point the driver at a relocated kubelet
+# root directory. It must match the node kubelet root, applies to every node the
+# DaemonSet selects, and the host paths and mounts are derived from it. It
+# defaults to /var/lib/kubelet.
```
fmuyassarov: same here, this `args.* that are non-empty always take priority over values set here.` is just describing the general point and having deeper explanation about only one option seems over emphasized than others. Could we perhaps make this visible somewhere else?

hack/ci/helm-render-check.sh
```diff
@@ -0,0 +1,83 @@
...
+
+CHART="deployment/helm/dra-driver-cpu"
+fail=0
+
+ok() { printf 'PASS: %s\n' "$1"; }
+bad() {
+	printf 'FAIL: %s\n' "$1"
+	fail=1
+}
+
```
fmuyassarov, line 16: Longer term, it might be worth considering/evaluating https://github.com/helm-unittest/helm-unittest for this kind of chart template testing. I think that will give more structured, less regex fragile assertions than a bash script.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
fmuyassarov: +1, I'm on the same opinion on this part too https://github.com/kubernetes-sigs/dra-driver-cpu/pull/244#discussion_r3648220208

thc1006: Thanks both. Moved the kubelet-root validation and path derivation into `_helpers.tpl` (`kubeletRootDir`, `pluginsDir`, `pluginRegistryDir`), so the DaemonSet keeps its copyright header at the top and just calls the helpers for the mount and hostPath values. Rendered output is unchanged for the roots I checked (default, custom, trailing slash, `..`, `/`).

docs/user/configuration.md
thc1006: Done, added the kubelet reference link on the `--root-dir` mention. Happy to add it to the README too if you would prefer it there.

pkg/driver/driver.go
thc1006: Done. Using `driverconfig.DefaultKubeletRootDir` directly now and dropped the local constant, which also let me remove the drift-guard test.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
thc1006: I tried the schema route first, but our generator (helm-values-schema-json) can't attach a per-item `not`/`pattern` to `extraArgs` through its annotations without a `$ref`/`$defs` detour that ends up more involved than the check itself. So I took your second suggestion and moved the loop into `_helpers.tpl` as `validateExtraArgs`, which keeps daemonset.yaml focused on structure. If you would rather have it in the schema, I can add the definition and reference it with `itemRef` instead.

deployment/helm/dra-driver-cpu/README.md
thc1006: Agreed. Trimmed it back to a short mention and moved the kubeletRootDir detail into `docs/user/configuration.md`. The README is generated from the values.yaml comment, so this updates there too.

deployment/helm/dra-driver-cpu/values.yaml
thc1006: Done. Cut the driverConfig comment back to the general point and pointed to `docs/user/configuration.md` for the per-field detail, so kubeletRootDir isn't singled out.

hack/ci/helm-render-check.sh
thc1006: Agree, helm-unittest would give sturdier assertions than regex matching in bash. I kept the bash check for now to lock down the argv-quoting regression; happy to port it to helm-unittest, in this PR or a follow-up, whichever you prefer.

thc1006:
Pushed a few follow-ups from the recent review passes.

- Config loading no longer logs a spurious error on startup. `--config` and klog's `--v` share the process FlagSet but are not Config fields, so every start that read a config file used to log two `flag not found in flagToJSONKey` errors. `Load` now only collects the mapped flags, and `TestFlagToJSONKey_CoversAllFlags` already guarantees at build time that every Config flag is mapped, so the runtime warning was redundant. Regression test: `TestLoad_IgnoresUnrelatedFlags`.
- Added a relocated-root `helm-e2e` variant (matrix). It brings up a kind cluster whose kubelet runs with `--root-dir=/var/lib/custom-kubelet` on every node and installs the chart with a matching `driverConfig.kubeletRootDir`. The DaemonSet reaching Ready and publishing a ResourceSlice is what proves the kubelet watches the relocated `plugins_registry` and the driver registers there, which is the #231 scenario end to end. The DRA registration directory follows `--root-dir` (the hardcoded legacy path is `device-plugins`, which this driver does not use). The default-root variant is unchanged.
- Docs: the "See docs/user/configuration.md" pointer in the chart is now an absolute repository link so it resolves from the packaged chart, the config guide gained a `kubeletRootDir` example, and I fixed a small typo.
- PR description now covers the argument quoting and the `extraArgs --config` behavior change, with a short migration note.

I also filed three follow-ups for pre-existing things I noticed but kept out of this PR: #253 (SIGTERM handling), #254 (surface kubelet registration errors), and #255 (`hostPath` type on the kubelet volumes).

Still glad to reshape the flag surface whenever the config-vs-flags direction lands.

hack/ci/helm-render-check.sh
thc1006: Update on this: the Helm E2E now includes a relocated-root variant. It brings up a kind cluster with every kubelet on `--root-dir=/var/lib/custom-kubelet` and installs the chart with a matching `driverConfig.kubeletRootDir`, then checks ResourceSlice publication, DaemonSet readiness, and the sockets under the relocated root. So the mount/hostPath/socket alignment is now covered by a real cross-layer test, not the render check alone.

I kept the bash render check as a fast smoke test. Moving it to helm-unittest still seems worthwhile, and I'm glad to do that as a separate follow-up whenever you'd prefer.

hack/ci/kind-ci-relocated-root.yaml
```diff
@@ -0,0 +1,59 @@
+# Copyright 2025 The Kubernetes Authors.
+#
```
fmuyassarov, line 2: nit, let's stick to the same format for the licence header: https://github.com/kubernetes-sigs/dra-driver-cpu/blob/abc8302d81343cedc5b9abf8912d5b32994206d1/internal/driverconfig/cfgfile.go#L2

hack/ci/kind-ci-relocated-root.yaml
```diff
@@ -0,0 +1,59 @@
...
+          vmodule: "allocator=6,dynamicresources=6" # structured/allocator.go, DRA scheduler plugin
+    controllerManager:
+        extraArgs:
+          v: "5"
+  - |
+    kind: InitConfiguration
+    nodeRegistration:
+      kubeletExtraArgs:
+        v: "5"
+        root-dir: "/var/lib/custom-kubelet"
```
fmuyassarov, line 41: why not to just add it to the existing kind-ci.yaml?

fmuyassarov (commented):
Thanks for all the work @thc1006 the testing depth here is great.

One thing separate from the flags vs config discussion above. I (and probably it is only me) feel like this PR grew beyond what the title says. The core of it is `kubeletRootDir`, the driver path derivation and the chart hostPath/mountPath changes is great. But it also picked up a few things that are not really about the kubelet root. 

For example, quoting every rendered arg + rejecting positional args in app.go (it is good hardening, but a different concern), `flagToJSONKey` log spam fix (legit bug but unrelated),  rejecting `--config ` in extraArgs is also a behavior change on its own. Also, the second e2e matrix variant for the relocated root cluster, plus widening make test-unit.

None of it is bad work, I just think it would be a lot easier to review (and safer to revert individually if needed) as separate patches rather than one. What do you think about splitting it up? No rush either way since we're still waiting on the broader flags vs config call anyway.

@ffromani @pravk03 thoughts on this?

fmuyassarov:
> Thanks for all the work @thc1006 the testing depth here is great.
> 
> One thing separate from the flags vs config discussion above. I (and probably it is only me) feel like this PR grew beyond what the title says. The core of it is `kubeletRootDir`, the driver path derivation and the chart hostPath/mountPath changes is great. But it also picked up a few things that are not really about the kubelet root.
> 
> For example, quoting every rendered arg + rejecting positional args in app.go (it is good hardening, but a different concern), `flagToJSONKey` log spam fix (legit bug but unrelated), rejecting `--config ` in extraArgs is also a behavior change on its own. Also, the second e2e matrix variant for the relocated root cluster, plus widening make test-unit.
> 
> None of it is bad work, I just think it would be a lot easier to review (and safer to revert individually if needed) as separate patches rather than one. What do you think about splitting it up? No rush either way since we're still waiting on the broader flags vs config call anyway.
> 
> @ffromani @pravk03 thoughts on this?


and one thing to add is that, we are in the middle of moving away from CLI args toward the config file, so there are more changes coming that could affect what and how we would want to test. I'm not against adding thorough tests for new fields in general, it is just it would be easier to follow the same testing format consistently across all fields.

thc1006:
> > Thanks for all the work @thc1006 the testing depth here is great.
> > One thing separate from the flags vs config discussion above. I (and probably it is only me) feel like this PR grew beyond what the title says. The core of it is `kubeletRootDir`, the driver path derivation and the chart hostPath/mountPath changes is great. But it also picked up a few things that are not really about the kubelet root.
> > For example, quoting every rendered arg + rejecting positional args in app.go (it is good hardening, but a different concern), `flagToJSONKey` log spam fix (legit bug but unrelated), rejecting `--config ` in extraArgs is also a behavior change on its own. Also, the second e2e matrix variant for the relocated root cluster, plus widening make test-unit.
> > None of it is bad work, I just think it would be a lot easier to review (and safer to revert individually if needed) as separate patches rather than one. What do you think about splitting it up? No rush either way since we're still waiting on the broader flags vs config call anyway.
> > @ffromani @pravk03 thoughts on this?
> 
> and one thing to add is that, we are in the middle of moving away from CLI args toward the config file, so there are more changes coming that could affect what and how we would want to test. I'm not against adding thorough tests for new fields in general, it is just it would be easier to follow the same testing format consistently across all fields.

OK roger!!!

thc1006:
Thanks @fmuyassarov, that is fair and I agree the PR outgrew its title. It picked those pieces up incrementally across the review rounds, and in hindsight they belong on their own.

I started with the clearest one: the `flagToJSONKey` log spam is a standalone bug on `main` (it fires even without any kubelet-root change), so I sent it as #259.

For the rest (the arg quoting plus positional-arg rejection, the `extraArgs --config` guard, and the relocated-root e2e variant), I am happy to peel each into its own PR the same way. Since the core `kubeletRootDir` shape and the test layout both depend on where the flags-vs-config direction lands, I would rather agree the order with you first than split now and redo it later. Whatever you and @ffromani / @pravk03 prefer works for me, no rush.

On testing: agreed that a consistent format across fields matters more than extra depth on one field. Once the shared format is settled I will move the kubeletRootDir tests onto it instead of keeping the heavier set here.

hack/ci/kind-ci-relocated-root.yaml
thc1006: Good catch. When I split the relocated-root e2e into its own PR I will match the header format from `cfgfile.go` (no year, same wording).

hack/ci/kind-ci-relocated-root.yaml
thc1006: I kept it separate so the existing default-root coverage stays: the workflow runs both `kind-ci.yaml` (default `/var/lib/kubelet`) and this relocated-root config as a matrix, so a regression in either path is caught. Folding `root-dir` into `kind-ci.yaml` would drop the default-root run. I will carry this into the separate relocated-root e2e PR and we can settle matrix-vs-single-file there.

thc1006:
#238 and #260 landing put this in conflict across the driverconfig files. Rather than resolve a large conflict I'd then re-split anyway, I'd prefer to realign the kubeletRootDir pieces onto the new driverconfig structure from #238 and land them as focused PRs (starting with #259), once you've had a chance to weigh in on the split. Happy to just rebase this as-is instead if you'd rather keep it reviewable in one place; let me know which you prefer.

thc1006:
Rebased onto main. The only conflict was in `cmd/dracpu/app.go`: this branch's positional-argument rejection is incompatible with the `introspect` subcommand from #238 (which dispatches on positional args), so I dropped it here, and the old `--show-metrics` handling is gone now that #238 moved it to the subcommand. The kubelet-root pieces themselves merged cleanly.

That is a good illustration of your point about splitting: the arg-hardening really is a separate concern and it collided with the subcommand work, so it does not belong in this branch. So this now carries the kubelet-root change plus the log-spam/test-unit hunks that overlap with #259. If you'd still like it split further I'm happy to peel those out; otherwise it's back to a reviewable state on current main.

ffromani:
@thc1006 here's the promised directional advice https://github.com/kubernetes-sigs/dra-driver-cpu/pull/261

thc1006:
Thanks, this helps a lot. Putting kubeletRootDir through your questions, it matches the third one: it exists only because the driver runs as a container and has to line up with the host kubelet root, the same class as bindAddress and exposePCIeRoots. That was the read I'd guessed earlier.

The one thing that keeps it from being a pure chart concern is the kubeletplugin contract that the registrar and plugin-data paths be identical inside and outside the container. The kubelet watches its own root-dir, so the driver has to register under the real root. A chart mount with a driver-side default would not line up.

My lean is therefore to treat it like bindAddress: the chart owns and supplies it, kept out of the user-facing config-file schema (schemaExcludedFields) rather than a knob users set in the file. That does reverse the config-file-only shape I settled on earlier, so I wanted to check it matches your thinking before redoing it. If it does, I'll rework #244 that way and split the arg-hardening and the relocated-root e2e into their own PRs (the log-spam fix is already up as #259).

thc1006:
Rebased and reworked on current main.

**Direction.** I took #261 as the answer to my question above, so thank you for writing it. `kubeletRootDir` lands in the third case: it exists only because the driver ships as a container that has to line up with the host kubelet. So the chart owns it, and it is in `schemaExcludedFields` next to `bindAddress` and `exposePCIeRoots`. A config file that sets it is now rejected with a pointer to the chart value, rather than silently disagreeing with the mounts the chart already rendered.

That reverses the config-file shape I had before. What forces it is the kubeletplugin contract: the kubelet looks for the registrar socket under its own root, so the chart mount and the value the driver derives its paths from have to be one value. The chart renders `<root>/plugins`, `<root>/plugins_registry` and `--kubelet-root-dir` from that single input, and `hack/ci/helm-render-check.sh` asserts all three agree.

@fmuyassarov this also reads as consistent with #263: the flags you deprecate there are the user-facing knobs, while `bind-address` and `expose-pcie-roots`, the two chart-owned ones, stay. `kubelet-root-dir` is the same kind, so it should not need deprecating later. Please tell me if you read it differently, since you have the clearer picture of where that series is going.

**Split.** Following your review:

- the `flagToJSONKey` log spam went out as #259 and is merged;
- the arg quoting and positional-argument rejection are dropped from this branch. The positional-arg part collided with the `introspect` subcommand from #238 anyway, which rather made your point for you;
- the `extraArgs --config` guard is dropped too, and I am happy to send it separately if you still want it;
- the `test-unit` widening rode along with #259, since that fix depended on it.

That takes the branch from 636 added lines to 530, and `cmd/dracpu/app.go` down to a single line.

I kept the relocated-root e2e variant here rather than splitting it. It is the only thing that actually exercises the bug: without this change the driver never registers on that cluster and the rollout times out. Splitting it would leave the feature with no end-to-end guard. Say the word if you would still rather have it on its own.

**Sequencing.** This touches `internal/driverconfig` and the chart, so it overlaps #263 either way. Whichever of the two you would rather land first works for me, and I will rebase onto the other. Given #263 is the larger reshaping, I would guess it is easier for this one to go first and for #263 to absorb one extra chart value, but you two know that series better than I do.

Locally: `go build`, `go test ./internal/... ./pkg/...`, `gofmt`, `helm lint --strict`, `make helm-render-check` and `make helm-docs` all pass.

thc1006:
One more push since that comment. `pre-commit` was failing on `mdformat` for the configuration doc I touched, so the head is now d96cfb0, and the branch is +522/-44 rather than the 530 I quoted above. Everything else in that comment still holds.

All checks pass now, with `tide` waiting on the labels.

ffromani:
> #238 and #260 landing put this in conflict across the driverconfig files. Rather than resolve a large conflict I'd then re-split anyway, I'd prefer to realign the kubeletRootDir pieces onto the new driverconfig structure from #238 and land them as focused PRs (starting with #259), once you've had a chance to weigh in on the split. Happy to just rebase this as-is instead if you'd rather keep it reviewable in one place; let me know which you prefer.

Please split each major area you worked - I fully agree every change is a good change, the main/only issue is bundling them together. How to split and the order of merges, if there are interdependencies, is up to you, just make sure to post each PR in such a way the dependencies are obvious and documented in the PR description.

Basically, whatever is easier for you, just make your intent well documented so we're all on the same page.

thc1006:
Thank you, that's clear. Splitting it three ways, each on top of the last:

1. The option itself: `internal/driverconfig`, the `pkg/driver` plumbing, and the configuration doc.
2. The Helm chart wiring, which needs 1.
3. The CI proof: the relocated-root kind job and the helm render check, which need 2.

Each description will say what it sits on and what it doesn't, and I'll link them from here as they go up. This one stays open until the first is ready.

ffromani:
> Thanks, this helps a lot. Putting kubeletRootDir through your questions, it matches the third one: it exists only because the driver runs as a container and has to line up with the host kubelet root, the same class as bindAddress and exposePCIeRoots. That was the read I'd guessed earlier.
> 
> The one thing that keeps it from being a pure chart concern is the kubeletplugin contract that the registrar and plugin-data paths be identical inside and outside the container. The kubelet watches its own root-dir, so the driver has to register under the real root. A chart mount with a driver-side default would not line up.
> 
> My lean is therefore to treat it like bindAddress: the chart owns and supplies it, kept out of the user-facing config-file schema (schemaExcludedFields) rather than a knob users set in the file. That does reverse the config-file-only shape I settled on earlier, so I wanted to check it matches your thinking before redoing it. If it does, I'll rework #244 that way and split the arg-hardening and the relocated-root e2e into their own PRs (the log-spam fix is already up as #259).

Sorry for late answer and sorry if this causes churn.
The guidelines needs to be unfortunately coarse grained in certain areas; and we can't make the doc too detailed, as we can't anticipate all the cases. So some iteration is required, and ultimately there's and likely there will always be a component of a judgement call.

There are tunables which are border-line indeed: --bind-address is a good example. Another good example is --expose-pcie-root; perhaps a less trivial but good example is the --sysfs-overlay. These are all deeply behavior-altering feature.

I think the conclusion you draw above is the correct one: these should be command line flags, not mirrored in the config file, and should be kept as flags. I want to add my thought process for the record and for the sake of the openness and community conversation.

First of all, some options are expected to be exercised rarely. For example, tuning reserved-cpus or the grouping mode is expected to be common, almost cluster-specific. I'd be surprised to NOT see customizations here. OTOH, bind-address, socket registrar, sysfs overlay all address very legit and real usecases, but we expect them to be more sporadic. Fully supported, key features, but less common.

The config file should target the common case tunables. The more exceptional a tunable is, the more the UX should signal it and I see the command line flags, which reflects into a manifest change, as a stronger signal.

In addition, there's a legit use case to split the responsabilities: the cluster admin can legitimately care about the daemonset, and defer the configmap carrying the dracpu driver configfile to other, less privileged personas (vs taking care of everything).
In that context we should bind the ablity to change the registrar path to the highest privilege - therefore the DaemonSet manifest, therefore the admin (note this suggest we should disallow sysfs overlay from the config file - but that's another issue).

I want to acknowledge and stress that the tunables we are discussing squarely fall into a grey zone. I'm sure there can be good arguments to _move in the config file_ most if not all of them - I just don't see them strongly now, hence my previous stance documented above. Therefore, we may pivot in the future, as the project evolves, context is added, experience is gained, more opinions weigh in. I think this is fine and actually good project evolution, and it's especially fine in our early stage of the project.

thc1006:
No churn caused: that's the shape #244 already has. `--kubelet-root-dir` is a flag, `schemaExcludedFields` carries it with a pointer to the chart value, and the DaemonSet passes it. The split above still applies as written.

Thank you for putting the reasoning on the record rather than just the ruling. The rarity signal and the privilege split between DaemonSet and ConfigMap are not things I would have read out of the doc on my own.

Happy to open the sysfs-overlay issue if that helps, or leave it with you.

thc1006:
First one is up: #272, the flag and its `schemaExcludedFields` entry, on current main. It depends on nothing.

The chart wiring and the user documentation follow in the second, and the relocated-root e2e in the third. I will link both here.

ffromani:
small addenda: @thc1006 let's postpone the `--sysfs-overlay` thingy, entirely. It depends if and how we will manage to leverage it for the sub-numa allocator which is slowly brewing. It makes sense both ways, so let's try to minimize the overall churn.

thc1006:
Understood, leaving it alone and not opening the issue.

Neither open PR moves it: `sysfsOverlay` stays an ordinary config-file field in #272 and in #273.

thc1006:
Closing this in favour of the three PRs it was split into, now that the last one is open.

- #272, the driver flag and its validation. Approved, and waiting on one hold.
- #274, the chart wiring and the render check.
- #282, the relocated-root e2e run. This is the one that carries `Fixes #231`, which is why I held off closing here until it existed.

Every file this PR touched is covered by those three, plus `pkg/driver/driver_linux_test.go`, which came out of review. Nothing is being dropped.

Thanks to everyone who reviewed it here. The comments on this PR are what shaped the split, and the review threads that were open have been carried across rather than lost.

PR #249, opened 2026-07-23
feat: implement device health reporting via WatchHealthStatus

### LINKED ISSUE #218: report relevant driver failures using the DRAResourceHealth API (KEP 4680)
An [EventRecorder](https://pkg.go.dev/k8s.io/client-go/tools/record#EventRecorder) that is wired into CPUDriver to emit warning level events on failures, for example in PrepareResourceClaims and the NRI CreateContainer hook, etc (to be agreed/discussed).

**Why is this needed**:
Currently when the driver fails to prepare a claim or fails during container creation, the errors are only visible in the driver logs. Users have to dig through driver pod logs to understand why their workload pods aren't starting. 
If we emit those situations via k8s events, we will probably make the life of the driver user easier as the failures will become more visible and easier to catch.

fmuyassarov:
#### What type of PR is this?
/kind feature

This implements device health reporting using the WatchHealthStatus kubelet plugin API added by
k/k [#139477](https://github.com/kubernetes/kubernetes/pull/139477). The idea (as described in #218) was to give users better visibility into failing devices without getting lost in the logs of driver.

Fixes: #218 

#### Release note
```release-note
Add per-device health reporting: the driver now reports device health (Healthy/Unknown) via the kubelet's WatchHealthStatus API, visible in`pod.status.containerStatuses[].allocatedResourcesStatus`. Unhealthy is not reported yet
```

#### Documentation updates
- README updates

### DIFF
--- a/docs/user/workload-requirements.md
+++ b/docs/user/workload-requirements.md
@@ -134,3 +134,11 @@ Kubernetes `status.extendedResourceClaimStatus` is for DRA-backed extended resou
 For example, a Pod that references a CPU `ResourceClaim` explicitly through `containers[].resources.claims` follows this driver's supported path. A Pod that only patches `status.extendedResourceClaimStatus` with `requestMappings[].resourceName: cpu` does not, because `cpu` is a native resource rather than a DRA-backed extended resource.
 
 For integrations that model native `cpu`, use the Kubernetes node-allocatable DRA status path when available instead.
+
+## Device Health Reporting
+
+The driver reports the health of every device it manages to the kubelet over the DRA `WatchHealthStatus` gRPC API, so the kubelet can reflect it in `pod.status.containerStatuses[].allocatedResourcesStatus`.
+
+- `Healthy`: the default state. The driver has not observed any failure for this device.
+- `Unhealthy`: reserved for failures attributable to the device itself. The driver does not currently report this for any condition. A claim that fails to prepare (for example, a CDI spec write error) surfaces as a claim error instead, not a device health change.
+- `Unknown`: reported by the kubelet itself and not the driver, when it stops receiving health updates for a device within its lease window (for example, if the driver process is down).
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -380,12 +380,6 @@ func (cp *CPUDriver) unprepareResourceClaim(logger logr.Logger, claim kubeletplu
 	return nil
 }
 
-// WatchHealthStatus is required by the kubeletplugin.DRAPlugin interface in the 1.37-dev
-// dynamic-resource-allocation library. This driver does not implement device health reporting.
-func (cp *CPUDriver) WatchHealthStatus(ctx context.Context, reports chan<- kubeletplugin.DeviceHealthReport) error {
-	return kubeletplugin.ErrHealthNotSupported
-}
-
 // HandleError is called by the kubelet plugin framework when an error occurs in the background,
 // for example while publishing ResourceSlices.
 func (cp *CPUDriver) HandleError(ctx context.Context, err error, msg string) {
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -899,40 +899,47 @@ func TestPrepareResourceClaimsDoesNotCommitAllocationWhenCDIFails(t *testing.T) 
 	}
 
 	testCases := []struct {
-		name                       string
-		driver                     *CPUDriver
-		claim                      *resourceapi.ResourceClaim
-		expectExistingAllocation   bool
-		expectedExistingAllocation cpuset.CPUSet
-		expectedSharedCPUs         cpuset.CPUSet
+		name               string
+		driver             *CPUDriver
+		claim              *resourceapi.ResourceClaim
+		expectAllocation   bool
+		expectedAllocation cpuset.CPUSet
+		expectedSharedCPUs cpuset.CPUSet
 	}{
 		{
+			// New claim: the reservation is removed again on failure so the
+			// CPUs go back to the shared pool immediately instead of being
+			// left stuck for a future Unprepare to find.
 			name:               "individual mode new claim",
 			driver:             individualDriver(false),
 			claim:              individualClaim("cpudev2", "cpudev3"),
+			expectAllocation:   false,
 			expectedSharedCPUs: cpuset.New(0, 1, 2, 3),
 		},
 		{
-			name:                       "individual mode existing claim",
-			driver:                     individualDriver(true),
-			claim:                      individualClaim("cpudev0", "cpudev1"),
-			expectExistingAllocation:   true,
-			expectedExistingAllocation: existingCPUs,
-			expectedSharedCPUs:         cpuset.New(2, 3),
+			// Existing claim: prepare reuses the already-committed allocation
+			// and never reserves/removes it, so it must remain untouched.
+			name:               "individual mode existing claim",
+			driver:             individualDriver(true),
+			claim:              individualClaim("cpudev0", "cpudev1"),
+			expectAllocation:   true,
+			expectedAllocation: existingCPUs,
+			expectedSharedCPUs: cpuset.New(2, 3),
 		},
 		{
 			name:               "grouped mode new claim",
 			driver:             groupedDriver(false),
 			claim:              groupedClaim(),
+			expectAllocation:   false,
 			expectedSharedCPUs: cpuset.New(0, 1, 2, 3),
 		},
 		{
-			name:                       "grouped mode existing claim",
-			driver:                     groupedDriver(true),
-			claim:                      groupedClaim(),
-			expectExistingAllocation:   true,
-			expectedExistingAllocation: existingCPUs,
-			expectedSharedCPUs:         cpuset.New(2, 3),
+			name:               "grouped mode existing claim",
+			driver:             groupedDriver(true),
+			claim:              groupedClaim(),
+			expectAllocation:   true,
+			expectedAllocation: existingCPUs,
+			expectedSharedCPUs: cpuset.New(2, 3),
 		},
 	}
 
@@ -949,9 +956,9 @@ func TestPrepareResourceClaimsDoesNotCommitAllocationWhenCDIFails(t *testing.T) 
 			require.Empty(t, mockCdiMgr.devices)
 
 			gotCPUs, ok := tc.driver.cpuAllocationStore.GetResourceClaimAllocation(claimUID)
-			require.Equal(t, tc.expectExistingAllocation, ok)
-			if tc.expectExistingAllocation {
-				require.True(t, tc.expectedExistingAllocation.Equals(gotCPUs), "claim cpus: got %s, want %s", gotCPUs, tc.expectedExistingAllocation)
+			require.Equal(t, tc.expectAllocation, ok)
+			if tc.expectAllocation {
+				require.True(t, tc.expectedAllocation.Equals(gotCPUs), "claim cpus: got %s, want %s", gotCPUs, tc.expectedAllocation)
 			}
 			require.True(t, tc.expectedSharedCPUs.Equals(tc.driver.cpuAllocationStore.GetSharedCPUs()), "shared cpus: got %s, want %s", tc.driver.cpuAllocationStore.GetSharedCPUs(), tc.expectedSharedCPUs)
 		})
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -92,8 +92,15 @@ type CPUDriver struct {
 	pcieRootMapper          *store.PCIeRootMapper
 	devicesPerResourceSlice int
 	metrics                 cpumetrics.Recorder
+	health                  healthTracker
 
 	kubeletRootDir string
+}
+
+// deviceHealthEntry is the last known health of a single device.
+type deviceHealthEntry struct {
+	status  kubeletplugin.HealthStatus
+	message string
 }
 
 // deviceTopology holds the CPU topology and device-to-CPU/socket/NUMA
@@ -182,6 +189,7 @@ func New(logger logr.Logger, providers Providers, config *Config) (*CPUDriver, e
 		pcieRootMapper:          store.NewPCIeRootMapper(),
 		devicesPerResourceSlice: config.DevicesPerResourceSlice(),
 		metrics:                 metricsRecorder,
+		health:                  newHealthTracker(),
 		kubeletRootDir:          config.KubeletRootDir,
 	}
 	sfs := providers.EnsureSysFS()
@@ -231,6 +239,14 @@ func New(logger logr.Logger, providers Providers, config *Config) (*CPUDriver, e
 		// Chunk devices into slices of at most devicesPerResourceSlice
 		plugin.topology.deviceSlices = slices.Collect(slices.Chunk(devices, plugin.devicesPerResourceSlice))
 	}
+
+	for _, d := range devices {
+		plugin.health.devices[d.Name] = &deviceHealthEntry{
+			status:  kubeletplugin.HealthStatusHealthy,
+			message: "device initialized",
+		}
+	}
+
 	return plugin, nil
 }
 
@@ -302,6 +318,7 @@ func (cp *CPUDriver) Start(ctx context.Context) (<-chan error, error) {
 		kubeletplugin.RegistrarDirectoryPath(registrarDir(cp.kubeletRootDir)),
 		kubeletplugin.PluginDataDirectoryPath(driverPluginPath),
 		kubeletplugin.EnableDeviceMetadata(true, []schema.GroupVersion{drametadatav1beta1.SchemeGroupVersion}),
+		kubeletplugin.HealthService(true),
 	}
 	d, err := kubeletplugin.Start(ctx, cp, kubeletOpts...)
 	if err != nil {
@@ -338,11 +355,19 @@ func (cp *CPUDriver) Start(ctx context.Context) (<-chan error, error) {
 	// publish available resources
 	go cp.PublishResources(ctx)
 
+	// periodically (every healthResendInterval) resend device health so
+	// the kubelet's lease on it does not expire (see WatchHealthStatus in
+	// health.go)
+	cp.health.wg.Add(1)
+	go cp.healthResendLoop(ctx)
+
 	return asyncErr, nil
 }
 
 // Stop stops the CPUDriver.
 func (cp *CPUDriver) Stop() {
+	close(cp.health.stopCh)
+	cp.health.wg.Wait()
 	cp.nriPlugin.Stop()
 	cp.draPlugin.Stop()
 }
--- a/pkg/driver/health.go
+++ b/pkg/driver/health.go
@@ -1,0 +1,213 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/kubernetes-sigs/dra-driver-cpu/internal/ctxlog"
+	"k8s.io/dynamic-resource-allocation/kubeletplugin"
+)
+
+// healthTracker owns the per-device health state and the set of active
+// WatchHealthStatus subscribers it is streamed to. mu protects devices,
+// clientsMu protects clients.
+type healthTracker struct {
+	mu        sync.RWMutex
+	devices   map[string]*deviceHealthEntry
+	clientsMu sync.RWMutex
+	clients   []chan kubeletplugin.DeviceHealthReport
+	stopCh    chan struct{}
+	wg        sync.WaitGroup
+}
+
+func newHealthTracker() healthTracker {
+	return healthTracker{
+		devices: make(map[string]*deviceHealthEntry),
+		stopCh:  make(chan struct{}),
+	}
+}
+
+// healthResendInterval is how often the driver resends the full device
+// health snapshot to the kubelet. It must be shorter than the kubelet's
+// default defaultKubeletHealthTimeout(30s), otherwise the kubelet will
+// treat devices as stale and report them as unknown between updates.
+var healthResendInterval = 10 * time.Second
+
+// updateDeviceHealth updates a device's health status. Returns true if it changed.
+func (h *healthTracker) updateDeviceHealth(deviceName string, status kubeletplugin.HealthStatus, message string) (changed bool) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	if h.devices == nil {
+		return false
+	}
+	entry, known := h.devices[deviceName]
+	if !known {
+		return false
+	}
+	changed = entry.status != status || entry.message != message
+	entry.status = status
+	entry.message = message
+	return changed
+}
+
+// markDevicesHealth updates the health of every device in deviceNames and
+// notifies subscribers once for the whole batch. Unknown devices are ignored so a
+// caller passing through a claim's allocation results cannot populate the health
+// map with devices from a different driver/pool.
+func (cp *CPUDriver) markDevicesHealth(logger logr.Logger, deviceNames []string, status kubeletplugin.HealthStatus, message string) {
+	anyChanged := false
+	for _, name := range deviceNames {
+		if changed := cp.health.updateDeviceHealth(name, status, message); changed {
+			logger.V(2).Info("device health changed", "device", name, "health", status, "message", message)
+			anyChanged = true
+		}
+	}
+	if anyChanged {
+		cp.notifyHealthClients()
+	}
+}
+
+// buildHealthReport snapshots the health of every device this driver
+// manages. All devices are published under a single pool named after this
+// node (see PublishResources), so PoolName is always cp.nodeName.
+func (cp *CPUDriver) buildHealthReport() kubeletplugin.DeviceHealthReport {
+	h := &cp.health
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+
+	if h.devices == nil {
+		return kubeletplugin.DeviceHealthReport{Devices: []kubeletplugin.DeviceHealth{}}
+	}
+
+	devices := make([]kubeletplugin.DeviceHealth, 0, len(h.devices))
+	for name, entry := range h.devices {
+		devices = append(devices, kubeletplugin.DeviceHealth{
+			PoolName:           cp.nodeName,
+			DeviceName:         name,
+			Health:             entry.status,
+			LastUpdated:        time.Now(),
+			HealthCheckTimeout: 2 * healthResendInterval,
+			Message:            entry.message,
+		})
+	}
+	return kubeletplugin.DeviceHealthReport{Devices: devices}
+}
+
+// notifyHealthClients pushes the current health snapshot to every active
+// WatchHealthStatus subscriber. Sends are best-effort. A subscriber whose
+// channel is full will simply pick up the next periodic resend instead of
+// blocking the caller that triggered the change.
+func (cp *CPUDriver) notifyHealthClients() {
+	report := cp.buildHealthReport()
+
+	h := &cp.health
+	h.clientsMu.RLock()
+	defer h.clientsMu.RUnlock()
+	for _, ch := range h.clients {
+		select {
+		case ch <- report:
+		default:
+		}
+	}
+}
+
+// healthResendLoop periodically resends the full health snapshot so that
+// the kubelet's per-device lease (HealthCheckTimeout) never expires while
+// the driver is healthy and simply has nothing new to report. WatchHealthStatus
+// deliberately does not resend on its own: a wedged driver should decay to
+// Unknown instead of being kept alive artificially.
+func (cp *CPUDriver) healthResendLoop(ctx context.Context) {
+	defer cp.health.wg.Done()
+
+	ticker := time.NewTicker(healthResendInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-cp.health.stopCh:
+			return
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			cp.notifyHealthClients()
+		}
+	}
+}
+
+// healthClientBufferSize buffers health reports per subscriber to absorb
+// short bursts without blocking. If full, newer reports are dropped.
+// It is an arbitrary small size.
+const healthClientBufferSize = 4
+
+// WatchHealthStatus implements kubeletplugin.DRAPlugin. The kubeletplugin
+// helper calls it whenever the kubelet subscribes to device health updates and
+// takes care of translating the reports produced here into DRAResourceHealth
+// gRPC API that the kubelet supports.
+func (cp *CPUDriver) WatchHealthStatus(ctx context.Context, reports chan<- kubeletplugin.DeviceHealthReport) error {
+	logger := ctxlog.FromContext(ctx)
+	logger.Info("started watching device health updates")
+	defer logger.Info("stopped watching device health updates")
+
+	// Buffered so a burst of markDevicesHealth calls doesn't block on a slow
+	// consumer. notifyHealthClients drops reports rather than blocking.
+	clientCh := make(chan kubeletplugin.DeviceHealthReport, healthClientBufferSize)
+	h := &cp.health
+	h.clientsMu.Lock()
+	h.clients = append(h.clients, clientCh)
+	h.clientsMu.Unlock()
+
+	defer func() {
+		h.clientsMu.Lock()
+		defer h.clientsMu.Unlock()
+		for i, ch := range h.clients {
+			if ch == clientCh {
+				h.clients = append(h.clients[:i], h.clients[i+1:]...)
+				break
+			}
+		}
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil
+	case <-h.stopCh:
+		return nil
+	case reports <- cp.buildHealthReport():
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-h.stopCh:
+			return nil
+		case report := <-clientCh:
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-h.stopCh:
+				return nil
+			case reports <- report:
+			}
+		}
+	}
+}
--- a/pkg/driver/health_test.go
+++ b/pkg/driver/health_test.go
@@ -1,0 +1,217 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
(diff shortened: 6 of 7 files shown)


fmuyassarov:
/hold 
due to dependency on the k8s

/cc @ffromani @pravk03

fmuyassarov:
I didn't include some test results from end to end testing yet, but Ì will paste it later on today so that it is also visible how it looks in action.

fmuyassarov:
Did another round of testing and this an example snapshot of the logs from my latest run:


- Driver instance registered and had the kubelet subscribe to its health stream:

  ```text
  I0724 11:44:11.445584       1 health.go:159] "kubelet subscribed to device health updates" logger="dra" driver="dra.cpu" requestID=3 method="/v1alpha1.DRAResourceHealth/NodeWatchResources" opID="0b830785"
  time="2026-07-24T11:44:11Z" level=info msg="Created plugin 00-dra.cpu (dracpu, handles CreateContainer,StopContainer,RemoveContainer)"
  I0724 11:44:11.635469       1 app.go:218] "driver started"
  ```

  Here kubelet discovers the plugin and opens the `NodeWatchResources` stream on startup, without prompting.

- Health report from the Pod status (**happy scenario**):

  ```bash
  kubectl get pod my-app-with-dra-cpu -o json | jq '.status.containerStatuses[].allocatedResourcesStatus'
  ```

  ```json
  [
    {
      "name": "claim:container1-claim",
      "resources": [
        {
          "health": "Healthy",
          "message": "prepared successfully",
          "resourceID": "dra.k8s.io/cpu=claim-10f72968-9ea8-4beb-a54a-26e9a58ac445"
        }
      ]
    }
  ]
  [
    {
      "name": "claim:container2-claim",
      "resources": [
        {
          "health": "Healthy",
          "message": "prepared successfully",
          "resourceID": "dra.k8s.io/cpu=claim-9d2c0443-6d0d-48d8-8468-16beda3a3659"
        }
      ]
    }
  ]
  ```

- Health report from the Pod status (**Unknown** decay. Here the driver process was killed via a bad image):

  ```bash
  kubectl get pod my-app-with-dra-cpu -o json -w | jq '.status.containerStatuses[].allocatedResourcesStatus'
  ```

  ```json
  [
    {
      "name": "claim:container1-claim",
      "resources": [
        {
          "health": "Unknown",
          "resourceID": "dra.k8s.io/cpu=claim-10f72968-9ea8-4beb-a54a-26e9a58ac445"
        }
      ]
    }
  ]
  [
    {
      "name": "claim:container2-claim",
      "resources": [
        {
          "health": "Unknown",
          "resourceID": "dra.k8s.io/cpu=claim-9d2c0443-6d0d-48d8-8468-16beda3a3659"
        }
      ]
    }
  ]
  ```

  No corresponding driver log, which is expected since the driver process wasn't running. This is purely the kubelet's own `HealthCheckTimeout` declaring the stream stale.

  Then the driver recovered:

  ```bash
  kubectl get pod my-app-with-dra-cpu -o json -w | jq '.status.containerStatuses[].allocatedResourcesStatus'
  ```

  ```json
  [
    {
      "name": "claim:container1-claim",
      "resources": [
        {
          "health": "Healthy",
          "message": "device initialized",
          "resourceID": "dra.k8s.io/cpu=claim-10f72968-9ea8-4beb-a54a-26e9a58ac445"
        }
      ]
    }
  ]
  [
    {
      "name": "claim:container2-claim",
      "resources": [
        {
          "health": "Healthy",
          "message": "device initialized",
          "resourceID": "dra.k8s.io/cpu=claim-9d2c0443-6d0d-48d8-8468-16beda3a3659"
        }
      ]
    }
  ]
  ```

- Finally, explicit **Unhealthy**. Here I restarted the driver pod and then ran `crictl stop` on the container to force kubelet to recreate it (same Pod, same claim UID) and re-trigger the NRI `CreateContainer` hook.

  Driver logs:

  ```bash
  kubectl -n kube-system logs "$POD" --all-containers=true --since=1h | grep -i unhealthy
  ```

  ```text
  I0724 11:54:56.612586       1 health.go:55] "device health changed" driver="dra.cpu" opID="fc9f5aee" pod="default/my-app-with-dra-cpu" podUID="c351bfd0-8fee-4674-9aa5-071c1da4d1c4" container="container1" containerID="c1c70729af93482098d178b71c07f2e79dff22bfd202fddb5ca91236e041a88d" device="cpudev000" health="Unhealthy" message="claim \"10f72968-9ea8-4beb-a54a-26e9a58ac445\" is not prepared by this driver"
  I0724 11:54:56.612682       1 health.go:55] "device health changed" driver="dra.cpu" opID="fc9f5aee" pod="default/my-app-with-dra-cpu" podUID="c351bfd0-8fee-4674-9aa5-071c1da4d1c4" container="container1" containerID="c1c70729af93482098d178b71c07f2e79dff22bfd202fddb5ca91236e041a88d" device="cpudev002" health="Unhealthy" message="claim \"10f72968-9ea8-4beb-a54a-26e9a58ac445\" is not prepared by this driver"
  I0724 11:54:56.612738       1 health.go:55] "device health changed" driver="dra.cpu" opID="fc9f5aee" pod="default/my-app-with-dra-cpu" podUID="c351bfd0-8fee-4674-9aa5-071c1da4d1c4" container="container1" containerID="c1c70729af93482098d178b71c07f2e79dff22bfd202fddb5ca91236e041a88d" device="cpudev001" health="Unhealthy" message="claim \"10f72968-9ea8-4beb-a54a-26e9a58ac445\" is not prepared by this driver"
  I0724 11:54:56.612809       1 health.go:55] "device health changed" driver="dra.cpu" opID="fc9f5aee" pod="default/my-app-with-dra-cpu" podUID="c351bfd0-8fee-4674-9aa5-071c1da4d1c4" container="container1" containerID="c1c70729af93482098d178b71c07f2e79dff22bfd202fddb5ca91236e041a88d" device="cpudev003" health="Unhealthy" message="claim \"10f72968-9ea8-4beb-a54a-26e9a58ac445\" is not prepared by this driver"
  ```

  Pod status confirmation:

  ```bash
  kubectl get pod my-app-with-dra-cpu -o json -w | jq '.status.containerStatuses[].allocatedResourcesStatus'
  ```

  ```json
  [
    {
      "name": "claim:container1-claim",
      "resources": [
        {
          "health": "Unhealthy",
          "message": "claim \"10f72968-9ea8-4beb-a54a-26e9a58ac445\" is not prepared by this driver",
          "resourceID": "dra.k8s.io/cpu=claim-10f72968-9ea8-4beb-a54a-26e9a58ac445"
        }
      ]
    }
  ]
  [
    {
      "name": "claim:container2-claim",
      "resources": [
        {
          "health": "Healthy",
          "message": "device initialized",
          "resourceID": "dra.k8s.io/cpu=claim-9d2c0443-6d0d-48d8-8468-16beda3a3659"
        }
      ]
    }
  ]
  ```

fmuyassarov:
added the last commit (5d2d4ca02a6a5c6504897a82084eea802fd79d3e) for the docs part and this follows the new docs structure.

pkg/driver/driver.go
```diff
@@ -90,6 +92,21 @@ type CPUDriver struct {
 	pcieRootMapper          *store.PCIeRootMapper
 	devicesPerResourceSlice int
 	metrics                 cpumetrics.Recorder
+	// Device health tracking (see health.go). healthMu protects deviceHealth
```
pravk03: nit: consider extracting the health fields + methods into their own struct (e.g. healthTracker)

The rest of `CPUDriver struct` could use some sub-grouping too, but that's for a different PR.

pkg/driver/health.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/kubernetes-sigs/dra-driver-cpu/internal/ctxlog"
+	resourceapi "k8s.io/api/resource/v1"
+	"k8s.io/dynamic-resource-allocation/kubeletplugin"
+	"k8s.io/utils/cpuset"
+)
+
+// healthResendInterval is how often the driver resends the full device
+// health snapshot to the kubelet. It must be shorter than the kubelet'sq
```
pravk03: nit: typo 
kubelet's

pkg/driver/health.go
```diff
@@ -0,0 +1,204 @@
...
+	cp.healthMu.Lock()
+	entry, known := cp.deviceHealth[deviceName]
+	if !known {
+		cp.healthMu.Unlock()
+		return
+	}
+	changed := entry.status != status || entry.message != message
+	entry.status = status
+	entry.message = message
+	cp.healthMu.Unlock()
```
pravk03: nit: Consider moving the locked sections to separate helpers so that we can use `defer cp.healthMu.Unlock()` instead of manual unlock in different paths.

pkg/driver/health.go
```diff
@@ -0,0 +1,204 @@
...
+// single device can cover many CPUs shared across claims and thus attributing a
+// container level failure to one specific device is not safe and is left as
+// future work.
+func (cp *CPUDriver) markCPUSetDevicesHealth(logger logr.Logger, cpus cpuset.CPUSet, status kubeletplugin.HealthStatus, message string) {
+	if len(cp.cpuIDToDeviceName) == 0 {
+		return
+	}
+	for _, cpuID := range cpus.List() {
+		if name, ok := cp.cpuIDToDeviceName[cpuID]; ok {
+			cp.setDeviceHealth(logger, name, status, message)
```
pravk03: `setDeviceHealth` calls `notifyHealthClients` per CPU, would it have any implications on bigger machines ?. 
Or, is it ok to call notify once after this entire loop ?

pkg/driver/health.go
```diff
@@ -0,0 +1,204 @@
...
+// single device, such as a shared CDI spec write covering the whole claim.
+func (cp *CPUDriver) markClaimDevicesHealth(logger logr.Logger, claim *resourceapi.ResourceClaim, status kubeletplugin.HealthStatus, message string) {
+	if claim.Status.Allocation == nil {
+		return
+	}
+	for _, allocResult := range claim.Status.Allocation.Devices.Results {
+		if allocResult.Driver != cp.driverName {
+			continue
+		}
+		cp.setDeviceHealth(logger, allocResult.Device, status, message)
```
pravk03: same comment as before about calling notifyHealthClients per device. This may be less relevant for grouped mode.

pkg/driver/health.go
```diff
@@ -0,0 +1,204 @@
...
+// gRPC API version the kubelet supports.
+func (cp *CPUDriver) WatchHealthStatus(ctx context.Context, reports chan<- kubeletplugin.DeviceHealthReport) error {
+	// No opID/begin-end flow tag here: this call blocks for the driver's
+	// lifetime, so it would never log a matching "end" (same as healthResendLoop).
+	logger := ctxlog.FromContext(ctx)
+	logger.V(2).Info("started watching device health updates")
+
+	// Buffered so a burst of setDeviceHealth calls doesn't block on a slow
+	// consumer. notifyHealthClients drops reports rather than blocking.
+	clientCh := make(chan kubeletplugin.DeviceHealthReport, 4)
```
pravk03: nit: document for why this is 4 would help

pkg/driver/dra_hooks.go
```diff
@@ -149,6 +149,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		case device.GROUP_BY_SOCKET:
 			socketID, ok := cp.deviceNameToSocketID[alloc.Device]
 			if !ok {
+				cp.setDeviceHealth(logger, alloc.Device, kubeletplugin.HealthStatusUnhealthy, "no valid socket ID found for device")
```
pravk03: IIUC, this would always be a no-op  ?. 

If a device is not present in deviceNameToSocketID, it would also not be present in cp.deviceHealth[deviceName] and we return early.

same comment for GROUP_BY_NUMA_NODE case

pkg/driver/dra_hooks.go
```diff
@@ -347,6 +354,11 @@ func (cp *CPUDriver) unprepareResourceClaim(logger logr.Logger, claim kubeletplu
 	if err := cp.cdiMgr.RemoveDevice(logger, getCDIDeviceName(claim.UID)); err != nil {
 		return err
 	}
+	if cpus, ok := cp.cpuAllocationStore.GetResourceClaimAllocation(claim.UID); ok {
+		// Here claim's devices are free again and thus clear any Unhealthy status left
+		// over from a previous failed prepare/container creation attempt.
+		cp.markCPUSetDevicesHealth(logger, cpus, kubeletplugin.HealthStatusHealthy, "unprepared")
```
pravk03: SHould this be cleared on UnPrepare ?

pkg/driver/nri_hooks.go
```diff
@@ -196,6 +197,7 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
 		for uid, cpus := range claimAllocations {
 			cLogger := logger.WithValues("claimUID", uid)
 			if err := cp.validatePreparedClaimAllocation(uid, cpus); err != nil {
+				cp.markCPUSetDevicesHealth(logger, cpus, kubeletplugin.HealthStatusUnhealthy, fmt.Sprintf("prepared claim allocation validation failed: %s", err))
```
pravk03: A buggy pod that has a stale/unauthorized DRA_CPUSET env var gets rejected here, but those devices can be used by a valid pod/claim. I don't think we should mark them as unhealthy.

pravk03:
Thanks @fmuyassarov for putting this PR together!

For grouped mode, I was thinking we could use CPU online state as the health signal. Today the driver reads `/sys/devices/system/cpu/online` once and advertise the capacity in ResourceSlice.  If CPUs are offline later, the advertised capacity becomes incorrect. Since this information is device-scoped, it maps cleanly to per-device health. We don't need to tie it to this PR, but wanted to get some thoughts.

pravk03:
@fmuyassarov @ffromani another consideration: how much of this should be individual mode specific, given #164 and the pending go/no-go decision on deprecation. Does it make sense to scope down this PR to the building blocks and add mode-specific changes in follow up PRs ?

fmuyassarov:
thanks for the review @pravk03 . For some reason, I lost an email notification about your reviews.
I will reply to your point in a bit.

fmuyassarov:
> Thanks @fmuyassarov for putting this PR together!
> 
> For grouped mode, I was thinking we could use CPU online state as the health signal. Today the driver reads `/sys/devices/system/cpu/online` once and advertise the capacity in ResourceSlice. If CPUs are offline later, the advertised capacity becomes incorrect. Since this information is device-scoped, it maps cleanly to per-device health. We don't need to tie it to this PR, but wanted to get some thoughts.

Sounds like a good idea, +1 from my side. Thanks for the suggestion. But I would rather keep it out of this PR though, because we would need a detection mechanism and decision on whether it should be reported as Unhealthy or a distinct capacity mismatch signal. If that's fine with you, let's land the health plumbing here and I'll take care of it as a follow-up.

fmuyassarov:
> @fmuyassarov @ffromani another consideration: how much of this should be individual mode specific, given #164 and the pending go/no-go decision on deprecation. Does it make sense to scope down this PR to the building blocks and add mode-specific changes in follow up PRs ?


Now you raised a good and tricky question 😄 @pravk03 . So most the changes here are mode agnostic, for example the health tracking core, `WatchHealthStatus` endpoint, NRI hook all apply regardless of mode. 

The only individual mode specific piece is the health wiring in `prepareResourceClaim` which is a fairly small. So I think it does makes sense to split it out and keep this PR to the core plumbing +  grouped mode (WDYT?) and move the individual mode wiring to a small follow-up maybe, once #164 lands one way or the other. This would probably save us from extra work in case individual mode ends up going away.

fmuyassarov:
> > @fmuyassarov @ffromani another consideration: how much of this should be individual mode specific, given #164 and the pending go/no-go decision on deprecation. Does it make sense to scope down this PR to the building blocks and add mode-specific changes in follow up PRs ?
> 
> Now you raised a good and tricky question 😄 @pravk03 . So most the changes here are mode agnostic, for example the health tracking core, `WatchHealthStatus` endpoint, NRI hook all apply regardless of mode.
> 
> The only individual mode specific piece is the health wiring in `prepareResourceClaim` which is a fairly small. So I think it does makes sense to split it out and keep this PR to the core plumbing + grouped mode (WDYT?) and move the individual mode wiring to a small follow-up maybe, once #164 lands one way or the other. This would probably save us from extra work in case individual mode ends up going away.

@ffromani any objections on this? Just want to make sure we all agree with the direction before I rebase and adjust the implementation.

pkg/driver/dra_hooks.go
fmuyassarov: Good catch, this is a dead code indeed

pkg/driver/health.go
fmuyassarov: Definitely, this is problematic indeed. Right now we are making the report per CPU,  so we lock, iterate through all the devices and send it to all subscribers. Eventually, we end up with N (number of cpus) number of separate notifications, N number of full health report snapshots and N channel sends. Fixing it.

pkg/driver/nri_hooks.go
fmuyassarov: Do'h, absolutely

pkg/driver/health.go
fmuyassarov: added, but to be clear that was just an arbitrary number.

pkg/driver/dra_hooks.go
fmuyassarov: do you mean we should be sure why the device was unhealthy and clear up the state only if it was unhealthy due the reasons that we know/expect?

pkg/driver/driver.go
fmuyassarov: filed https://github.com/kubernetes-sigs/dra-driver-cpu/pull/268 patch to take care of the rest

fmuyassarov:
> > @fmuyassarov @ffromani another consideration: how much of this should be individual mode specific, given #164 and the pending go/no-go decision on deprecation. Does it make sense to scope down this PR to the building blocks and add mode-specific changes in follow up PRs ?
> 
> Now you raised a good and tricky question 😄 @pravk03 . So most the changes here are mode agnostic, for example the health tracking core, `WatchHealthStatus` endpoint, NRI hook all apply regardless of mode.
> 
> The only individual mode specific piece is the health wiring in `prepareResourceClaim` which is a fairly small. So I think it does makes sense to split it out and keep this PR to the core plumbing + grouped mode (WDYT?) and move the individual mode wiring to a small follow-up maybe, once #164 lands one way or the other. This would probably save us from extra work in case individual mode ends up going away.

I've now dropped the individual mode related bits @pravk03

fmuyassarov:
@pravk03 please let me know if there is something else you want to see here

pravk03:
Sorry for the delay. I havn't looked through the new changes yet, will do by tomorrow.

fmuyassarov:
> Sorry for the delay. I havn't looked through the new changes yet, will do by tomorrow.

no worries and thanks in advance

pkg/driver/nri_hooks.go
pravk03: I think this got left out ?

pkg/driver/health.go
```diff
@@ -0,0 +1,262 @@
...
+			anyChanged = true
+		}
+	}
+	if anyChanged {
+		cp.notifyHealthClients()
+	}
+}
+
+// markCPUSetDevicesHealth marks the individual CPU devices backing cpus with
+// the given health status. Only meaningful in ungrouped device mode, where
```
pravk03: nit: ungrouped device mode -> individual mode

pkg/driver/health.go
```diff
@@ -0,0 +1,262 @@
...
+// notifies any active WatchHealthStatus callers. deviceName must be a key
+// already known from device enumeration. Unknown devices are ignored so a
+// caller passing through a claim's allocation results cannot populate the
+// health map with devices from a different driver/pool.
+func (cp *CPUDriver) setDeviceHealth(logger logr.Logger, deviceName string, status kubeletplugin.HealthStatus, message string) {
+	if changed := cp.health.updateDeviceHealth(deviceName, status, message); changed {
+		logger.V(2).Info("device health changed", "device", deviceName, "health", status, "message", message)
+		cp.notifyHealthClients()
+	}
+}
```
pravk03: Do we need both `setDeviceHealth` and `markDevicesHealth` ?. I think one of them can be removed.

docs/user/workload-requirements.md
```diff
@@ -79,3 +79,15 @@ Kubernetes `status.extendedResourceClaimStatus` is for DRA-backed extended resou
...
 For integrations that model native `cpu`, use the Kubernetes node-allocatable DRA status path when available instead.
+
+## Device Health Reporting
+
+The driver reports the health of every device it manages to the kubelet over the DRA `WatchHealthStatus` gRPC API, so the kubelet can reflect it in `pod.status.containerStatuses[].allocatedResourcesStatus`.
+
+- **`Healthy`**: the default state. The driver has not observed any failure for this device.
+- **`Unhealthy`**: the driver observed a failure attributable to this device, for example:
+  - writing the CDI spec for a claim failed (marks every device backing that claim)
+  - NRI side validation of a previously prepared claim's CPU allocation failed
```
pravk03: nit: We should remove this if we remove the check in nri_hooks.go

docs/user/workload-requirements.md
```diff
@@ -79,3 +79,15 @@ Kubernetes `status.extendedResourceClaimStatus` is for DRA-backed extended resou
...
 
 For integrations that model native `cpu`, use the Kubernetes node-allocatable DRA status path when available instead.
+
+## Device Health Reporting
+
+The driver reports the health of every device it manages to the kubelet over the DRA `WatchHealthStatus` gRPC API, so the kubelet can reflect it in `pod.status.containerStatuses[].allocatedResourcesStatus`.
+
+- **`Healthy`**: the default state. The driver has not observed any failure for this device.
+- **`Unhealthy`**: the driver observed a failure attributable to this device, for example:
+  - writing the CDI spec for a claim failed (marks every device backing that claim)
```
pravk03: nit: I am still think if this should be marked as an error on the device, or if this is more of a driver issue ?

pkg/driver/health.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	devices := make([]kubeletplugin.DeviceHealth, 0, len(h.devices))
+	for name, entry := range h.devices {
+		devices = append(devices, kubeletplugin.DeviceHealth{
+			PoolName:           cp.nodeName,
+			DeviceName:         name,
+			Health:             entry.status,
+			LastUpdated:        time.Now(),
+			HealthCheckTimeout: 2 * healthResendInterval,
```
pravk03, line 108: The comment earlier says "It must be shorter than the kubelet's
// default defaultKubeletHealthTimeout(30s)," 

Here, we are setting to 40s ?

pkg/driver/dra_hooks.go
pravk03: I think this comment was tagged in the wring line. This was for L260. 
When we mark the device as unhealthy, the claim is not added to
cpuAllocationStore ,  so the unprepare path never marks them "unprepared". Unless a prepare retry eventually succeeds, they will be stuck at "failed to write CDI spec" even after the pod/claim is gone ?

pravk03:
Sorry again for the delay. Thanks @fmuyassarov for making the changes, the rework looks good. I've added a few comments, mostly nits. This looks like a good building block for adding in device-scoped health signals (like the CPU online state) as follow-ups.

pkg/driver/health.go
fmuyassarov: defaultKubeletHealthTimeout is 30 sec indeed but healthResendInterval is set to 10s https://github.com/Nordix/dra-driver-cpu/blob/561562f7b12ca1d03cccacbc0f642be60e052d06/pkg/driver/health.go#L54, so we are setting `HealthCheckTimeout` to 20sec here.

fmuyassarov:
/hold
while addressing some comments and testing my recent changes

fmuyassarov:
/hold cancel
This is ready for review (addressed the previous comments and rebased to get go mod changes as well). If that helps, when the CDI spec fails to be written, the user signal looks something like this

```
$ kubectl describe pod pod-cpu-dra-claim
    DownwardAPI:             true
QoS Class:                   Guaranteed
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason                         Age   From               Message
  ----     ------                         ----  ----               -------
  Normal   Scheduled                      3s    default-scheduler  Successfully assigned default/pod-cpu-dra-claim to dra-driver-cpu-worker2
  Warning  FailedPrepareDynamicResources  3s    kubelet            Failed to prepare dynamic resources: prepare dynamic resources: NodePrepareResources failed for ResourceClaim claim-cpu-capacity-10: failed to write CDI spec "dra.k8s.io-cpu_claim-e8bd0bff-318e-4d2d-a2d4-f831f44eeaa2.json": failed to create Spec file: open /var/run/cdi/spec.716602905.tmp: read-only file system
```

```
$ kubectl logs -n kube-system
I0817 12:14:34.368899       1 health.go:83] "device health changed" logger="dra" driver="dra.cpu" requestID=4 method="/k8s.io.kubelet.pkg.apis.dra.v1.DRAPlugin/NodePrepareResources" opID="6333fd63" claim="default/claim-cpu-capacity-10" claimUID="e8bd0bff-318e-4d2d-a2d4-f831f44eeaa2" device="cpudevnuma000" health="Unhealthy" message="failed to write CDI spec: failed to write CDI spec \"dra.k8s.io-cpu_claim-e8bd0bff-318e-4d2d-a2d4-f831f44eeaa2.json\": failed to create Spec file: open /var/run/cdi/spec.716602905.tmp: read-only file system"
I0817 12:14:34.368997       1 health.go:83] "device health changed" logger="dra" driver="dra.cpu" requestID=4 method="/k8s.io.kubelet.pkg.apis.dra.v1.DRAPlugin/NodePrepareResources" opID="6333fd63" claim="default/claim-cpu-capacity-10" claimUID="e8bd0bff-318e-4d2d-a2d4-f831f44eeaa2" device="cpudevnuma000" health="Healthy" message="prepare failed, CPUs returned to shared pool"
```

pkg/driver/dra_hooks.go
```diff
@@ -273,6 +281,9 @@ func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.Resou
 	deviceName := getCDIDeviceName(claim.UID)
 	envVar := fmt.Sprintf("%s_%s=%s", cdiEnvVarPrefix, claim.UID, claimCPUSet.String())
 	if err := cp.cdiMgr.AddDevice(logger, deviceName, envVar); err != nil {
+		// The CDI spec write is shared by every device backing this claim, so
+		// a failure here affects all of them, not just one device.
+		cp.markClaimDevicesHealth(logger, claim, kubeletplugin.HealthStatusUnhealthy, fmt.Sprintf("failed to write CDI spec: %s", err))
```
pravk03: This gets immediately cleared in the calling function when the CPUs are released back to the shared pool. There may still be value in that, but just noting.

Also, since we mark the device unhealthy here, in grouped mode, would existing pod's status (with [KEP-4680](https://github.com/kubernetes/enhancements/pull/5689)) using the same device also transition Healthy → Unhealthy → Healthy when a CDI creation fails for a new pod ?.

pravk03 (commented):
Thanks for making the changes. The health signal implementation and plumbing parts look great to me. I added one comment on the usage. Should be ok to iterate on it in follow-ups.

/lgtm

pkg/driver/dra_hooks.go
pravk03: This is kind-of related to my question https://github.com/kubernetes-sigs/dra-driver-cpu/pull/249#discussion_r3739594708

Should CDI write error should be treated as device unhealthy ?

fmuyassarov: I think I messed up my tree during the rebase and squashing the commits.  Anyways, I don't believe that we should report the device unhealthy due to CDI write failure. So, I've removed it. That said, I've ended up with a patch that basically adds the skeleton for health reporting. We now support healthy & unknown. Unhealthy exists as a status in the API but there's not much on the driver side yet. I will think about unhealth as follow up.

fmuyassarov:
Thanks for the reviews @pravk03 . I've come to a conclusion that I need to really revisit about unhealthy part, which is the most important one I guess. So, right now it is out of the scope and I will discuss with you and @ffromani about that in the original issue before filing a follow up. So, now the scope is smaller but at least the mechanism is in place.

pravk03:
Thanks @fmuyassarov. Agree, we can build on this solid foundation in subsequent PR's. I am happy with the current shape.

/lgtm 

(Happy to add `approve` but holding off for now in case @ffromani you want to take a pass.)

ffromani:
I'd be glad to review mostly to catch up. I think I'm close to be able to focus again on the project with the intensity it deserves. I'm fine with merging a base and keep iterating. Will get back to this work ASAP.

docs/user/workload-requirements.md
```diff
@@ -134,3 +134,11 @@ Kubernetes `status.extendedResourceClaimStatus` is for DRA-backed extended resou
 For example, a Pod that references a CPU `ResourceClaim` explicitly through `containers[].resources.claims` follows this driver's supported path. A Pod that only patches `status.extendedResourceClaimStatus` with `requestMappings[].resourceName: cpu` does not, because `cpu` is a native resource rather than a DRA-backed extended resource.
 
 For integrations that model native `cpu`, use the Kubernetes node-allocatable DRA status path when available instead.
+
+## Device Health Reporting
+
+The driver reports the health of every device it manages to the kubelet over the DRA `WatchHealthStatus` gRPC API, so the kubelet can reflect it in `pod.status.containerStatuses[].allocatedResourcesStatus`.
+
+- `Healthy`: the default state. The driver has not observed any failure for this device.
+- `Unhealthy`: reserved for failures attributable to the device itself. The driver does not currently report this for any condition. A claim that fails to prepare (for example, a CDI spec write error) surfaces as a claim error instead, not a device health change.
```
ffromani, line 143: (disclosure: I didn't catch up with previous conversation, sorry if redundant) I'm thinking about what could be a legit unhealthy status for a CPU core. I can only think of a core being offlined hotunplugged. 
Linux allows to offline cores via software, see for example https://unix.stackexchange.com/questions/686459/disable-intel-alder-lake-efficiency-cores-on-linux . I think this is so weird and unexpected that could be worth surfacing.

pkg/driver/driver.go
```diff
@@ -338,11 +355,19 @@ func (cp *CPUDriver) Start(ctx context.Context) (<-chan error, error) {
...
+	cp.health.wg.Add(1)
+	go cp.healthResendLoop(ctx)
+
 	return asyncErr, nil
 }
 
 // Stop stops the CPUDriver.
 func (cp *CPUDriver) Stop() {
+	close(cp.health.stopCh)
+	cp.health.wg.Wait()
```
ffromani, line 370: minor: maybe wrap these two into a Stop() method for consistency? non-blocking, can be a followup

pkg/driver/driver.go
```diff
@@ -338,11 +355,19 @@ func (cp *CPUDriver) Start(ctx context.Context) (<-chan error, error) {
 	// publish available resources
 	go cp.PublishResources(ctx)
 
+	// periodically (every healthResendInterval) resend device health so
+	// the kubelet's lease on it does not expire (see WatchHealthStatus in
+	// health.go)
+	cp.health.wg.Add(1)
```
ffromani, line 361: can this counter ever exceeds 1? if not, a WaitGroup is probably not the best sync primitive. Can be fixed as followup though.

pkg/driver/dra_hooks_test.go
```diff
@@ -899,40 +899,47 @@ func TestPrepareResourceClaimsDoesNotCommitAllocationWhenCDIFails(t *testing.T)
...
-		driver                     *CPUDriver
-		claim                      *resourceapi.ResourceClaim
-		expectExistingAllocation   bool
-		expectedExistingAllocation cpuset.CPUSet
-		expectedSharedCPUs         cpuset.CPUSet
+		name               string
+		driver             *CPUDriver
+		claim              *resourceapi.ResourceClaim
+		expectAllocation   bool
+		expectedAllocation cpuset.CPUSet
```
ffromani, line 906: why are we doing these renames in this change, again? a link to the prior conversation is fine

pkg/driver/health.go
```diff
@@ -0,0 +1,213 @@
...
+	if anyChanged {
+		cp.notifyHealthClients()
+	}
+}
+
+// buildHealthReport snapshots the health of every device this driver
+// manages. All devices are published under a single pool named after this
+// node (see PublishResources), so PoolName is always cp.nodeName.
+func (cp *CPUDriver) buildHealthReport() kubeletplugin.DeviceHealthReport {
+	h := &cp.health
```
ffromani, line 93: this seems a great candidate to become a method of `healthTracker`

pkg/driver/health.go
```diff
@@ -0,0 +1,213 @@
...
+		})
+	}
+	return kubeletplugin.DeviceHealthReport{Devices: devices}
+}
+
+// notifyHealthClients pushes the current health snapshot to every active
+// WatchHealthStatus subscriber. Sends are best-effort. A subscriber whose
+// channel is full will simply pick up the next periodic resend instead of
+// blocking the caller that triggered the change.
+func (cp *CPUDriver) notifyHealthClients() {
```
ffromani, line 119: likewise this seems a great candidate to be moved to `healthTracker`

pkg/driver/health.go
```diff
@@ -0,0 +1,213 @@
...
+
+	defer func() {
+		h.clientsMu.Lock()
+		defer h.clientsMu.Unlock()
+		for i, ch := range h.clients {
+			if ch == clientCh {
+				h.clients = append(h.clients[:i], h.clients[i+1:]...)
+				break
+			}
+		}
```
ffromani, line 186: nit: nonblocking: any function in the `slices` package can help here?

pkg/driver/health_test.go
```diff
@@ -0,0 +1,217 @@
...
+	for _, name := range deviceNames {
+		cp.health.devices[name] = &deviceHealthEntry{status: kubeletplugin.HealthStatusHealthy, message: "device initialized"}
+	}
+	return cp
+}
+
+func TestMarkDevicesHealth(t *testing.T) {
+	logger := testr.New(t)
+
+	t.Run("unknown device is ignored", func(t *testing.T) {
```
ffromani, line 46: please let's use subtests (`t.Run`) only for tabular tests or tests which share heavily data ("context" in general programming term). Please prefer individual tests if there's not sharing. Can be fixed as followup

ffromani (commented):
/approve

So this is a nice feature and it was in review from a while. I'm fine merging this work and keep iteraring - including the 0.3.0 timeframe.

I think we should have a quick followup to extract the functionality and move it as much as possible out of `CPUDriver` and inside `healthTracker`. We need to push back against the gravitational pull of `CPUDriver` before it becomes a god-object like we had plenty of sad examples in the kubelet.

Defining the "unhealthy" cpus can probably wait

fmuyassarov:
> /approve
> 
> So this is a nice feature and it was in review from a while. I'm fine merging this work and keep iteraring - including the 0.3.0 timeframe.
> 
> I think we should have a quick followup to extract the functionality and move it as much as possible out of `CPUDriver` and inside `healthTracker`. We need to push back against the gravitational pull of `CPUDriver` before it becomes a god-object like we had plenty of sad examples in the kubelet.
> 
> Defining the "unhealthy" cpus can probably wait

thanks @ffromani for the review. I will address them and file the fix later on today.
regarding the unhealthy, there were some discussion during the review above, but they got buried as part of comments. I will dig them out and bring them for the discussion as follow up. Maybe on the issue first.

pkg/driver/driver.go
fmuyassarov: +1

pkg/driver/driver.go
fmuyassarov: agree, took it out

pkg/driver/dra_hooks_test.go
fmuyassarov: this is actually unintentional and should not be touched. I know we use similar naming in other tests but this has nothing to do with the tests related the change. Reverting it and thanks for catching

pkg/driver/health.go
fmuyassarov: yup, done

pkg/driver/health_test.go
fmuyassarov: sure, took care of it in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/306

PR #247, opened 2026-07-23
Validate grouped CPU capacities for KEP-5075

AutuSnow:
#### What type of PR is this?

/kind bug
/kind documentation

#### What this PR does / why we need it

This driver uses the consumable-capacity model introduced by KEP-5075 for grouped CPU devices. The KEP was merged in kubernetes/enhancements#5104, and the v1.37 update in kubernetes/enhancements#6254 is extending `CapacityRange` and `ValidValues` with fractional quantity semantics.

The DRA allocator preserves exact fractional consumed capacities, while this driver currently converts `dra.cpu/cpu` with `resource.Quantity.Value()`. That conversion rounds away from zero, which can pin one CPU for `500m` or two CPUs for `1500m` and diverge from scheduler capacity accounting.

Because this driver allocates indivisible logical CPUs, this PR rejects fractional, zero, and negative quantities before CPU allocation. It adds regression coverage and documents the positive whole-CPU constraint.

#### Which issue(s) this PR is related to

Fixes #246

KEP: https://github.com/kubernetes/enhancements/issues/5075
Original merged KEP PR: https://github.com/kubernetes/enhancements/pull/5104
v1.37 fractional quantity update (currently open): https://github.com/kubernetes/enhancements/pull/6254

#### Special notes for reviewers

This intentionally does not add fractional CPU pinning. Positive whole CPU quantities retain their existing behavior, while unsupported quantities fail before CPU allocation or CDI device creation.

Validated with `go test ./pkg/driver`.

#### Release note

```release-note
Reject fractional, zero, and negative CPU capacity quantities when preparing grouped-mode claims.
```

#### Documentation updates

- README grouped-mode capacity constraints

PR #251, opened 2026-07-24
Validate grouped duplicate Prepare allocation equivalence

### LINKED ISSUE #250: Validate grouped duplicate Prepare allocation equivalence before reusing CPUs
**What happened**:

In grouped mode, `prepareGroupedResourceClaim` returns early when the claim UID already exists in `cpuAllocationStore`.
﻿
The existing CPU assignment is reused without validating the current allocation result,
﻿
This bypasses validation of:
﻿
- allocated devices and topology boundaries;
- consumed `dra.cpu/cpu` capacity;
- machine-mode opaque cpuset configuration.
﻿
As a result, a duplicate Prepare with a stale, malformed, or different grouped allocation can succeed while continuing to use CPUs selected from the previous allocation.
﻿
Individual mode already compares the current allocation with the stored CPU set and fails when they differ.

**What you expected to happen**:

Before reusing an existing grouped CPU assignment, the driver should verify that the current allocation is equivalent to the stored assignment.
﻿
For socket and NUMA grouping, validation should include the allocated device boundaries and consumed CPU quantities. For machine grouping, the opaque cpuset should match the stored assignment exactly.
﻿
If equivalence cannot be established, Prepare should fail closed instead of regenerating CDI state from the old CPU assignment.

AutuSnow:
#### What type of PR is this?

/kind bug

Grouped mode currently reuses the CPU assignment stored for a claim UID before validating the allocation in the duplicate `PrepareResourceClaims` request. A stale or malformed allocation can therefore succeed and regenerate CDI state for CPUs selected from an earlier allocation.

This PR validates the current grouped allocation before reusing stored CPUs:

- socket and NUMA grouping aggregate consumed CPU capacity per topology boundary and require the stored CPU set to match those boundaries exactly;
- machine grouping parses and validates each opaque cpuset, excludes the current claim from active-allocation conflict checks, and requires the reconstructed cpuset to equal the stored assignment;
- missing, fractional, non-positive, or otherwise malformed duplicate capacities fail closed;
- failed equivalence checks return before CDI state is rewritten and leave the CPU allocation store unchanged.

Equivalent duplicate Prepare calls remain idempotent and still recreate CDI state as before.

Fixes #250

Allocation equivalence is semantic rather than a byte-for-byte comparison of `AllocationResult`: repeated results for the same socket or NUMA device are equivalent when their aggregate capacity matches the stored CPUs in that boundary.

PR #247 adds the same positive whole-capacity checks to the initial grouped Prepare path. This PR keeps the duplicate-Prepare fix independently reviewable and applies those checks only while validating reuse; after #247 merges, the small shared validation can be consolidated during rebase if desired.

Validation performed:

- `make test-unit`
- `make lint`
- `make modernize`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 make build`
- `uvx pre-commit run --all-files` (all hooks pass except the existing darwin `dracputester` no-main build-tag issue)
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 uvx pre-commit run go-build-mod --all-files`

#### Release note

```release-note
Reject mismatched grouped-mode duplicate Prepare calls instead of reusing stale CPU assignments.
```

#### Documentation updates

NONE

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -113,6 +113,11 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 
 	if existingCPUs, ok := cp.cpuAllocationStore.GetResourceClaimAllocation(claim.UID); ok {
 		logger.V(2).Info("claim already has allocated CPUs in store, reusing assignment", "cpus", existingCPUs.String())
+		if err := cp.validateGroupedAllocationForReuse(logger, claim, existingCPUs); err != nil {
+			return kubeletplugin.PrepareResult{
+				Err: fmt.Errorf("claim %s/%s is already prepared with CPUs %s which do not match the current allocation: %w", claim.Namespace, claim.Name, existingCPUs.String(), err),
+			}
+		}
 		// Even if the claim is already allocated in our in-memory store (which happens when a duplicate prepare
 		// call is invoked without an intermediate unprepare), we must call prepareDevices and return the result back to Kubelet.
 		// If the CDI file is already created on disk, the CDI manager will safely overwrite it with the same configuration.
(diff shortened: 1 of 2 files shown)


AutuSnow:
/cc @GunaKKIBM

I noticed you assigned #250. I put together this implementation and would appreciate your review; happy to coordinate if you already have work in progress.

pravk03:
I do not fully understand the need for the extra checks. Claim status is immutable, so a duplicate Prepare call should always have an identical allocation. IOW, we should never be out of sync with the initial allocation even in grouped mode.

AutuSnow:
I agree with the normal-case invariant: while a claim remains prepared, repeated Prepare calls should contain the same immutable status.allocation. This check is not intended to handle a legitimate change to that allocation.
The case I’m guarding against is a mismatch between the allocation in the Prepare request and the assignment already present in the driver. The latter may be reconstructed by NRI Synchronize from DRA_CPUSET_* state after a driver restart, so it is a separate local state source. Reusing it without validation would bypass the grouped allocation checks and could recreate CDI state from stale or inconsistent state.
I’ll update the tests to seed the existing store directly instead of using two different allocation objects as duplicate Prepare calls, and make clear that this is a fail-closed consistency check rather than a different duplicate-Prepare contract.

pravk03:
> The latter may be reconstructed by NRI Synchronize from DRA_CPUSET_* state after a driver restart, so it is a separate local state source. Reusing it without validation would bypass the grouped allocation checks and could recreate CDI state from stale or inconsistent state.

Should this check be in the `Synchronize()` path in that case , since the driver restart  doesn't re-trigger `PrepareResourceClaims`. We should also be mindful of adding new checks in the NRI functions as they have a default timeout.

AutuSnow:
You’re right, I mixed two things there. A driver restart doesn’t trigger Prepare, so this check doesn’t validate recovery at restart time.
I wouldn’t move it to Synchronize() either. It only has the container/env state, not the claim’s AllocationResult. Getting that would require API access in the NRI path and introduce the timeout concern you mentioned.
I’ll narrow this PR to the duplicate Prepare case: if the store already contains the claim, grouped mode verifies it before recreating CDI, just like individual mode. I’ll adjust the tests and description and leave recovery validation out of this PR.

pravk03:
Ack. It would also help if you could list the exact sequence of events where the stored assignment and the claim's allocation end up different at a duplicate Prepare — that's the part I'm still not clear with the API restrictions that we currently have.

AutuSnow:
Fair question. I traced the current lifecycle again and couldn’t find a supported sequence that produces this mismatch at duplicate Prepare.
In the normal path, duplicate Prepare sees the same allocation. If the claim is deallocated and allocated again, Unprepare should remove the old driver state first. So the tests are currently constructing an inconsistent state rather than reproducing a reachable one.
I’m going to close this for now. If the recovery/checkpoint work introduces a real path to stale state, we can add the validation at that boundary. Thanks for pushing me to verify this properly.

/close

pravk03:
That was what I was thinking but still wanted to confirm. Thanks so much for checking! We can always pick this back up whenever we need to.

PR #252, opened 2026-07-24
codeorg: simplify futher CPUEnumerator

ffromani:
#### What type of PR is this?

/kind cleanup

in commit be6a4d3f6cb472c769a2c1b9f9acd7ea4bd96a55 we introduced the cpuenumerator abstraction to split the CPUDriver god object. This entity was in charge of building the DRA devices.

At time, the attempt followed the natural seams and split a big object along the easiest fracture lines, but the abstraction was still clunky and kinda artificial.

The way we use this code is in the setup flow; this is a data transformation code with no real state - and its state we throw it away anyway. From this awkward split it also stemmed the hard-to-document `MapDeviceNamesToIDs` method.

The best and natural shape for this code is a function. In this commit we simplify the code further dropping the interface and temp objects; we just use closures and functions.

This further refactoring also couples the map and the devices, which is more natural; the code is now equally loosely coupled but more cohesive, better documented and yet with less pressing documentation needs.

N/A

N/A

#### Release note

```release-note
NONE
```

#### Documentation updates

N/A

### DIFF
--- a/pkg/device/builder.go
+++ b/pkg/device/builder.go
@@ -38,63 +38,35 @@ const (
 	CPUDeviceMachineGrouped      = "cpudevmachine"
 )
 
-type CPUEnumerator struct {
-	topo           *cpuinfo.CPUTopology
-	pcieRootMapper *store.PCIeRootMapper
-	deviceInfos    []cpuDeviceInfo
-}
+// DeviceBuilderFunc produces DRA device objects and the device-name-to-device-id mapping.
+// The mapping is needed at claim preparation time to resolve internal devices.
+type DeviceBuilderFunc func(logger logr.Logger) ([]resourceapi.Device, map[string]int)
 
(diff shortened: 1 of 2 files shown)


AutuSnow:
One small follow-up question: since `DeviceBuilderFunc` is selected and invoked immediately in `New`, could `NewDeviceBuilder` and `NewGroupedDeviceBuilder` return `(devices, nameToID)` directly instead of returning a closure? That would remove the function type and the one-shot closure. No correctness concern, just a possible further simplification.

ffromani:
> One small follow-up question: since `DeviceBuilderFunc` is selected and invoked immediately in `New`, could `NewDeviceBuilder` and `NewGroupedDeviceBuilder` return `(devices, nameToID)` directly instead of returning a closure? That would remove the function type and the one-shot closure. No correctness concern, just a possible further simplification.

Good point! https://github.com/kubernetes-sigs/dra-driver-cpu/pull/260

PR #257, opened 2026-07-25
Handle SIGTERM for graceful shutdown

### LINKED ISSUE #253: dracpu does not handle SIGTERM, so normal Pod termination skips graceful cleanup
Found while reviewing #244; filing separately since it is unrelated to the kubelet root work.

`cmd/dracpu/app.go` sets up signal handling with:

```go
signal.Notify(signalCh, os.Interrupt, unix.SIGINT)
```

On Unix `os.Interrupt` is `SIGINT`, so both arguments denote the same signal and `SIGTERM` is not registered. The kubelet and the container runtime send `SIGTERM` first on a normal Pod stop (rollout, node drain, delete), and Go's default action for an un-handled `SIGTERM` is to terminate the process immediately. So `run()` never returns through its normal path, and the graceful shutdown it already implements is skipped: the HTTP `server.Shutdown`, `dracpu.Stop()` (which stops the DRA plugin and closes/removes its registrar and service sockets), and the NRI plugin cleanup.

This is primarily a graceful-shutdown correctness issue rather than a persistent failure — a left-behind socket is cleaned up on the next start — but the common termination path never runs the cleanup that exists.

Separately, `signalCh` is closed without a preceding `signal.Stop`, which the os/signal docs note is unsafe. Switching to `signal.NotifyContext(ctx, os.Interrupt, unix.SIGTERM)` would cover both signals and let the manual channel go away. Happy to send a small PR if that sounds right.

thc1006:
Fixes #253.

The driver registered `signal.Notify(signalCh, os.Interrupt, unix.SIGINT)`; on Unix both arguments denote SIGINT, so SIGTERM was never handled. SIGTERM is the default stop signal for a normal Pod termination (rollout, drain, delete) when no custom stop signal is configured, so the process was killed by the default disposition before the graceful shutdown ran — the DRA plugin stop, its socket cleanup, and the HTTP server shutdown.

This switches to `signal.NotifyContext(ctx, unix.SIGINT, unix.SIGTERM)`, covering interactive interruption and the default Pod stop signal. After either the signal or the async-error path wakes the main loop, the default signal behavior is restored (`stop()`) before graceful cleanup, so a second signal can still force termination if cleanup gets stuck. It also drops the manual signal channel, which was closed without a preceding `signal.Stop`.

Following @ffromani's note in #253 (register SIGINT once, add SIGTERM, move to `NotifyContext`).

No behavior change on SIGINT (Ctrl+C) or on the NRI async-error path: both still cancel the context and run the graceful shutdown. The startup window before the handler is installed is unchanged and out of scope here.

#### Release note

```release-note
dra-driver-cpu: handle SIGTERM so a normal Pod termination runs the driver's graceful shutdown instead of killing the process abruptly.
```

### DIFF
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -180,17 +180,11 @@ func run(logger logr.Logger, cfg driverconfig.Config) error {
 		return fmt.Errorf("can not obtain the node name, use the hostname-override flag if you want to set it to a specific value: %w", err)
 	}
 
-	// trap Ctrl+C and call cancel on the context
+	// Cancel the context on interactive interruption (SIGINT) or the default Pod
+	// stop signal (SIGTERM), so a normal termination runs the graceful shutdown.
 	ctx := ctxlog.NewContext(context.Background(), logger)
-	ctx, cancel := context.WithCancel(ctx)
-
-	// Enable signal handler
-	signalCh := make(chan os.Signal, 2)
-	defer func() {
-		close(signalCh)
-		cancel()
-	}()
-	signal.Notify(signalCh, os.Interrupt, unix.SIGINT)
+	ctx, stop := signal.NotifyContext(ctx, unix.SIGINT, unix.SIGTERM)
+	defer stop()
 
 	driverConfig := driver.Config{
 		DriverName:       driverName,
@@ -220,13 +214,13 @@ func run(logger logr.Logger, cfg driverconfig.Config) error {
 	var fatalErr error
 
 	select {
-	case <-signalCh:
-		logger.Info("exiting", "reason", "received signal")
-		cancel()
 	case <-ctx.Done():
-		logger.Info("exiting", "reason", "context cancelled")
+		// Restore the default signal behavior as soon as we start exiting, so a
+		// second signal can force termination if graceful cleanup gets stuck.
+		stop()
+		logger.Info("exiting", "reason", "received signal")
 	case err := <-asyncErr:
-		cancel()
+		stop()
 		fatalErr = fmt.Errorf("NRI driver error: %w", err)
 	}
 


cmd/dracpu/app.go
```diff
@@ -220,16 +214,16 @@ func run(logger logr.Logger, cfg driverconfig.Config) error {
 	var fatalErr error
 
 	select {
-	case <-signalCh:
-		logger.Info("exiting", "reason", "received signal")
-		cancel()
 	case <-ctx.Done():
```
ffromani, line 217: Please move the stop() call there, duplicating in other select clauses if needed.

Rationale:
reading both https://pkg.go.dev/os/signal#NotifyContext and https://henvic.dev/posts/signal-notify-context/ seems to suggest we should call `stop` here. I think the real message is rather `The stop function releases resources associated with it, so code should call stop as soon as the operations running in this Context complete and signals no longer need to be diverted to the context. ` - for the examples the best way to show this is to call stop() right in the handler of `ctx.Done()`.

In our case "as soon as possible" is legitimately around line 223 _in the current code shape_. Leaving the stop() call to there is nowadays functionally identical, but leaves the code a little more vulnerable for the future - no linter or automation can catch this I believe, and I'm skeptical around gen-AI code reviews, which we can't afford yet anyway.

ffromani (commented):
thanks for the PR. One little change which seems safest and cheap, and we can merge

cmd/dracpu/app.go
thc1006: Done — moved `stop()` into the `ctx.Done()` and async-error handlers, and kept `defer stop()` for the early-return paths. Agreed it reads more clearly right where the context's work is finished, and it isn't something a linter would catch later. Thanks for the pointer.

ffromani (commented):
/approve
/lgtm

thanks!

PR #259, opened 2026-07-27
Do not log an error for flags that are not Config fields

thc1006:
`Load` runs `fs.Visit` over the whole process FlagSet and logs an error for any flag not in `flagToJSONKey`. But `--config` and klog's `--v` share that FlagSet and are not Config fields, so every startup that reads a config file logs two spurious `flag not found in flagToJSONKey` errors:

```
"config: flag not found in flagToJSONKey map; its explicit CLI value may be silently overridden by the config file" flag="config"
"config: flag not found in flagToJSONKey map; ..." flag="v"
```

This collects only the mapped flags and ignores the rest. A genuinely unmapped Config flag is caught by `TestFlagToJSONKey_CoversAllFlags`, so the runtime warning was redundant. The existing `TestLoad_UnmappedFlagLogsError` (which pinned the old behavior) is replaced by `TestLoad_IgnoresUnrelatedFlags`, asserting no error is logged for `--config`/`--v` while a mapped CLI flag still overrides the file and a file-only field is still applied.

The safety argument above depends on `TestFlagToJSONKey_CoversAllFlags` actually running in CI, but it lives under `internal/` and `make test-unit` only ran `./pkg/...`. This adds `./internal/...` to the target so both that coverage test and the new regression test run. Without it, a future Config flag added without a `flagToJSONKey` entry could merge green and have its CLI value silently overridden by the file. That is why the `test-unit` change rides with this fix rather than going separately: it is the guard this fix relies on.

Split out of #244 as a standalone bug fix, per @fmuyassarov's suggestion there to separate the unrelated changes.

#### Release note

```release-note
dra-driver-cpu: stop logging spurious "flag not found in flagToJSONKey" errors on startup when a config file is used.
```


internal/driverconfig/config_test.go
```diff
@@ -170,30 +170,48 @@ func TestLoad_MissingFileIsError(t *testing.T) {
...
 }
 
-// TestLoad_UnmappedFlagLogsError: a flag missing from flagToJSONKey logs an error.
-func TestLoad_UnmappedFlagLogsError(t *testing.T) {
+// TestLoad_IgnoresUnrelatedFlags: flags that are not Config fields (such as
+// --config and the klog --v flag that share the process FlagSet) must not
+// produce any error log, while a mapped flag set on the command line still
+// overrides the file. Regression guard for the spurious "flag not found in
+// flagToJSONKey" errors that used to fire on every startup with a config file.
+func TestLoad_IgnoresUnrelatedFlags(t *testing.T) {
```
ffromani, line 178: Is not obvious to me why changing this test. Is there a strong reason I'm missing?
I think the test you are adding makes a lot of sense, but I'd prefer a new test rather than modifying this one, as it seems to me they are covering different angles:
1. a test (the existing one) covers flags we should have mapped, and we didn't
2. a test (the added one) covers flags which we _should not have mapped_.

internal/driverconfig/load.go
```diff
@@ -47,14 +47,19 @@ func Load(base Config, filePath string, fs *flag.FlagSet, logger logr.Logger) (C
...
 	}
 
+	// Collect the Config-backed flags that were set explicitly on the command
+	// line so the file does not override them. Flags that are not in
+	// flagToJSONKey (for example --config or the klog --v flag) are not Config
+	// fields, so the file cannot override them and they are simply ignored here.
+	// TestFlagToJSONKey_CoversAllFlags verifies in unit tests (run in CI via
+	// make test-unit) that every AddFlags flag has a mapping, so a genuinely
+	// unmapped Config flag is caught there rather than needing a runtime warning
+	// that also fires on unrelated flags.
```
ffromani: Having test coverage is a prerequisite, so this rationale fits in the commit message and/or in the PR description, but it is unnecessary in the code docs

ffromani (commented):
thanks for the change. The concept LGTM, we need some adjustments.
Just checking, is your email in the signed-off-by tag correct? it would be important for proper attribution

internal/driverconfig/load.go
thc1006: Good point, moved it out. The comment now only says what the loop does; the coverage rationale lives in the commit message and the PR description.

internal/driverconfig/config_test.go
thc1006: The reason I replaced it rather than adding alongside: the old test asserts the error *is* logged (`assert.Contains(logs.String(), "unmapped-flag")`), and removing that log is exactly what this PR does, so keeping it would make it fail.

The angle it was really guarding (a flag we should have mapped but didn't) is covered by `TestFlagToJSONKey_CoversAllFlags`. It `VisitAll`s the flags AddFlags registers and errors on any without a `flagToJSONKey` entry, so it catches a missing mapping statically, without depending on the flag showing up on the command line. That is stronger than the runtime warning was, and the `./internal/...` change in this PR is what makes it actually run in CI (it did not before).

So both angles are covered: `CoversAllFlags` for "should have mapped and didn't", and the new `TestLoad_IgnoresUnrelatedFlags` for "should not have mapped". The old test was really pinning the runtime warning that this PR removes.

If you'd still prefer a different split I'm happy to adjust, for example by renaming or rescoping the new test to make the "should not map" angle more explicit.

fmuyassarov (commented):
Thanks for the fix @thc1006. This one is on me, as I was the one who introduced the bug. I'll wait for @ffromani to take another round of review, but it is LGTM for my side.

A somewhat related (and unrelated at the same time) question,  @ffromani are there any guidelines in k/k on what should and shouldn't be included in the release notes?

thc1006:
Thanks @fmuyassarov, and no worries about the origin. Glad it was a quick one to fix, and I appreciate the review. I'll wait for @ffromani's pass.

internal/driverconfig/config_test.go
ffromani: thanks for adding details, now it fully makes sense to me. We're good the way you did.

ffromani:
> A somewhat related (and unrelated at the same time) question, @ffromani are there any guidelines in k/k on what should and shouldn't be included in the release notes?

Nothing obvious I can recall besides what we are already trying to do. Worth a recheck later on.

ffromani:
/approve
/lgtm

thanks and sorry for the review delay. I'm OOO with very limited/sporadic availability this week and the next one.

PR #260, opened 2026-07-28
codeorg: device: builder: remove closure

ffromani:
#### What type of PR is this?

/kind cleanup

Following up 2d9e2e3 we use a closure immediately; we can therefore remove an indirection layer, and a type as we don't need to carry context.

There's further refactoring potential down the road: we can probably unify the device->id map instead of having 3 mutually exclusive variants, and the relevant number of parameters of device builders suggest we are operating at the wrong abstraction layer - topology, reserved, online cpu and the pcie root mapper probably belong together on a system hw view. This API shape hasn't clearly emerged yet, so we abstain from further cleanup yet.

N/A

Suggested in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/252#issuecomment-5104513350

### DIFF
--- a/pkg/device/builder.go
+++ b/pkg/device/builder.go
@@ -38,35 +38,27 @@ const (
 	CPUDeviceMachineGrouped      = "cpudevmachine"
 )
 
-// DeviceBuilderFunc produces DRA device objects and the device-name-to-device-id mapping.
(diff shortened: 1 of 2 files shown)


AutuSnow:
Thanks !! 

/lgtm

Issue #234, opened 2026-07-14
Separate operational-mode flags from tunable knobs (drop showMetrics from driverConfig; move introspection to subcommands)


AutuSnow:
**What would you like to be added**:

A clear rule for what belongs in the driver config file (and the Helm `driverConfig`
map), plus the cleanup that follows from it.

Coming out of the discussion in #213
(https://github.com/kubernetes-sigs/dra-driver-cpu/pull/213#discussion_r3579987453),
the proposed rule is:

- **Tunable knobs** of the current operational mode belong in the config file /
  `driverConfig` (e.g. `cpuDeviceMode`, `groupBy`, `reservedCPUs`,
  `hostnameOverride`, `kubeconfig`, `sysfsOverlay`).
- **Operational-mode switches** — flags that change *what the driver does* rather
  than tune the running driver (e.g. `--show-metrics`, which prints metadata and
  exits) — should stay CLI-only and, longer term, become subcommands, e.g.
  `dracpu introspect metrics` instead of `dracpu --show-metrics`.

Concrete items:

- [ ] **Near term:** drop `showMetrics` from the Helm `driverConfig` schema
  (`deployment/helm/dra-driver-cpu/driverconfig.schema.json` + regenerated
  `values.schema.json`), mirroring how `exposePCIeRoots`/`bindAddress` were already
  excluded in #213. No change to the Go `Config` struct or the CLI flag.
- [ ] **Before 0.3.0:** introduce subcommands to express operational modes
  (`introspect`/`debug`/…), migrate `--show-metrics` accordingly, and handle
  backward compatibility (we're 0.2.0/alpha, but let's avoid breaking users more
  than necessary).
- [ ] Document the mode-vs-knob rule so new flags are classified consistently and
  don't reintroduce the same sharp edge.

Current Helm `driverConfig` surface, for reference:
`apiVersion, kubeconfig, hostnameOverride, reservedCPUs, cpuDeviceMode, groupBy,
showMetrics, sysfsOverlay`.

**Why is this needed**:

After #213, `showMetrics` is still advertised in the Helm `driverConfig` schema, but
the `--show-metrics` handler runs in `main()` *before* `driverconfig.Load()` and reads
`driverFlags.ShowMetrics` (the flag), not the merged `cfg`. So a value set only via
`driverConfig.showMetrics` is loaded into `cfg` (and even logged by `LogValues()`), yet
the print-and-exit path never acts on it — it is silently inert on the config-file
[...]

PR #261, opened 2026-07-29
docs: add dev guidelines about adding configuration

ffromani:
#### What type of PR is this?

/kind documentation

Add the promised documentation guidelines mentioned in #234 

partially 234

This is my first iteration, open for discussion

#### Release note

```release-note
NONE
```

#### Documentation updates

N/A

### DIFF
--- a/docs/dev/configuration.md
+++ b/docs/dev/configuration.md
@@ -1,0 +1,67 @@
+# Adding configuration tunables
+
+The configurable behavior of the DRA CPU driver can be changed using two approaches:
+
+- command line flags
+- configuration file
+
+The preferred way to add or extend the configurable behavior is through the
+configuration file. The reasons why we prefer this approach are:
+
+- user friendliness: adding options quickly makes the command line long and awkward
+- auditable, reviewable artifact: the configuration file can be versioned, compared
+  and verified more easily
+- extensibility: it's usually easier to process configuration files with programs
+  than command line options
+- configuration files are easier to version and evolve compatibly than command
+  line flags.
+
+Configuration values are applied incrementally:
+
+- the driver has compiled in defaults
+- the defaults are overridden by the configuration file, resulting in a merged configuration
+- the explicitly set command line flags override the merged configuration, resulting in the
+  final configuration which will be used.
+
+**NOTE**: only flags that are actually passed on the command line take precedence; the unset
+flags inherit the prior value from the previous incremental layer (or the default).
+
+The current configuration-file first approach has a pretty large consensus, and there is no
+obvious reason to change again in the foreseeable future, but is not written in stone;
+we may revisit this stance in the future should the circumstances change.
+
+Adding command line options is however **NOT** prohibited, but should be approached
+with a healthy dose of skepticism.
+
+If you are trying to add configurable behavior to the driver, a few questions can
+help your decision making process:
+
+- does the behavior need to be configured at all? Can we just change the defaults?
+- is the option more like a feature flag or feature gate or something users may genuinely
+  decide to flip depending on their setup?
+  should you need a feature flag/gate, please get in touch to discuss the best approach.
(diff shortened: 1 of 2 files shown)


ffromani:
/cc @pravk03 @AutuSnow @fmuyassarov 

As promised. folks, open for discussion

docs/dev/configuration.md
```diff
@@ -0,0 +1,64 @@
...
+help your decision making process:
+
+- does the behavior need to be configured at all? Can we just change the defaults?
+- is the option more like a feature flag or feature gate or something users may genuinely
+  decide to flip depending on their setup?
+  should you need a feature flag/gate, please get in touch to discuss the best approach.
+- is your option a tunable knob of the current operational mode? if so it belongs
+  in the configuration file.
+- is your option an operational-mode switch — a flag that changes what the driver does
+  rather than tune the running driver? if so, a command line flag is likely warranted.
```
fmuyassarov: Good that you mentioned this. Do you think we would benefit from adding a note that there are certain external parameters around the driver that we may want to make configurable. In most cases, these knobs don't require any changes to the actual driver flags/code. they are simply Helm settings, essentially things around the driver Pod. If we do want to modify the driver's behavior though, that should preferably go through the configuration. If for some reason (which I doubt), it doesn't fit into the configuration, then driver flags would be the next option. 

I'm preparing depreciation of certain flags in favor of config and adding similar lines https://github.com/kubernetes-sigs/dra-driver-cpu/commit/a46adae9c57fe318e5712490dc7a09d22a354661 .

fmuyassarov (approved):
Thanks @ffromani . 
LGTM. Just a single (non blocking) question below.

docs/dev/configuration.md
fmuyassarov: or maybe we don't mix helm level settings with driver settings...

ffromani: yes, I think it makes sense to address this point, thanks for raising

fmuyassarov (commented):
Thank you.
/lgtm
/hold
adding hold just to avoid merging before others also take a look

docs/dev/configuration.md
```diff
@@ -0,0 +1,67 @@
...
+
+The preferred way to add or extend the configurable behavior is through the
+configuration file. The reasons why we prefer this approach are:
+
+- user friendliness: adding options quickly makes the command line long and awkward
+- auditable, reviewable artifact: the configuration file can be versioned, compared
+  and verified more easily
+- extensibility: it's usually easier to process configuration files with programs
+  than command line options
+- backward and forward compatible: configuration files are easier to version
```
AutuSnow: “Backward and forward compatible” sounds like a guarantee. Could we say configuration files are easier to version and evolve compatibly instead?

docs/dev/configuration.md
```diff
@@ -0,0 +1,67 @@
+# Adding configuration tunables
```
AutuSnow, line 1: Could we link this guideline from the README’s Developer Documentation section so contributors can discover it?

pravk03:
Thank you!! 

/lgtm

docs/dev/configuration.md
ffromani: right to point this out. This is not meant to be a guarantee (we're in that position yet). We position ourselves to be _more likely_ to be, this is the intent. I'll rephrase accordingly.

docs/dev/configuration.md
ffromani: absolutely, adding.

docs/dev/configuration.md
ffromani: after some iterations, adopted your wording, seems like the best and simplest fix.

AutuSnow:
Thank you, I see that all reviewers and comments have been processed, and I think it's time to merge them
/lgtm
/unhold

PR #262, opened 2026-07-29
WIP: RFC: add Introspect config mode

ffromani:
#### What type of PR is this?

/kind feature
/kind cleanup

Add `dracpu introspect config`, enabling users to check which configuration the driver is using, or to create configuration templates

N/A

I believe this small feature has merit, but it seems unavoidable to load the config in its own path, which works but opens the gate for subtle drifts and subtle bugs. Exposing something different from reality defies the very point of this feature. Do we still deem it worthy?

I'm using AI assistance for the mechanical work of porting the tests (not for the core logic changes) and for self-review.

#### Release note
```release-note
NONE
```

#### Documentation updates

N/A

### DIFF
--- a/internal/subcommands/subcommands.go
+++ b/internal/subcommands/subcommands.go
@@ -57,7 +57,11 @@ func runIntrospect(args []string, stdout, stderr io.Writer) error {
 	fs := flag.NewFlagSet("dracpu introspect", flag.ContinueOnError)
 	fs.SetOutput(stderr)
 	fs.Usage = func() {
-		fmt.Fprintf(fs.Output(), "Usage: %s <subcommand>\n\nAvailable subcommands:\n  metrics\tPrint JSON metadata for custom dra_cpu_* metrics.\n", fs.Name())
+		fmt.Fprintf(fs.Output(), `Usage: %s <subcommand>
+
+Available subcommands:
+metrics\tPrint JSON metadata for custom dra_cpu_* metrics.
+`, fs.Name())
 	}
 
 	if err := fs.Parse(args); err != nil {
--- a/internal/subcommands/subcommands.go
+++ b/internal/subcommands/subcommands.go
@@ -31,10 +31,9 @@ import (
 
 // Options configures subcommand execution.
 type Options struct {
-	DriverConfig driverconfig.Config
-	Logger       logr.Logger
-	Stdout       io.Writer
-	Stderr       io.Writer
+	Logger logr.Logger
+	Stdout io.Writer
+	Stderr io.Writer
 }
 
 // Run dispatches a dracpu subcommand.
@@ -47,20 +46,21 @@ func Run(args []string, opts Options) error {
 	case "gatherinfo":
 		return gatherinfo.Run(args[1:], gatherinfo.Options{}, opts.Logger)
 	case "introspect":
-		return runIntrospect(args[1:], opts.Stdout, opts.Stderr)
+		return runIntrospect(args[1:], opts)
 	default:
 		return fmt.Errorf("unknown subcommand %q; supported subcommands: gatherinfo, introspect", args[0])
 	}
 }
 
-func runIntrospect(args []string, stdout, stderr io.Writer) error {
+func runIntrospect(args []string, opts Options) error {
 	fs := flag.NewFlagSet("dracpu introspect", flag.ContinueOnError)
-	fs.SetOutput(stderr)
+	fs.SetOutput(opts.Stderr)
 	fs.Usage = func() {
 		fmt.Fprintf(fs.Output(), `Usage: %s <subcommand>
 
 Available subcommands:
 metrics\tPrint JSON metadata for custom dra_cpu_* metrics.
+config \tPrint YAML configuration from the supported sources.
 `, fs.Name())
 	}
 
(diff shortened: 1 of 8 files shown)


ffromani:
e2e failures are a known issue and actually a good thing due the RFC nature of this PR

AutuSnow:
I see that the current PR is in WIP status, but I still want to check the content of this PR in advance

internal/subcommands/subcommands.go
```diff
@@ -99,3 +106,38 @@ func runMetrics(args []string, stdout, stderr io.Writer) error {
...
+	if fs.NArg() > 0 {
+		configFile = fs.Args()[0]
+	}
+
+	cfg, err := driverconfig.Load(opts.DriverConfig, configFile, fs, opts.Logger)
+	if err != nil {
+		return err
+	}
+
+	bindata, err := yaml.Marshal(cfg)
```
AutuSnow: Could we only emit fields supported by the config file? The default output includes bindAddress, so feeding the generated template back into this command fails.

internal/subcommands/subcommands.go
```diff
@@ -49,17 +50,21 @@ func Run(args []string, opts Options) error {
...
+func runIntrospect(args []string, opts Options) error {
 	fs := flag.NewFlagSet("dracpu introspect", flag.ContinueOnError)
-	fs.SetOutput(stderr)
+	fs.SetOutput(opts.Stderr)
 	fs.Usage = func() {
-		fmt.Fprintf(fs.Output(), "Usage: %s <subcommand>\n\nAvailable subcommands:\n  metrics\tPrint JSON metadata for custom dra_cpu_* metrics.\n", fs.Name())
+		fmt.Fprintf(fs.Output(), `Usage: %s <subcommand>
+
+Available subcommands:
+metrics\tPrint JSON metadata for custom dra_cpu_* metrics.
```
AutuSnow, line 62: Could we add config to the introspect help and supported-subcommand errors? They still list only metrics.

internal/subcommands/subcommands.go
```diff
@@ -99,3 +106,38 @@ func runMetrics(args []string, stdout, stderr io.Writer) error {
...
+	}
+	if fs.NArg() > 1 {
+		return fmt.Errorf("config accepts up to 1 positional arguments: %s", strings.Join(fs.Args(), " "))
+	}
+	configFile := ""
+	if fs.NArg() > 0 {
+		configFile = fs.Args()[0]
+	}
+
+	cfg, err := driverconfig.Load(opts.DriverConfig, configFile, fs, opts.Logger)
```
AutuSnow: Could we clarify or address the “effective config” case? This only merges defaults and the file, while root CLI flags cannot be combined with subcommands, so CLI overrides used by the running driver are missing.

AutuSnow (commented):
Thanks for exploring this. I think introspect config is useful, but it currently mixes two goals: generating a reusable config template and reporting the effective runtime config. The former must round-trip through Load, while the latter must include CLI overrides. I left comments on both, plus a small help-text issue. Once these semantics are clarified, the overall direction looks reasonable.

internal/subcommands/subcommands.go
ffromani: yes, this is something I'm thinking about. The main concern is not implementing the filtering - that's fairly trivial, but the fact it exhacerbates the drift between the main flow and this introspect flow. I think this is the main concern for this little feature.

internal/subcommands/subcommands.go
ffromani: yes, good point, thanks.

internal/subcommands/subcommands.go
ffromani: yes, this is another gap I need to figure out

ffromani:
@AutuSnow thanks for your early engagement. Please be aware I'm playing with the code shape trying to assess how would it look like if we want to be able to inspect the effective runtime config rather than the loaded config (you also flagged this concern). I'm keeping my experiments in this PR but the code is highly volatile and highly debatable; I'm not sure myself if we want to keep it, move to separate PRs, postpone or else.

All this work was spun off by my desire to have good coverage for the changes I'm making in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/229 - a e2e_local test would make for a trivial and robust check the flag is working.
But it's probably time to stop the game and timebox this effort.

AutuSnow:
Thanks, this context helps.
﻿
My main concern is still the semantic split between a round-trippable config template and effective runtime config including CLI overrides. Given the code is still exploratory, I think it makes sense to timebox this and either narrow the scope or split the goals before going deeper on review.
﻿
Happy to take another pass once the intended behavior is settled a bit more.

ffromani:
> Thanks, this context helps. ﻿ My main concern is still the semantic split between a round-trippable config template and effective runtime config including CLI overrides. 

Absolutely. The final product MUST be round-trippable. Yet, my ultimate goal is to accurately reflect the running config. If and how we can achieve this is yet to be seen, perhaps incrementally, perhaps adding options to the hypothetic `dracpu introspect config`. Libvirt, for example, does something similar so this approach won't be totally crazy.

I think I found an interesting refactoring while exploring this concept, I'll push as separate PR in the coming days.

Will ping you folks once I have something reviewable. Early review of WIP/drafts is always welcome but please be aware of the highly volatile nature of the code till the WIP/draft is lifted.

ffromani:
/hold

still very WIP, please keep ignoring this PR for the time being

PR #263, opened 2026-07-29
chore: reduce driver CLI flags in favor of config file

### LINKED ISSUE #245: Reduce driver CLI flags in favor of config file
Reduce driver CLI flags in favor of `driverConfig` (config file as primary, not exclusive). Following up on the direction discussed in #213 to reduce the CLI flag surface by migrating flags that only control internal driver behavior to be config file first, deprecating their standalone flags. This is not a plan to eliminate flags entirely.

List of the flags I'm considering to start dropping with:
```
--cpu-device-mode
--group-by
--reserved-cpus
--hostname-override
--sysfs-overlay
```

- `config` - stays explicit. Auto-loading a file if present would be an implicit-config footgun (a stale ConfigMap/hostPath leftover could get picked up unintentionally).
- `kubeconfig`- not tied to which cluster the driver behaves on, just which cluster it talks to. Never set in the in cluster/Helm anyway and every client go tool exposes it as a flag, so I thought of keeping it.
- `kubelet-registrar-directory-path` and `kubelet-plugins-directory-path` from #244 should probably stay as override flags even after `driverConfig` becomes primary. `kubectl patch ds` on the pod spec triggers an automatic rollout; editing a ConfigMap directly doesn't, so it's a real fast-path for exactly the kind of cluster-wide outage #231 describes. `bind-address` can probably tag along for consistency, even if it's a smaller deal by itself.
- `show-metrics` is one shot print and exit action, not persistent state. As such, it probably doesn't belong in a steady state config file at all in my opinion.

My thinking is to approach this on multiple fronts to reduce the side affect on users. That said, we would first provide clear signals that these flags are being deprecated (through logs, docs, changelogs, etc.). Then, during Helm installation/templating, we would automatically convert the CLI flags into the equivalent `driverConfig` fields. Finally, we would remove the flags from the binary as the actual breaking change.

1. Mark these flags as deprecated in their `--help` text, pointing users to the equivalent `driverConfig` key.
2. Add startup warning log message when `fs.Visit` (during Load()) detects that one of these flags was explicitly passed. This introduces no precedence or behavior changes. It basically should increase the visibility.
3. Update the Helm internals so that `.Values.args.cpuDeviceMode` and the other deprecated flags are folded into the generated `driverConfig` instead of being emitted as CLI flags. Existing users' `values.yaml` files should continue to work as before and only the transport mechanism changes under the hood. This should in theory, migrate users away from the individual CLI flags before we make any changes to the binary's flag parsing, with no visible impact to them (_I hope_).
4. When it's time for removal, we start deleting the flags from `AddFlags` entirely rather than turning them into silent no ops. Go's `flag` package already produces a clear error when an unknown flag is provided, which is the right failure mode.
5. Add documentation that maps every removed CLI flag to its corresponding `driverConfig` key.

We should be careful to not let precedence flip mid-window (flags must keep winning over file as today, until they are fully gone).

Lastly, I'm sure there are still a few things we will need to align on regarding what should be exposed as a CLI flag versus what belongs in `driverConfig`. Happy to discuss it here or leave to a separate discussion as @ffromani mentioned it [here](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/244#discussion_r3622654691)

**Why is this needed**:
Follow up improvment to #213

fmuyassarov:
#### What type of PR is this?
/kind deprecation

As a follow up to the migration to the config file. 
Current behavior, cpuDeviceMode, groupBy, hostnameOverride and reservedCPUs end up as args.
```sh
helm install dra-driver-cpu deployment/helm/dra-driver-cpu/ \
--set args.logLevel=6 \
--set args.cpuDeviceMode=individual \
--set args.groupBy=socket \
--set args.reservedCPUs=0-1 \
--set args.hostnameOverride=dra-driver-cpu-worker \
--set args.exposePCIeRoots=true \
--set image.repository=ttl.sh/cpu \
--set image.tag=v1 \
--set fullnameOverride=dracpu-b
```
```
kubectl get ds dracpu-b -o json | jq '.spec.template.spec.containers[0].args'
```
```JSON
[
  "/dracpu",
  "--v=6",
  "--cpu-device-mode=individual",
  "--group-by=socket",
  "--bind-address=:8080",
  "--reserved-cpus=0-1",
  "--hostname-override=dra-driver-cpu-worker",
  "--expose-pcie-roots=true"
] 
```

With this change (using `driverConfig`)
```
helm install dra-driver-cpu deployment/helm/dra-driver-cpu/ \
....
--set driverConfig.cpuDeviceMode=individual \
--set driverConfig.groupBy=socket \
--set driverConfig.reservedCPUs=0-1 \
--set driverConfig.hostnameOverride=dra-driver-cpu-worker \
....
```
or using `args`
```
helm install dra-driver-cpu deployment/helm/dra-driver-cpu \
...
--set args.cpuDeviceMode=individual \
--set args.groupBy=socket \
--set args.reservedCPUs=0-1 \
--set args.hostnameOverride=dra-driver-cpu-worker \
...
```
will result in 

```
kubectl get ds dracpu-b2 -o json | jq '.spec.template.spec.containers[0].args'
```
```JSON
[
  "/dracpu",
  "--v=4",
  "--bind-address=:8080",
  "--expose-pcie-roots=false",
  "--config=/etc/dracpu/config.yaml"
]
```
and those driver configuration options end up as a configMap (config) settings

```YAML
kubectl get configmap dracpu-a-config -o yaml
apiVersion: v1
data:
  config.yaml: |
    cpuDeviceMode: individual
    groupBy: socket
    hostnameOverride: dra-driver-cpu-worker
    reservedCPUs: 0-1
kind: ConfigMap
...
```

Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/245

As you will notice during the review, we are not dropping any flags yet and we should probably come to some consensus on the release timeline for finally removing these flags. This patch is more of a signal to the user that are moving away from the flags. 

We print warning in the logs only when flags are explicitly passed on the command line regardless of the value given, even if it matches the built-in default.  The warning is tied to flag usage, not to the resolved config value. Values reaching the driver via `driverConfig` (a mounted config file, whether written directly or Helm folded from `args.*`) never trigger this warning, since no deprecated flag is present on the CLI in that case. This is intentional since it's the migration path we want to encourage.

In other words, warnnig is visible for example when driver is configured via `extraArgs`

```
helm install dra-driver-cpu deployment/helm/dra-driver-cpu \
--set fullnameOverride=dracpu-warn \
--set-string extraArgs[0]=--cpu-device-mode=individual \
--set image.repository=ttl.sh/cpu \
--set image.tag=v2
```

```
kubectl logs dracpu-warn-6jbz9 | grep -i "deprecated\|Warning"
I0728 08:28:52.016591       1 load.go:67] "flag is deprecated and will be removed in a future release; prefer the equivalent driverConfig field instead" flag="cpu-device-mode" driverConfigField="cpuDeviceMode"
Warning: Flag --cpu-device-mode has been deprecated, prefer driverConfig field "cpuDeviceMode"; this flag will be removed in a future release

```

#### Release note
```release-note
Deprecated: the CLI flags `--cpu-device-mode`, `--group-by`, `--reserved-cpus`, `--hostname-override`, `--sysfs-overlay` in favor of `driverConfig`. Deprecated flags now show a deprecation notice in `--help` and log a startup warning when used. The Helm chart automatically migrates equivalent `args.*` values into the generated `driverConfig` ConfigMap, with no behavior change for existing users.
```

#### Documentation updates
- README updates
- Helm chart docs

### DIFF
--- a/docs/user/configuration.md
+++ b/docs/user/configuration.md
@@ -1,53 +1,105 @@
 # Configuration
 
-**IMPORTANT:** The kubelet's CPUManager implements assignment of exclusive CPUs to workloads. The CPUManager and this DRA driver are mutually incompatible and only
-one can be enabled at a time on any given node.
-
-## Kubelet configuration prerequisites
-
-You need to disable the CPUManager on the nodes you wish to run this DRA driver.
-
-1. The default settings of the kubelet are compatible with this DRA driver. If you never fine-tuned the kubelet, you are probably fine.
-1. Make sure `cpuManagerPolicy: "none"` is set in the kubelet [configuration file](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/).
-1. If you changed the kubelet configuration, restart the kubelet to take effect. **NOTE:** you may need to [delete the CPUManager state file](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#changing-the-cpu-manager-policy).
-1. You may now proceed with deploying and configuring this DRA driver.
+This page covers configuring the driver itself, the kubelet prerequisites required to run it,
+and the command-line flags currently being deprecated in favour of the configuration file.
 
 ## Driver configuration
 
-**The configuration file is the preferred configuration mechanism.** Command-line flags are
-still supported, but they are kept mainly for backward compatibility and are expected to be
-phased out in favour of the configuration file over time (see
-[Versioning and backward compatibility](#versioning-and-backward-compatibility) below). If a
-field is set both ways, the flag wins, so avoid mixing the two for the same field and configure
-new deployments via `driverConfig` / the config file alone.
+The driver supports two configuration mechanisms:
+
+- The config file - the main, preferred way to configure the driver.
+- Command-line flags — kept for a small set of important, mostly behavior altering settings and for backward compatibility. Some flags are being deprecated in favor of their config file equivalent (see [Helm: deprecated args values vs driverConfig](#helm-deprecated-args-values-vs-driverconfig) below).
+
+If the same setting is provided both ways, the explicit command-line flag wins, so avoid mixing
+the two for the same field.
 
 ### Configuration file
 
-When deploying with Helm, set `driverConfig` in your values file. The chart serializes the
-map to YAML, stores it in a ConfigMap, mounts it as `/etc/dracpu/config.yaml` inside the
-driver container, and passes `--config` automatically.
+The config file is a YAML file passed to the driver via `--config <path>`.
 
-#### Schema
+When deploying with Helm, you don't write this file yourself: set the `driverConfig` value in
+your values file and the chart serializes it to YAML, stores it in a ConfigMap, mounts it as
+`/etc/dracpu/config.yaml` inside the driver container and passes `--config` automatically. For
+example, with a `values.yaml` containing:
 
-The config file is a **flat** YAML map - there are no nested groups. All fields are optional
+```yaml
+# values.yaml
+# Driver configuration
+driverConfig:
+  cpuDeviceMode: individual
+  reservedCPUs: "0-1"
+# Other tuning knobs of Helm chart
+image:
+  tag: v0.2.0
+resources:
+  requests:
+    cpu: "200m"
+    memory: "100Mi"
+nodeSelector:
+  kubernetes.io/os: linux
+```
+
+```console
+helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driver-cpu -f values.yaml
+```
+
+Individual fields can also be tuned with `--set` instead of a values file, e.g. to switch
+`cpuDeviceMode` to `individual` and reserve CPUs `0-1`:
+
+```shell
+helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driver-cpu \
+    --set driverConfig.cpuDeviceMode=individual \
+    --set driverConfig.reservedCPUs="0-1"
+```
+
+#### driverConfig sub-fields
+
+The config file is a flat YAML map - there are no nested groups. All fields are optional
 except where noted. Unknown fields are rejected at startup to catch typos early.
 
-| Field              | Type   | Default     | Description                                                                              |
-| ------------------ | ------ | ----------- | ---------------------------------------------------------------------------------------- |
-| `apiVersion`       | string | *(omitted)* | When present, must be `v1alpha1`. Rejected otherwise.                                    |
-| `cpuDeviceMode`    | string | `grouped`   | CPU exposure mode: `individual` or `grouped`.                                            |
-| `groupBy`          | string | `numanode`  | Grouping strategy when `cpuDeviceMode` is `grouped`: `numanode`, `socket`, or `machine`. |
-| `reservedCPUs`     | string | *(none)*    | CPUs excluded from allocation, e.g. `"0-1"`.                                             |
-| `hostnameOverride` | string | *(none)*    | Override the node hostname the driver registers under.                                   |
-| `exposePCIeRoots`  | bool   | `false`     | Add PCIe root attributes to CPU devices (requires `DRAListTypeAttributes` feature gate). |
-| `kubeconfig`       | string | *(none)*    | Path to a kubeconfig file (for out-of-cluster use).                                      |
+These fields only affect the driver's own behavior (CPU allocation, hostname, sysfs path,
+etc.). Anything about the driver's Pod itself - image, resources, node placement, and so
+on - is configured through other Helm values, not through this file.
 
-#### Versioning and backward compatibility
+`apiVersion` (string)
 
-The schema is versioned via the optional `apiVersion` field (currently `v1alpha1`). The layout
-is intentionally flat for now. If a nested hierarchy is introduced in the future, the
-`apiVersion` field will be bumped so that older config files continue to be accepted or produce a
-an error.
+- When present, must be `v1alpha1`. Rejected otherwise.
+
+`cpuDeviceMode` (string, default: `grouped`)
+
+- `individual`: exposes each allocatable CPU as a separate device in the `ResourceSlice`.
+  This mode provides fine-grained control, as it exposes granular information specific
+  to each CPU as device attributes.
+- `grouped`: exposes a single device representing a group of CPUs. This mode treats CPUs
+  as a [consumable capacity](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5075-dra-consumable-capacity/README.md)
+  within the group, improving scalability by reducing the number of API objects.
+
+`groupBy` (string, default: `numanode`)
+
+- Grouping strategy used when `cpuDeviceMode` is `grouped`.
+- `numanode`: groups CPUs by NUMA node.
+- `socket`: groups CPUs by socket.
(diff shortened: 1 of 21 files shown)


fmuyassarov:
> /cc @ffromani @pravk03

no rush folks in the review though. I know that there is a lot going on here and outside as well.

ffromani:
thanks @fmuyassarov . I like the general direction of deprecating before removing.
My very initial and very open to be refined idea is we should deprecate the flags in the 0.3.0 allowing user to transition to configuration, add anything new as configuration options (bar clear exception) and removem the flags in 0.4.0.
This is slow and conservative, and unnecessarily so: the driver still call itself alpha. But OTOH we can afford with minimal to none extra burden, just some bookkeeping. So why not?

fmuyassarov:
> My very initial and very open to be refined idea is we should deprecate the flags in the 0.3.0 allowing user to transition to configuration, add anything new as configuration options (bar clear exception) and removem the flags in 0.4.0.
> This is slow and conservative, and unnecessarily so: the driver still call itself alpha. But OTOH we can afford with minimal to none extra burden, just some bookkeeping. So why not?

Make sense @ffromani and thanks for the opinion. Like you said, it is going to be slow (but not too slow either) but probably safer and common path in such situations. Basically, 0.3.0 for depreciation and 0.4.0 for removal sounds good to me.

internal/driverconfig/load.go
```diff
@@ -36,6 +36,29 @@ var flagToJSONKey = map[string]string{
 	"sysfs-overlay":     "sysfsOverlay",
 }
 
+// deprecatedFlags are standalone CLI flags being phased out in favour of the
+// same-named driverConfig field (issue #245).
+var deprecatedFlags = map[string]bool{
```
ffromani: should this be a set?

internal/driverconfig/deprecated_internal_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"flag"
+	"fmt"
+	"strings"
+	"testing"
+)
+
+// Every deprecatedFlags entry maps to a real driverConfig field.
+func TestDeprecatedFlags_AreSubsetOfFlagToJSONKey(t *testing.T) {
+	for f := range deprecatedFlags {
```
ffromani: I'm not completely fine testing private data, this is arguably even worse than testing private functions.
Plus, in this case the test code is subtly coupled with the production code, reusing the same data we're essentially using the same flow in both place.
Can we have a single test which hardcodes the expected deprecated flag set?

ffromani: or perhaps, let's check the coverage we add in `config_test.go` is sufficient. If so, we may not need this internal test at all.

docs/user/configuration.md
```diff
@@ -1,53 +1,105 @@
...
 
-You need to disable the CPUManager on the nodes you wish to run this DRA driver.
+The driver supports two configuration mechanisms:
 
-1. The default settings of the kubelet are compatible with this DRA driver. If you never fine-tuned the kubelet, you are probably fine.
-1. Make sure `cpuManagerPolicy: "none"` is set in the kubelet [configuration file](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/).
-1. If you changed the kubelet configuration, restart the kubelet to take effect. **NOTE:** you may need to [delete the CPUManager state file](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#changing-the-cpu-manager-policy).
-1. You may now proceed with deploying and configuring this DRA driver.
+- The config file — the preferred mechanism for all new deployments.
+- Command-line flags — kept mainly for backward compatibility, and being phased out over time in favour of the config file (see [Helm: deprecated args values vs driverConfig](#helm-deprecated-args-values-vs-driverconfig) below).
```
ffromani: The intent is good, but this read too hard. We should convey the config file is the main vehicle, and that command line flags are kept for exceptional cases. IOW, the nuance is that the direction is not "no more flags" but rather "use flags sparingly and for truly important, deeply behavior-altering cases".

ffromani (commented):
Partial review, but looks good already

internal/driverconfig/deprecated_internal_test.go
fmuyassarov: Yes sorry for this and good that you raised it. I now remember similar comment in some other patch. Totally agree and updating the commit to drop it.

docs/user/configuration.md
fmuyassarov: makes sense, how about for example:

```
- The config file - the main, preferred way to configure the driver.
- Command-line flags - kept for a small set of important mostly behavior altering settings and for backward compatibility. Some flags are being deprecated in favor of their config file equivalent
```

internal/driverconfig/load.go
fmuyassarov: absolutely, let's add idiomatic code rather

docs/user/configuration.md
ffromani: LGTM, thanks

internal/driverconfig/deprecated_internal_test.go
ffromani: no worries. This is hard to enforce automatically, sowe depend on review to catch this. Some repetition is expected :)

internal/driverconfig/config_test.go
```diff
@@ -170,48 +170,103 @@ func TestLoad_MissingFileIsError(t *testing.T) {
...
+	}, funcr.Options{})
+
+	driverconfig.WarnDeprecatedFlags(fs, logger)
+
+	assert.NotContains(t, logs.String(), "deprecated")
+}
+
+// deprecatedFlagNames is the hardcoded set of flags expected to be marked
+// as deprecated in --help. Update it whenever a flag's deprecation status changes.
+var deprecatedFlagNames = map[string]struct{}{
```
ffromani: Sorry for being too terse. I meant k8s.io/apimachinery/pkg/util/sets - probably NOT worth a resubmit just to fix this.

internal/driverconfig/load.go
```diff
@@ -36,6 +36,29 @@ var flagToJSONKey = map[string]string{
 	"sysfs-overlay":     "sysfsOverlay",
 }
 
+// deprecatedFlags is the set of standalone CLI flags being phased out in
+// favour of the same-named driverConfig field (issue #245).
+var deprecatedFlags = map[string]struct{}{
```
ffromani: ditto about `k8s.io/apimachinery/pkg/util/sets`

ffromani (commented):
/lgtm

one minor nit. @fmuyassarov I'm fine merging as-is, your call

internal/driverconfig/config_test.go
fmuyassarov: ah my bad. Let's fix it.

fmuyassarov: we are good now here @ffromani

ffromani:
thanks @fmuyassarov , doing the (I think) the last pass. FYI we're discussing in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/244#issuecomment-5141477333 if `sysfs-overlay` should be a command line-only flag. The convo is still ongoing and I don't think is worth adding more churn to this PR but rather post a followup once we reach agreement. Therefore if everything looks in order I'd make this PR merge.

ffromani (commented):
/approve
/lgtm

thanks for the updates

PR #265, opened 2026-07-29
feat: publish device attributes as KEP-5304 metadata

### LINKED ISSUE #264: Publish device attributes as KEP-5304 metadata

Publish device attributes as KEP-5304 metadata via the DRA device metadata
downward API, so that consumers can read per-device attributes (NUMA node,
socket ID, core ID, etc.) from metadata files mounted into containers.

The kubeletplugin framework at v0.36.0 already supports this — the driver
just needs to enable it and populate the `Metadata` field on each prepared
device. The implementation is small (~30 lines): enable
`EnableDeviceMetadata`, `MetadataVersions`, and `PluginDataDirectoryPath`
in `Start()`, and copy the existing ResourceSlice device attributes into
`DeviceMetadata` in `prepareDevices()`.

The kubeletplugin documentation recommends this approach:

> As a best practice, drivers should include all attributes that are
> published in the ResourceSlice for the device, so that the same
> information is available to workloads at runtime.

**What attributes would be published**:

The metadata mirrors the ResourceSlice device attributes for each mode:

**Individual mode** — per CPU device:
- `dra.cpu/cpuID`, `dra.cpu/coreID`, `dra.cpu/socketID`, `dra.cpu/numaNodeID`
- `dra.cpu/cacheL3ID`, `dra.cpu/coreType`, `dra.cpu/smtEnabled`
- `dra.net/numaNode` (compatibility)
- `resource.kubernetes.io/pcieRoot` (if enabled)

**Grouped by NUMA** — per NUMA group device:
- `dra.cpu/numaNodeID`, `dra.cpu/socketID`, `dra.cpu/numCPUs`, `dra.cpu/smtEnabled`
- `dra.net/numaNode` (compatibility)

**Grouped by socket** — per socket device:
- `dra.cpu/socketID`, `dra.cpu/numCPUs`, `dra.cpu/smtEnabled`

**Grouped by machine** — single device:
- `dra.cpu/numCPUs`, `dra.cpu/smtEnabled`

Note: in grouped modes, the metadata describes the *group device* (the pool),
not the specific CPUs selected by `TakeByTopologyNUMAPacked`. The resolved
cpuset continues to be communicated via the `DRA_CPUSET_*` CDI env var.

**Follow-up consideration**: for grouped modes (socket, machine), topology-aware
consumers would benefit from per-allocated-CPU entries in the metadata (with
`cpuID`, `numaNodeID`, etc.) so they can map the resolved cpuset to NUMA
topology without reading host sysfs. This would be a separate change since
it goes beyond mirroring ResourceSlice attributes.

**Why is this needed**:

Today the driver communicates CPU assignments solely via the `DRA_CPUSET_*`
CDI environment variable, which provides the allocated cpuset but no
topology information (NUMA node, socket, core). Consumers that need to
understand the topology of their allocated CPUs must read host sysfs,
which adds a host dependency and requires additional mount configuration.

KEP-5304 metadata solves this by making the same device attributes that
are published in the ResourceSlice available inside the container as
structured JSON files. This enables:

- Topology-aware workloads to discover CPU NUMA placement at runtime
  without reading host sysfs
- VM managers like KubeVirt to construct guest NUMA topology from
  device attributes
- Orchestration tools to discover device attributes through a standard
  mechanism across DRA device types

**Related**:

- KEP-5304: https://github.com/kubernetes/enhancements/issues/5304
- #181 (assigned.cpuset parity via CDI — complementary, not overlapping)

**Implementation**:

PR: #265

/kind feature

johnahull:
## Summary

- Enable KEP-5304 device metadata publishing so consumers can read
  per-device attributes (NUMA node, socket, core ID, etc.) from
  metadata files mounted into containers
- Add `EnableDeviceMetadata`, `MetadataVersions`, and
  `PluginDataDirectoryPath` options to `kubeletplugin.Start()`
- Populate `Metadata` field on each prepared device in `prepareDevices()`
  with the same attributes published in the ResourceSlice
- Add `getDeviceAttributes()` helper that looks up attributes from
  the existing `deviceSlices` field

Fixes #264

## Test plan

- [x] Existing unit tests updated to verify metadata is populated
- [x] All unit tests pass (`go test ./pkg/driver/...`)
- [x] `golangci-lint` passes with zero issues
- [ ] Manual verification: deploy on test cluster, create CPU claim,
      verify metadata JSON files at
      `/var/run/kubernetes.io/dra-device-attributes/` inside container

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -283,7 +283,7 @@ func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.Resou
 		if allocResult.Request != "" {
 			preparedDevice.Requests = []string{allocResult.Request}
 		}
-		if attrs, ok := cp.getDeviceAttributes(allocResult.Device); ok && len(attrs) > 0 {
+		if attrs, ok := getDeviceAttributes(cp.topology.deviceSlices, allocResult.Device); ok && len(attrs) > 0 {
 			metadataAttrs := make(map[string]resourceapi.DeviceAttribute, len(attrs))
 			for k, v := range attrs {
 				metadataAttrs[string(k)] = v
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -283,6 +283,16 @@ func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.Resou
 		if allocResult.Request != "" {
 			preparedDevice.Requests = []string{allocResult.Request}
 		}
+		if attrs, ok := cp.getDeviceAttributes(allocResult.Device); ok && len(attrs) > 0 {
+			metadataAttrs := make(map[string]resourceapi.DeviceAttribute, len(attrs))
+			for k, v := range attrs {
+				metadataAttrs[string(k)] = v
+			}
+			preparedDevice.Metadata = &kubeletplugin.DeviceMetadata{
+				Attributes: metadataAttrs,
+			}
+			logger.V(6).Info("added device metadata", "device", allocResult.Device, "numAttrs", len(metadataAttrs))
+		}
 		preparedDevices = append(preparedDevices, preparedDevice)
 	}
 
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -288,6 +288,12 @@ func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.Resou
 			for k, v := range attrs {
 				metadataAttrs[string(k)] = v
 			}
+			if quantity, ok := allocResult.ConsumedCapacity[device.CPUResourceQualifiedName]; ok {
+				allocatedCount := quantity.Value()
+				metadataAttrs[string(device.AttributeAllocatedNumCPUs)] = resourceapi.DeviceAttribute{
+					IntValue: &allocatedCount,
+				}
+			}
 			preparedDevice.Metadata = &kubeletplugin.DeviceMetadata{
 				Attributes: metadataAttrs,
 			}
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -643,8 +643,10 @@ func TestPrepareResourceClaims(t *testing.T) {
 			expectedCdiDevice:       cdiDeviceName,
 			expectedCdiEnvVar:       fmt.Sprintf("%s_%s=%s", cdiEnvVarPrefix, claimUID, "0-1"),
 			expectedPreparedDevices: []kubeletplugin.Device{
-				{PoolName: testNodeName, DeviceName: "cpudev000", CDIDeviceIDs: []string{cdiQualifiedName}},
-				{PoolName: testNodeName, DeviceName: "cpudev002", CDIDeviceIDs: []string{cdiQualifiedName}},
+				{PoolName: testNodeName, DeviceName: "cpudev000", CDIDeviceIDs: []string{cdiQualifiedName},
+					Metadata: metadataFromCPUInfo(cpuinfo.CPUInfo{CpuID: 0, CoreID: 0, SocketID: 0, NUMANodeID: 0, CoreType: cpuinfo.CoreTypePerformance}, false)},
+				{PoolName: testNodeName, DeviceName: "cpudev002", CDIDeviceIDs: []string{cdiQualifiedName},
+					Metadata: metadataFromCPUInfo(cpuinfo.CPUInfo{CpuID: 1, CoreID: 1, SocketID: 0, NUMANodeID: 0, CoreType: cpuinfo.CoreTypePerformance}, false)},
 			},
 		},
 		{
@@ -747,7 +749,8 @@ func TestPrepareResourceClaims(t *testing.T) {
 			expectedCdiDevice:       cdiDeviceName,
 			expectedCdiEnvVar:       fmt.Sprintf("%s_%s=%s", cdiEnvVarPrefix, claimUID, "0"),
 			expectedPreparedDevices: []kubeletplugin.Device{
-				{PoolName: testNodeName, DeviceName: "cpudev000", CDIDeviceIDs: []string{cdiQualifiedName}},
+				{PoolName: testNodeName, DeviceName: "cpudev000", CDIDeviceIDs: []string{cdiQualifiedName},
+					Metadata: metadataFromCPUInfo(cpuinfo.CPUInfo{CpuID: 0, CoreID: 0, SocketID: 0, NUMANodeID: 0, CoreType: cpuinfo.CoreTypePerformance}, false)},
 			},
 		},
 		{
@@ -798,11 +801,7 @@ func TestPrepareResourceClaims(t *testing.T) {
 					require.Empty(t, result.Devices)
 				} else {
 					require.NoError(t, result.Err)
-					expected := tc.expectedPreparedDevices
-					for i := range expected {
-						expected[i].Metadata = expectedDeviceMetadata(driver, expected[i].DeviceName)
-					}
-					require.ElementsMatch(t, expected, result.Devices)
+					require.ElementsMatch(t, tc.expectedPreparedDevices, result.Devices)
 				}
 			}
 
@@ -1227,13 +1226,16 @@ func TestPrepareResourceClaimsGroupedMode(t *testing.T) {
 					// Build expected devices based on the claim request
 					expectedPreparedDevices := []kubeletplugin.Device{}
 					if tc.expectedCPUSet.Size() != 0 || tc.expectedError {
+						// testSysFS doesn't include cpu/smt/control, so
+						// the driver's cpuTopology.SMTEnabled is always false
+						smtEnabled := false
 						for _, res := range tc.claims[0].Status.Allocation.Devices.Results {
 							expectedPreparedDevices = append(expectedPreparedDevices, kubeletplugin.Device{
 								PoolName:     res.Pool,
 								DeviceName:   res.Device,
 								CDIDeviceIDs: []string{cdiQualifiedName},
 								Requests:     []string{res.Request},
-								Metadata:     expectedDeviceMetadata(driver, res.Device),
+								Metadata:     expectedGroupMetadata(tc.groupBy, tc.cpuInfos, tc.reservedCPUs, res.Device, smtEnabled),
 							})
 						}
 					}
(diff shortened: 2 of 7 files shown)


ffromani:
/ok-to-test

thanks! the manual verification should be implemented as e2e tests

pkg/driver/dra_hooks_test.go
```diff
@@ -2004,3 +2009,20 @@ func createCPUDriverForTest(t *testing.T, groupBy string, cpuInfos []cpuinfo.CPU
 	}
 	return driver
 }
+
+// expectedDeviceMetadata builds the DeviceMetadata that prepareDevices would
+// produce for a device, by looking up its attributes from the driver's
+// deviceSlices. Returns nil if the device is not found or has no attributes.
+func expectedDeviceMetadata(driver *CPUDriver, deviceName string) *kubeletplugin.DeviceMetadata {
+	attrs, ok := driver.getDeviceAttributes(deviceName)
```
ffromani: test code should not reuse the production code in general if possible, but this is especially true if we are reusing the very code we are testing. We need to express this logic in a different way. It would be preferred to match against a static, precomputed data, for example.

pkg/driver/driver.go
```diff
@@ -36,6 +36,7 @@ import (
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	drametadatav1alpha1 "k8s.io/dynamic-resource-allocation/api/metadata/v1alpha1"
```
ffromani, line 39: just to be sure, this is enabled by default and we don't need to opt-in in any feature gate, right?

pkg/driver/driver.go
```diff
@@ -281,6 +285,17 @@ func (cp *CPUDriver) Stop() {
...
+func (cp *CPUDriver) getDeviceAttributes(deviceName string) (map[resourceapi.QualifiedName]resourceapi.DeviceAttribute, bool) {
+	for _, slice := range cp.deviceSlices {
+		for _, dev := range slice {
+			if dev.Name == deviceName {
+				return dev.Attributes, true
+			}
+		}
+	}
+	return nil, false
+}
```
ffromani, line 351: this can be turned in a free function. We are trying to fight back the gravitational pull of the `CPUDriver` demi-god object and split responsabilities as much as possible

ffromani (commented):
few relatively minor implementation comments, the main missing thing is to improve testing a bit; most notably write e2e test(s) which implement the suggested manual verification

pkg/driver/dra_hooks.go
```diff
@@ -282,6 +282,16 @@ func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.Resou
 		if allocResult.Request != "" {
 			preparedDevice.Requests = []string{allocResult.Request}
 		}
+		if attrs, ok := cp.getDeviceAttributes(allocResult.Device); ok && len(attrs) > 0 {
+			metadataAttrs := make(map[string]resourceapi.DeviceAttribute, len(attrs))
+			for k, v := range attrs {
```
pravk03, line 288: Should we publish all attributes here, or should we be selective?

For example, in grouped mode, we have `numCPUs` attribute for cpu's per NUMA/Socket. Published that to all the containers might be misleading ?

pkg/driver/driver.go
pravk03: +1. I see that KEP-5304 went beta in 1.37. We still need the config option to disable this in older clusters ?

johnahull: Yes, enabled by default. No feature gate needed. `EnableDeviceMetadata` is a framework-side option, not a kubelet gate. The kubelet processes the resulting CDI specs through the standard DRA path. If the kubelet doesn't support it, the CDI specs aren't generated and metadata is silently skipped. I'm getting access to a test system next week and will confirm end-to-end.

pkg/driver/driver.go
johnahull: Fixed in 4a67698. Converted to a free function: `getDeviceAttributes(deviceSlices [][]resourceapi.Device, deviceName string)`.

pkg/driver/dra_hooks_test.go
johnahull: Fixed in 4a67698. Replaced with helpers that build expected metadata from static test data without calling production code.

pkg/driver/driver.go
ffromani: fine, thanks for checking. Let's have a final check and we can settle this. If it's easier, a kind cluster should be fine for this check (e.g. our CI!)

pkg/driver/dra_hooks.go
johnahull: I don't think we need to be selective, since we're just mirroring what's already in the resource slices. But I'm open if you think we should be.

pravk03: My only concern is `numCPUs`, as it's easy to misinterpret when a container reads this. In grouped mode every claim consuming from the same device gets the same value, regardless of how many CPUs it was actually allocated. 
We could publish `allocatedNumCPUs` to indicate what the claim actually got ? 
And/Or 
`numCPUs` should be disambiguated to something like `numaNodeNumCPUs`/`socketNumCPUs` to indicate this value is related to the topology and not the claim request.

johnahull: I like the idea of allocatedNumCPUs. i think that would be beneficial. I'll add dra.cpu/allocatedNumCPUs as a metadata-only attribute that reflects the claim's actual allocation from ConsumedCapacity. That way containers see both numCPUs (device capacity) and allocatedNumCPUs (what this claim got). I'll put it in a follow-up commit.

ffromani (commented):
adding a metadata-only attribute is fine by me.
I think we should have e2e test to implement the "manual verification" step before to merge. Other than this LGTM

johnahull:
Added in 0c4f007. New `/dracpumetadata` binary in the test image reads metadata JSON files from inside the container (test image is distroless, no shell). E2e test creates a claim and pod, execs the binary, and verifies attributes match the device mode. Verified passing on a 4-NUMA Dell XE9680.

pkg/driver/dra_hooks.go
johnahull: Added `dra.cpu/allocatedNumCPUs` in c19f268. Containers now see both `numCPUs` (device capacity) and `allocatedNumCPUs` (what this claim got).

johnahull:
Verified on K8s 1.35 (kind cluster) and K8s 1.37 (Dell XE9680, 4 NUMA nodes).

On 1.37: metadata files appear at the expected path with correct attributes. CDI env var works alongside metadata. Tested grouped-by-NUMA and grouped-by-socket modes.

On 1.35 (pre-KEP-5304): driver starts and runs without errors with `EnableDeviceMetadata(true)`. No crashes, no metadata-related failures.

test/e2e/metadata_test.go
```diff
@@ -0,0 +1,226 @@
...
+		var err error
+		rootFxt, err = fixture.ForGinkgo()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot create root fixture")
+
+		restConfig, err = e2eclient.NewK8SConfig()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "cannot create Kubernetes config")
+
+		daemonSet, err := rootFxt.K8SClientset.AppsV1().DaemonSets("kube-system").Get(ctx, "dracpu", metav1.GetOptions{})
+		if err != nil {
+			daemonSet, err = rootFxt.K8SClientset.AppsV1().DaemonSets("kube-system").Get(ctx, "dra-driver-cpu", metav1.GetOptions{})
```
pravk03: nit: I don't think we need this fallback. All the other tests already depend on `dracpu`.

test/e2e/metadata_test.go
```diff
@@ -0,0 +1,226 @@
...
+			gomega.Expect(req.Name).To(gomega.Equal("request-cpus"))
+			gomega.Expect(req.Devices).ToNot(gomega.BeEmpty())
+
+			dev := req.Devices[0]
+			gomega.Expect(dev.Driver).To(gomega.Equal("dra.cpu"))
+
+			ginkgo.By("verifying mode-specific attributes based on device name")
+			devName := dev.Name
+			switch {
+			case strings.HasPrefix(devName, device.CPUDeviceNUMAGroupedPrefix):
```
pravk03: nit: We already read cpuDeviceMode and groupBy above, we could reuse those instead of prefix matching the device name

pkg/driver/driver.go
pravk03, line 39: Just a note - we'll be bumping to 1.37 as soon as there's a release (this week or next week), and we can add v1beta1 here then. Not sure if it involves any changes

pravk03: Thanks for checking.  Out of curiosity, do you think we would we ever have a reason to turn this off?. The cost of always publishing this seems small, so I dont have a strong opinion on having a config option. 

Separate question, and maybe one for a follow-up: once containers start depending on these attributes being there, do we owe any backward compatibility if we rename or remove attributes later?. One example I can think of is `dra.net/numaNode` is the case I have in mind - we use this as a temporary way for alignment with NIC since the NUMA Node was not a standard attribute until now.

pravk03 (commented):
Thanks for making the changes and adding the e2e tests. The changes mostly LGTM. Left a couple of minot nits, and a follow-up question.

test/e2e/metadata_test.go
johnahull: Removed in 1a6c5ee.

test/e2e/metadata_test.go
johnahull: Fixed in 1a6c5ee. Now uses cpuDeviceMode/groupBy variables with default cases for NUMA grouping.

pkg/driver/driver.go
johnahull: Sounds good. Adding v1beta1 to MetadataVersions should be straightforward when the dependency bumps.

johnahull: I can't think of a scenario where we would want to turn the metadata off. For backward compatibility, I would think it's probably needed. Who knows what attributes will be used by users, and it's a consideration for both the metadata and the resource slices themselves.

pravk03: SG. I wonder what the general stance on attribute stability is with Downward API. Beyond the scope of this PR though.

pravk03:
The only thing I missed earlier is a docs update. Fine as a follow-up PR, or alongside the v1beta1 bump. Rest looks good to me. Thanks for adding the changes. 

/lgtm
/approve

Issue #219, opened 2026-06-30
Publish standard resource.kubernetes.io/numaNode attribute


enoodle:
## Summary

The driver currently publishes NUMA topology as `dra.cpu/numaNodeID` and the compatibility alias `dra.net/numaNode`, but not the standardized `resource.kubernetes.io/numaNode`.

Now that kubernetes/enhancements#6073 merged KEP-6072, we should add the standard attribute too.

/kind feature

## Proposed change

Publish scalar `resource.kubernetes.io/numaNode` wherever the driver already exposes a single concrete NUMA node:

- individual CPU devices
- `group-by=numa`

Keep the existing `dra.cpu/numaNodeID` and `dra.net/numaNode` attributes for compatibility.

Do not add the scalar standard attribute to groupings that may span multiple NUMA nodes, such as `group-by=socket` or `group-by=machine`.

## Notes

#65 attempted this earlier, but was closed because the attribute was not standardized yet. That concern should no longer apply.

Initial support can use the scalar integer form, which does not require `DRAListTypeAttributes`. A later PR can switch to upstream `deviceattribute` constants/helpers once available in this repo’s Kubernetes dependency version.

AutuSnow:
Support this.
﻿
The main blocker from #65 no longer applies: `resource.kubernetes.io/numaNode` is now standardized via kubernetes/enhancements#6073 , so adding it here is no longer using the standard prefix for a repo-local convention.
﻿
The proposed scope in this issue also looks right to me:
﻿
- publish it for individual CPU devices
- publish it for grouped `--group-by=numanode`
- do not publish it for grouped `--group-by=socket` or `--group-by=machine`, because those groupings may span multiple NUMA nodes
﻿
That matches the current publication points in the driver, where NUMA attributes are only attached to individual devices and NUMA-grouped devices today.
﻿
One implementation note: this repo currently depends on `k8s.io/dynamic-resource-allocation v0.36.0`, which already has standard helpers/constants for `pcieRoot` and `pciBusID`, but not yet one for `numaNode`. So it probably makes sense to add the qualified name locally for now and switch to the upstream constant/helper in a follow-up once the dependency version provides it.
﻿

ffromani:
+1, it's a so natural evolution path to the point I almost took for granted we will implement once we rebase on top of kubernetes 1.37

[...]

PR #267, opened 2026-07-30
Publish node allocatable resource mappings (KEP-5517)

### LINKED ISSUE #266: Support KEP-5517 node allocatable resource mappings
An opt-in driver configuration option that makes the driver publish `nodeAllocatableResources` mappings in `ResourceSlice`, as proposed in [KEP-5517](https://github.com/kubernetes/enhancements/issues/5517). With the mapping, the scheduler counts claim CPUs against node allocatable and the kubelet accounts for them in pod- and container-level cgroup limits.

Requires Kubernetes 1.37+ with the `DRANodeAllocatableResources` alpha feature gate enabled.

**Why is this needed**:
Today the scheduler and kubelet are unaware of CPUs allocated through DRA claims: the scheduler can overcommit nodes, and cgroup limits are computed from spec values only. The current workaround is to duplicate the claim's CPU count into container requests/limits (see [workload requirements](https://github.com/kubernetes-sigs/dra-driver-cpu/blob/main/docs/user/workload-requirements.md)), which is fragile and incompatible with advanced DRA features. This option removes that requirement and makes onboarding to the driver easier.

pravk03:
#### What type of PR is this?

/kind feature

Adds support for [KEP-5517: DRA Node Allocatable Resources](https://github.com/kubernetes/enhancements/issues/5517) API. When enabled, the driver
publishes a `nodeAllocatableResources` mapping in `ResourceSlice`, translating its DRA
allocation into node allocatable `cpu`:

- grouped mode: the consumed `dra.cpu/cpu` capacity maps 1:1 via `capacityKey`
- individual mode: each device maps to one CPU via `deviceMultiplier`

With the mapping, the scheduler counts claim CPUs against node allocatable and the kubelet
accounts for them in pod- and container-level cgroup limits. Enabling this config removes current requirement
to duplicate the claim's CPU count into container requests/limits.

The feature is off by default and enabled via the config file only:
`driverConfig.publishNodeAllocatableResourceMapping: true`. It requires the
`DRANodeAllocatableResources` feature gate (alpha, Kubernetes 1.37+).

Fixes: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/266

#### References
- KEP: https://github.com/kubernetes/enhancements/issues/5517
- Kubernetes API/implementation: https://github.com/kubernetes/kubernetes/pull/140009


ffromani:
thanks for filng the PR! looking forward to see this, is it already reviewable?

pravk03:
> thanks for filng the PR! looking forward to see this, is it already reviewable? 

Thanks. I'd recommend holding off until tomorrow for a detailed review. The core implementation for the mapping is done and surprisingly small, but I want to take another pass tomorrow to introduce more tests, changes for CI and likely some cleanup.

Also, to merge this PR, we'll need a 1.37 release version which we do not have yet.

ffromani:
> > thanks for filng the PR! looking forward to see this, is it already reviewable?
> 
> Thanks. I'd recommend holding off until tomorrow for a detailed review. The core implementation for the mapping is done and surprisingly small, but I want to take another pass tomorrow to introduce more tests, changes for CI and likely some cleanup.
> 
> Also, to merge this PR, we'll need a 1.37 release version which we do not have yet.

No worries, I'll wait for your ping then (perhaps just undraft the PR).
Regarding the 1.37 rebase, fully agree. There's also https://github.com/kubernetes-sigs/dra-driver-cpu/issues/219 which depends on the 1.37 tag. No pressure, just keeping our bookings correct.

pravk03:
/hold

Un-drafting the PR. I think it can be reviewed, barring the TODOs for updating the release tag and dependencies once a Kubernetes release containing the `DRANodeAllocatableResources` API is available.  I've left placeholders in the commits to update these later. I think we will have a new release within the next 2 weeks, the CI runs are with the feature disabled.

pravk03:
Failing e2e test will be fixed by https://github.com/kubernetes-sigs/dra-driver-cpu/pull/269

test/e2e/README.md
```diff
@@ -31,25 +31,36 @@ honor these settings, where applicable.
...
   Future version may autodetect it.
+
 - `DRACPU_E2E_CPU_DEVICE_MODE`: (optional): If set, change the behavior of the tests to assume
   the driver is configured with the corresponding device mode. Is meant to mirror the driver
   `--cpu-device-mode` setting.
   NOTE: the Makefile defaults to "grouped" if not set, because a nonempty value is needed internally.
   NOTE: Likewise the e2e tests, the Makefile also behaves differently depending on this setting.
   It is recommended to set it before to run targets like `make ci-manifests`.
+
+- `DRACPU_E2E_NODE_ALLOCATABLE`: (optional, default `false`): if set to `true`,
```
AutuSnow: The kind configs currently leave DRANodeAllocatableResources disabled, so this does not work as documented yet. Could we align the text with the current setup?

test/e2e/sharing_test.go
```diff
@@ -180,10 +179,24 @@ var _ = ginkgo.Describe("Claim sharing", ginkgo.Serial, ginkgo.Ordered, ginkgo.C
...
-			fixture.By("creating a second pod consuming the ResourceClaim on %q, ensuring it gets ContainerCreate Error", fxt.Namespace.Name)
 			createdPod2, err := fxt.K8SClientset.CoreV1().Pods(testPod.Namespace).Create(ctx, &testPod, metav1.CreateOptions{})
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
-			EventuallyFailedToCreate(ctx, fxt, createdPod2)
+			driverConfig := getDriverConfig(ctx, fxt.K8SClientset)
+			if driverConfig.PublishNodeAllocatableResourceMapping {
+				// With the node allocatable mapping, kube-scheduler rejects sharing a mapped claim across pods, so the second pod never
+				// reaches the node. Only assert the outcome that the pod stays unscheduled.
+				fixture.By("ensuring the second pod consuming the ResourceClaim on %q stays unscheduled", fxt.Namespace.Name)
+				gomega.Eventually(func(ctx context.Context) (*v1.Pod, error) {
```
AutuSnow, line 187: Could we also check the scheduling failure message? Any scheduling failure satisfies this assertion, so it does not prove that claim sharing was rejected.

AutuSnow, line 187: Could we also check the scheduling failure message? Any scheduling failure satisfies this assertion, so it does not prove that claim sharing was rejected.

AutuSnow (commented):
The core implementation looks good to me, and #269 explains the current grouped-machine E2E failures. I only found two small test and documentation issues.

test/e2e/README.md
pravk03: Updated. Was waiting for a 1.37 release.

test/e2e/sharing_test.go
pravk03: done

pravk03:
@ffromani  I have updated the dependencies to use v1.37.0-rc.0. This PR is ready for review. 

Note: all e2e jobs currently run with the node allocatable mapping enabled. Will add a follow-up PR in kubernetes/test-infra will set DRACPU_E2E_NODE_ALLOCATABLE: "false" on a subset of the jobs so both modes stay covered. Since the API is alpha and the driver ships with the option disabled by default, so its better to have coverage for both options

ffromani:
> @ffromani I have updated the dependencies to use v1.37.0-rc.0. This PR is ready for review.
> 
> Note: all e2e jobs currently run with the node allocatable mapping enabled. Will add a follow-up PR in kubernetes/test-infra will set DRACPU_E2E_NODE_ALLOCATABLE: "false" on a subset of the jobs so both modes stay covered. Since the API is alpha and the driver ships with the option disabled by default, so its better to have coverage for both options

missed this comment! I'll make sure to find time to review this work ASAP. Thanks!

pravk03:
> missed this comment! I'll make sure to find time to review this work ASAP. Thanks!

Thank you. No rush.

ffromani:
/hold cancel

I see no reason to hold anymore since it is reviewable. Will readd later if needed; likewise feel free to readd if there's a reason I missed

go.mod
```diff
@@ -2,27 +2,29 @@ module github.com/kubernetes-sigs/dra-driver-cpu
 
 go 1.26.0
 
+// TODO: Bump the k8s.io/* modules from v0.37.0-rc.0 to the final v0.37.0 tags once
+// Kubernetes 1.37.0 is released.
```
ffromani, line 6: Yep. An early bump to merge the feature and then a fixup bump before the release is fine by me

deployment/helm/dra-driver-cpu/values.schema.json
```diff
@@ -54,7 +54,7 @@
       "additionalProperties": false
     },
     "driverConfig": {
-      "description": "Driver config file contents. When non-empty, or when a deprecated `args.*` field below is set, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. `args.*` fields that mirror a deprecated CLI flag (`cpuDeviceMode`, `groupBy`, `reservedCPUs`, `hostnameOverride`) are folded into the generated config automatically and take priority over the same field set here. Example: driverConfig: cpuDeviceMode: individual groupBy: socket reservedCPUs: \"0-3\"",
+      "description": "Driver config file contents. When non-empty, or when a deprecated `args.*` field below is set, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. `args.*` fields that mirror a deprecated CLI flag (`cpuDeviceMode`, `groupBy`, `reservedCPUs`, `hostnameOverride`) are folded into the generated config automatically and take priority over the same field set here. Example: driverConfig: cpuDeviceMode: individual groupBy: socket reservedCPUs: \"0-3\" publishNodeAllocatableResourceMapping: true  # KEP-5517; requires the DRANodeAllocatableResources feature gate",
```
ffromani, line 57: while this addition per se is fine, it made me realize we should stop adding examples here and just defer to a more stable user doc for config examples. OFC not a blocker for this PR.

cmd/dracpu/app.go
```diff
@@ -209,14 +209,15 @@ func run(logger logr.Logger, cfg driverconfig.Config) error {
...
-		ExposePCIeRoots:  cfg.ExposePCIeRoots,
-		Metrics:          cpumetrics.New(prometheus.DefaultRegisterer),
-		KubeletRootDir:   cfg.KubeletRootDir,
+		DriverName:                            driverName,
+		NodeName:                              nodeName,
+		ReservedCPUs:                          reservedCPUSet,
+		CPUDeviceMode:                         cfg.CPUDeviceMode,
+		CPUDeviceGroupBy:                      cfg.GroupBy,
+		ExposePCIeRoots:                       cfg.ExposePCIeRoots,
+		PublishNodeAllocatableResourceMapping: cfg.PublishNodeAllocatableResourceMapping,
```
ffromani, line 218: nit: I'd like to spend some cycles thinking about a shorter name. Obviously non blocking, and if I don't comment further on the matter we can just move on as-is. (or we can _briefly_ revisit just before the release)

pkg/device/builder.go
```diff
@@ -58,7 +59,34 @@ func BuildGrouped(logger logr.Logger, groupBy string, topo *cpuinfo.CPUTopology,
...
+	if !enabled {
+		return nil
+	}
+	return map[v1.ResourceName]resourceapi.NodeAllocatableResource{
+		v1.ResourceCPU: {
+			Mapping: &resourceapi.NodeAllocatableMapping{
+				DeviceMultiplier: new(resource.MustParse("1")),
+			},
+		},
+	}
```
ffromani, line 89: the logic and the concept seems fine, the code shape seems clunky but we have a larger issue in this area and we will need some polish. Polish partially done in #229 (which should be merged _after_ this one), which is not blocking and which we can wait in the 0.4.0 cycle and beyond.

deployment/helm/dra-driver-cpu/README.md
```diff
@@ -41,7 +41,7 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
...
-| driverConfig | object | `{}` | Driver config file contents. When non-empty, or when a deprecated `args.*` field below is set, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. `args.*` fields that mirror a deprecated CLI flag (`cpuDeviceMode`, `groupBy`, `reservedCPUs`, `hostnameOverride`) are folded into the generated config automatically and take priority over the same field set here. Example:   driverConfig:     cpuDeviceMode: individual     groupBy: socket     reservedCPUs: "0-3" |
+| driverConfig | object | `{}` | Driver config file contents. When non-empty, or when a deprecated `args.*` field below is set, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. `args.*` fields that mirror a deprecated CLI flag (`cpuDeviceMode`, `groupBy`, `reservedCPUs`, `hostnameOverride`) are folded into the generated config automatically and take priority over the same field set here. Example:   driverConfig:     cpuDeviceMode: individual     groupBy: socket     reservedCPUs: "0-3"     publishNodeAllocatableResourceMapping: true  # KEP-5517; requires the DRANodeAllocatableResources feature gate |
```
ffromani, line 44: filed a tracking issue: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/296 about this portion of the docs being cramped

test/e2e/e2e_suite_test.go
```diff
@@ -107,6 +108,52 @@ func EventuallyFailedToCreate(ctx context.Context, fxt *fixture.Fixture, pod *v1
 	}).WithTimeout(time.Minute).WithPolling(2 * time.Second).Should(podmatchers.BeFailedToCreate(fxt.Log))
 }
 
+// expectNodeAllocClaimStatus verifies the scheduler-populated node allocatable claim status
```
ffromani, line 111: the test logic looks ok to me, we can tighten a bit in some areas but the existing coverage looks sufficient. What can be improved implementation wise is reuse/add custom matchers to make the tests more expressive and composable. I'm not sure myself how a better test shape could look like so I'm fine merging in this form and iterating later.

test/e2e/e2e_suite_test.go
```diff
@@ -120,19 +167,27 @@ type CPUAllocation struct {
...
+	var lastErr error
+	for _, line := range slices.Backward(lines) {
+		err := json.Unmarshal([]byte(line), v)
+		if err == nil {
+			return nil
+		}
+		lastErr = err
+	}
+	return fmt.Errorf("no JSON line found in %d log lines: %w", len(lines), lastErr)
+}
```
ffromani, line 181: we indeed had this patter repeating so the consolidation makes sense. Likewise, the next step should be try to see if and where a higher level construct (e.g. custom matcher) better represent this test. But totally followup work.

ffromani (commented):
/approve
/lgtm

thanks for this important PR. Supporting KEP 5517 is one of **THE** milestones for the release - solving accounting is a major issue for the project - and arguably for the project in its current phase: even the strongest feature set can't compensate for a bad UX, and the UX is the single most severe sore point of the driver atm.

Furthermore, the provisions of the KEP 5517 evolve also thanks to the feedback of this driver, being one of the key adopters and testbed for the node allocatable resource feature.

Therefore, I identified some possible improvements, but those are against an already solid base, and I think is best to merge a PR which is already good rather than seek further polishing. We can and should iterate incrementally

ffromani:
I think the first followup would be to try to expand the test coverage. But this somehow is tied with the e2e polishing I mentioned, so we'll need to figure out the fine details.

ffromani:
another thought I had: should we detect SMT disabled, should we change the multipliers? I'd argue *no*, therefore this code is still correct, because the system-visible meaning of "core" will shift underneath: a "core" would become a physical core, not a logical core.
I'm noting down these thoughts here for due diligence, and probably reiterating the obvious.

pravk03:
> another thought I had: should we detect SMT disabled, should we change the multipliers? 

Really good question, I had not thought about this. I think this also depends on how kubelet publishes node allocatable with SMT disabled. With SMT off, if kubelet counts 1 core as 1 CPU in node allocatable capacity, we should still have a multiplier of 1 in the driver and no change is necessary. I haven't looked into the Kubelet code yet, let me dig deep.

pravk03:
From the kubelet code, it gets the number of CPUs from `/sys/devices/system/cpu/online` (through [cadvisor API](https://github.com/kubernetes/kubernetes/blob/52ba90138eb40cab0987dac73e05c838149bdd1c/pkg/kubelet/nodestatus/setters.go#L220)). It does not consider SMT state before publishing the node capacity. This makes the current driver behavior consistent with kubelet.

Where a non-1 multiplier could become useful - We can have a mode in the future where SMT is ON, but the driver publishes and allocates physical cores only and not hyperthreads. In this case we can have `capacityKey: dra.cpu/cores` and `capacityMultipler: 2`. This would be an equivalent of `full-pcpus-only` CPUManager policy.

deployment/helm/dra-driver-cpu/README.md
pravk03: To be fixed with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/300

test/e2e/e2e_suite_test.go
pravk03: To be fixed with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/300

deployment/helm/dra-driver-cpu/values.schema.json
pravk03: To be fixed with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/300

ffromani:
> > another thought I had: should we detect SMT disabled, should we change the multipliers?
> 
> Really good question, I had not thought about this. I think this also depends on how kubelet publishes node allocatable with SMT disabled. With SMT off, if kubelet counts 1 core as 1 CPU in node allocatable capacity, we should still have a multiplier of 1 in the driver and no change is necessary. I haven't looked into the Kubelet code yet, let me dig deep.

I agree, this is why I LGTM'd: the current code is correct as-is to the best of my knowledge, and it matches the current capabilities and functionality of the kubelet. We *may* possibly leverage > 1 multipliers in the future. The `full-pcpus-only` idea is really nice and something we can explore. How do we compose the tools to reach the solution is yet to be seen, and I think iteration is unavoidable and expected.

PR #268, opened 2026-07-30
chore: group CPUDriver topology fields into their own struct

fmuyassarov:
#### What type of PR is this?
/kind cleanup

This is a minor/cosmetic improvement that was asked during the review of #249. All it does is sub grouping of `CPUDriver` topology fields. No behavioral change. Also, removed [GetCPUInfos()](https://github.com/kubernetes-sigs/dra-driver-cpu/blob/f08f3278be85cd0a0885b39a8d48991f1bdb620a/pkg/driver/driver.go#L66) along the way from the `CPUInfoProvider` interface, which is unused piece of code as it was noted. 

Fixes #

#### Release note

```release-note
NONE
```

#### Documentation updates

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -45,13 +45,13 @@ func (cp *CPUDriver) PublishResources(ctx context.Context) {
 	logger.V(4).Info("begin: publishing resources")
 	defer logger.V(4).Info("end: publishing resources")
 
-	if cp.deviceSlices == nil {
+	if cp.topology.deviceSlices == nil {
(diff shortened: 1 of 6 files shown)


ffromani:
/lgtm

thanks. I'm in favor of this change: whatever help us (within reason!) to split the CPUDriver big object into objects with clear and scoped responsabilities is welcome. In this case the split is possibly not optimal, but it goes in the right direction, it is surely an improvement and I don't have better suggestions :)

ffromani:
/approve

I'm taking the liberty to approve. It's a nice little tidiness change which should be no bother.

PR #269, opened 2026-07-30
fix: treat an empty requests list in opaque configs as applying to all requests

pravk03:
#### What type of PR is this?
/kind bug

https://github.com/kubernetes/kubernetes/pull/139731 changed how the allocator copies a claim's device configs into the allocation result: when a config applies to every request in the claim, the requests list is omitted (the API defines an empty list as "applies to all requests"). The driver currently requires the list to contain exactly one element, so on 1.37+ it rejects every config, and machine mode claim with opaque config fail `NodePrepareResources`. This change updates the driver to match upstream behavior. 

https://github.com/kubernetes-sigs/dra-driver-cpu/issues/266

#### Release note

```release-note
None
```

#### Documentation updates

None

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


AutuSnow:
oh...Sorry, I missed this before the merge. Could we follow up by using resourceclaim.ConfigForResult here? Exact matching misses the API rule that a config targeting foo also applies to foo/bar, causing prioritized-list allocations to fail with zero matching configs.

PR #270, opened 2026-07-31
fix: authenticate DRA CPUSET allocations during synchronize

### LINKED ISSUE #248: Authenticate DRA_CPUSET allocation provenance during NRI Synchronize
During NRI `Synchronize`, the driver currently treats every syntactically valid
`DRA_CPUSET_<claimUID>=<cpuset>` environment entry reported by the runtime as an
authoritative CPU allocation.
﻿
For each entry, `Synchronize`:
﻿
1. parses the claim UID and CPU set from the container environment;
2. records claim ownership;
3. adds the CPU set to a newly created allocation store;
4. reconciles the container cgroup to that CPU set.
﻿
There is no check that the claim was prepared by this driver, that a corresponding
driver-owned CDI device exists, or that the CPU set matches another trusted source.

AutuSnow:
#### What type of PR is this?
/kind bug

Validates `DRA_CPUSET_<claimUID>` allocations reported during NRI `Synchronize` against the corresponding driver-owned CDI device.

Synchronization now fails closed when the CDI device is missing or its CPU set does not match the runtime environment, instead of treating the runtime environment as authoritative.

Fixes #248

### DIFF
--- a/pkg/driver/cdi.go
+++ b/pkg/driver/cdi.go
@@ -91,11 +91,16 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
 	return nil
 }
 
+// Refresh reloads the CDI specs managed by the cache.
+func (c *CdiManager) Refresh() error {
+	if err := c.cache.Refresh(); err != nil {
+		return fmt.Errorf("failed to refresh CDI cache: %w", err)
+	}
+	return nil
+}
+
 // GetDeviceEnv returns the environment edits for a specific CDI device allocation.
 func (c *CdiManager) GetDeviceEnv(deviceName string) ([]string, error) {
-	if err := c.cache.Refresh(); err != nil {
-		return nil, fmt.Errorf("failed to refresh CDI cache: %w", err)
-	}
 	device := c.cache.GetDevice(cdiparser.QualifiedName(cdiVendor, cdiClass, deviceName))
 	if device == nil {
 		return nil, fmt.Errorf("failed to find CDI device %q", deviceName)
--- a/pkg/driver/cdi.go
+++ b/pkg/driver/cdi.go
@@ -99,7 +99,8 @@ func (c *CdiManager) Refresh() error {
 	return nil
 }
 
-// GetDeviceEnv returns the environment edits for a specific CDI device allocation.
+// GetDeviceEnv returns the environment edits for a specific CDI device allocation
+// from the current cache. Call Refresh before lookup to load the latest on-disk specs.
 func (c *CdiManager) GetDeviceEnv(deviceName string) ([]string, error) {
 	device := c.cache.GetDevice(cdiparser.QualifiedName(cdiVendor, cdiClass, deviceName))
 	if device == nil {
--- a/pkg/driver/cdi.go
+++ b/pkg/driver/cdi.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/go-logr/logr"
 	cdiapi "tags.cncf.io/container-device-interface/pkg/cdi"
+	cdiparser "tags.cncf.io/container-device-interface/pkg/parser"
 	cdiSpec "tags.cncf.io/container-device-interface/specs-go"
 )
 
@@ -88,6 +89,18 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
 
 	logger.V(4).Info("Added CDI device", "deviceName", deviceName, "specName", specName, "env", envVar)
 	return nil
+}
+
+// GetDeviceEnv returns the environment edits for a specific CDI device allocation.
+func (c *CdiManager) GetDeviceEnv(deviceName string) ([]string, error) {
+	if err := c.cache.Refresh(); err != nil {
+		return nil, fmt.Errorf("failed to refresh CDI cache: %w", err)
+	}
+	device := c.cache.GetDevice(cdiparser.QualifiedName(cdiVendor, cdiClass, deviceName))
+	if device == nil {
+		return nil, fmt.Errorf("failed to find CDI device %q", deviceName)
+	}
+	return append([]string{}, device.ContainerEdits.Env...), nil
 }
 
 // RemoveDevice deletes the dedicated CDI spec file for a single device allocation.
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -65,10 +65,16 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				allGuaranteedCPUs := cpuset.New()
 				for uid, cpus := range claimAllocations {
 					caLogger := cLogger.WithValues("claimUID", uid)
-					if err := cp.validateSynchronizedClaimAllocation(caLogger, uid, cpus); err != nil {
+					deviceName := getCDIDeviceName(uid)
+					envs, err := cp.cdiMgr.GetDeviceEnv(deviceName)
+					if err != nil {
+						return nil, fmt.Errorf("claim %q is not prepared by this driver: %w", uid, err)
+					}
+					err = validateSynchronizedClaimAllocation(caLogger, uid, cpus, envs)
+					if err != nil {
 						return nil, err
 					}
-					err := claimTracker.SetOwner(caLogger, uid, types.UID(pod.Uid), container.Name)
+					err = claimTracker.SetOwner(caLogger, uid, types.UID(pod.Uid), container.Name)
 					if err != nil {
 						return nil, err
 					}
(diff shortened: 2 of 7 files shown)


pkg/driver/nri_hooks.go
```diff
@@ -142,6 +145,32 @@ func (cp *CPUDriver) validatePreparedClaimAllocation(uid types.UID, cpus cpuset.
 	return nil
 }
 
+func (cp *CPUDriver) validateSynchronizedClaimAllocation(logger logr.Logger, uid types.UID, cpus cpuset.CPUSet) error {
+	if cp.cdiMgr == nil {
+		return fmt.Errorf("cannot validate claim %q during synchronize: CDI manager is not initialized", uid)
+	}
```
ffromani: can this ever happen?

pkg/driver/nri_hooks.go
```diff
@@ -142,6 +145,32 @@ func (cp *CPUDriver) validatePreparedClaimAllocation(uid types.UID, cpus cpuset.
...
+
+	deviceName := getCDIDeviceName(uid)
+	envs, err := cp.cdiMgr.GetDeviceEnv(deviceName)
```
ffromani: it seems to me the only reason why this is a CPUDriver method is to get the `cdiMgr` method. Can we turn this to a free function taking a `cdiMgr` instance? or perhaps, even better, just taking `envs` ?

pkg/driver/nri_hooks.go
```diff
@@ -65,6 +65,9 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				allGuaranteedCPUs := cpuset.New()
 				for uid, cpus := range claimAllocations {
 					caLogger := cLogger.WithValues("claimUID", uid)
+					if err := cp.validateSynchronizedClaimAllocation(caLogger, uid, cpus); err != nil {
```
ffromani: we should be consistent with the surrounding code which doesn't use inline `err` in a `if`. The project doesn't have a convention (and I'm not sure it's possible at all). Probably best addressed in a followup sweeping PR (+ CI enforcement?) though

pkg/driver/cdi.go
```diff
@@ -90,6 +91,18 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
 	return nil
 }
 
+// GetDeviceEnv returns the environment edits for a specific CDI device allocation.
```
ffromani: aI was a bit struck by this naming but after some thinking I could not find a better name which is not overly verbose, which would be bad in a different way.

pkg/driver/nri_hooks_test.go
```diff
@@ -376,7 +377,12 @@ func TestNRISynchronize(t *testing.T) {
 				podConfigStore:     store.NewPodConfig(),
 				cpuAllocationStore: store.NewCPUAllocation(topo, cpuset.New()),
 				claimTracker:       store.NewClaimTracker(),
-				topology:           deviceTopology{cpuTopology: topo},
+				cdiMgr: func() *mockCdiMgr {
+					mgr := newMockCdiMgr()
+					mgr.devices[getCDIDeviceName("claim-A")] = fmt.Sprintf("%s_claim-A=%s", cdiEnvVarPrefix, "0,1")
+					return mgr
```
ffromani: we have a pattern on which we create a mock CdiMgr, we feed some `devices` assignment and then we use it. It seems better to have a helper constructor which abstracts this part and avoid us accessing the private fields from then outside. Repeats pretty consistently in the rest of the changes to this file.

ffromani (commented):
Thanks. Conceptually the PR looks fine, some comments about the implementation.

pkg/driver/cdi.go
ffromani: Checking the CDI methods, this is probably fine.

pkg/driver/nri_hooks.go
AutuSnow: No, I rechecked the link and it shouldn't happen in a production environment because Start() initializes cdiMgr before starting the NRI plugin. I will cancel this check.

pkg/driver/nri_hooks.go
AutuSnow: Agreed！

pkg/driver/nri_hooks.go
AutuSnow: Agreed. I'll use a separate err assignment here for consistency. A broader convention can be handled separately.

pkg/driver/nri_hooks_test.go
AutuSnow: Agreed. I'll add a helper constructor for mock CDI allocations and use it in these tests.

pkg/driver/nri_hooks.go
```diff
@@ -65,7 +65,16 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
...
 					caLogger := cLogger.WithValues("claimUID", uid)
-					err := claimTracker.SetOwner(caLogger, uid, types.UID(pod.Uid), container.Name)
+					deviceName := getCDIDeviceName(uid)
+					envs, err := cp.cdiMgr.GetDeviceEnv(deviceName)
+					if err != nil {
+						return nil, fmt.Errorf("claim %q is not prepared by this driver: %w", uid, err)
+					}
+					err = validateSynchronizedClaimAllocation(caLogger, uid, cpus, envs)
+					if err != nil {
+						return nil, err
```
pravk03: One failed validation (misbehaving container) should not affect other containers/pods ?
I think we should log error, skip the invalid enforcement and keep the blast radius limited.

pkg/driver/nri_hooks.go
```diff
@@ -65,7 +65,16 @@ func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, co
 				allGuaranteedCPUs := cpuset.New()
 				for uid, cpus := range claimAllocations {
 					caLogger := cLogger.WithValues("claimUID", uid)
-					err := claimTracker.SetOwner(caLogger, uid, types.UID(pod.Uid), container.Name)
+					deviceName := getCDIDeviceName(uid)
+					envs, err := cp.cdiMgr.GetDeviceEnv(deviceName)
+					if err != nil {
+						return nil, fmt.Errorf("claim %q is not prepared by this driver: %w", uid, err)
+					}
+					err = validateSynchronizedClaimAllocation(caLogger, uid, cpus, envs)
```
pravk03: I think the intent of this fix is good and much needed, but my concern here is that we may run into scalability challenges. Synchronize() is time bound to 2 seconds by default and we must keep it as lean as possible.

pkg/driver/cdi.go
```diff
@@ -90,6 +91,18 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
...
+func (c *CdiManager) GetDeviceEnv(deviceName string) ([]string, error) {
+	if err := c.cache.Refresh(); err != nil {
```
pravk03, line 96: Given that this is called from Synchronize() for each device, 
do we need to call `c.cache.Refresh()` in every call ? 

Can we move the `c.cache.Refresh()` out into a separate API so that we call it once per `Synchronize()`

pkg/driver/nri_hooks.go
pravk03: May not need anything actionable in this PR, but I filed https://github.com/kubernetes-sigs/dra-driver-cpu/issues/275 for the long term.

pkg/driver/nri_hooks.go
AutuSnow: Good point. Invalid claim allocations are now logged and ignored, so other containers can still be synchronized. A container with no validated claims is restored as shared.

pkg/driver/nri_hooks.go
AutuSnow: Thanks for filing #275. This PR now avoids repeated CDI cache refreshes; broader NRI latency coverage can be tracked there.

pkg/driver/cdi.go
AutuSnow: Agreed. Refresh is now separate from device lookup and runs at most once per Synchronize.

pkg/driver/cdi.go
```diff
@@ -90,6 +91,23 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
...
+func (c *CdiManager) Refresh() error {
+	if err := c.cache.Refresh(); err != nil {
+		return fmt.Errorf("failed to refresh CDI cache: %w", err)
+	}
+	return nil
+}
+
+// GetDeviceEnv returns the environment edits for a specific CDI device allocation.
+func (c *CdiManager) GetDeviceEnv(deviceName string) ([]string, error) {
+	device := c.cache.GetDevice(cdiparser.QualifiedName(cdiVendor, cdiClass, deviceName))
```
pravk03, line 105: nit: GetDeviceEnv now silently depends on the cache being current. Could we document that for future reference.

pkg/driver/nri_hooks.go
pravk03: nit: Not introduced by this PR, but the same point about one malformed container crashing the driver for all pods. We should probably log an error and continue here as well. 

I am ok with fixing it here or in a future PR.

pkg/driver/cdi.go
```diff
@@ -90,6 +91,23 @@ func (c *CdiManager) AddDevice(logger logr.Logger, deviceName string, envVar str
 	return nil
 }
 
+// Refresh reloads the CDI specs managed by the cache.
+func (c *CdiManager) Refresh() error {
+	if err := c.cache.Refresh(); err != nil {
```
pravk03, line 96: I am thinking if we should log an error and continue. Refresh scans the entire dir and the error maybe unrelated to this drivers CDI file. Or, maybe there is a way filter the error to this driver ?

pravk03:
Thanks @AutuSnow for the fix and for addressing all of the review comments. Everything looks good to me!
Taking the liberty to approve this since @ffromani is OOO.

/lgtm
/approve

PR #272, opened 2026-07-31
fix: make the kubelet root directory configurable

thc1006:
#### What type of PR is this?

/kind bug

The kubelet root directory is compiled in as `/var/lib/kubelet`, so a cluster that relocates it cannot run the driver: the registration and plugin-data paths have to match inside and outside the container, because the kubelet watches its own root.

This adds `--kubelet-root-dir`, defaulting to the current path. The value is cleaned and must be absolute.

An empty value is refused rather than defaulted. A tunable that takes a value should be given one, and the way this one arrives empty in practice is a chart that rendered nothing into it, in which case the hostPath mounts rendered from the same value are wrong too. Quietly falling back to the standard root is how the driver would end up registering somewhere the kubelet is not watching, which is the failure this flag exists for.

It is a flag rather than a `driverConfig` field, with a `schemaExcludedFields` entry saying so, for the reason discussed in #244: the chart renders hostPath mounts from the same value, and a config file the chart does not write cannot keep the two in step. Same class as `bindAddress` and `exposePCIeRoots`.

`values.schema.json` is regenerated because the excluded-field list is embedded in it. That is the only file here under `deployment/`.

Review follow-ups since the first push: the driver no longer repeats the config layer's validation, which also dropped `pkg/driver`'s import of `internal/driverconfig`; the two path derivations are plain functions rather than methods on `CPUDriver`; and the tests use `DefaultKubeletRootDir` instead of the literal.

Part of #231. This is the binary plumbing only, so the issue stays open: the chart still hardcodes `/var/lib/kubelet`, and setting the flag alone through `extraArgs` gives a driver whose socket is outside the mounts. The chart wiring is #274, which renders the flag and both hostPath mounts from one value. The relocated-root end-to-end coverage follows in a third PR, and the issue stays open until it lands.

First of three, splitting #244 as agreed there.

`Load` now validates the effective configuration even when no config file was given. Previously it returned early on an empty path, so a bad value from a flag alone reached the kubeletplugin helper unchecked. This is a wider behaviour change than the title suggests: an invalid `cpuDeviceMode` supplied through a non-standard caller now fails at startup. The production CLI validates its enum flags through `flag.Value` already, so the practical exposure is small.

Config-file key handling is deliberately not touched here. `encoding/json` matches struct fields without regard to case, so a spelling like `reservedcpus` survives the explicit-flag precedence pass and overrides the flag. That predates this PR, affects every field rather than this one, and is a behaviour change of its own, so it goes in its own PR rather than riding along here.

#### Release note

**#273 should land first.** This adds `kubeletRootDir` to the excluded fields, and that check compares names exactly while `encoding/json` does not: until #273 is in, a config file spelling it `kubeletrootdir` reaches the field anyway and outranks an explicitly passed `--kubelet-root-dir`, which is the driver-versus-mounts divergence this PR exists to prevent. I checked both: the folded key wins even with the flag set, and the chart's values schema rejects the key in either spelling, so the path is a hand-written `--config` file. Once #273 merges I will rebase and add `kubeletrootdir` to its folded-exclusion cases.
```release-note
The kubelet root directory is now configurable, so the driver can run on clusters whose kubelet uses a non-default `--root-dir`. Set `kubeletRootDir` in the Helm values, or pass `--kubelet-root-dir` to the binary directly; it must be an absolute path. Installs on the standard `/var/lib/kubelet` are unaffected.
```

#### Documentation updates

NONE — the user-facing documentation for the chart value follows in the chart PR.

#### AI usage


thc1006:
The config-key handling is now #273, out of this one as offered above.

It turned out to be wider than the excluded-field check I had here. The explicit-flag precedence pass deletes by exact name too, so a file with `reservedcpus` also survives it and overwrites `--reserved-cpus`. #273 refuses a non-canonical spelling outright rather than folding the comparisons, which also covers a file setting both spellings of one field.

The two are independent; either can merge first.

internal/driverconfig/config.go
```diff
@@ -29,6 +29,11 @@ type Config struct {
 	GroupBy          string `json:"groupBy,omitempty"`
 	ExposePCIeRoots  bool   `json:"exposePCIeRoots,omitempty"`
 	SysFSOverlay     string `json:"sysfsOverlay,omitempty"`
+	// KubeletRootDir is the kubelet root directory. The plugin registration and
+	// plugins directories are derived from it as <root>/plugins_registry and
+	// <root>/plugins. Set it when the kubelet --root-dir is not the default
+	// /var/lib/kubelet.
```
ffromani, line 35: I wonder is there a case to give the 2 parameters independently so, for example, they can be fully independent paths?  Does even the kubelet support such configuration? Should we ever want to go this direction, we can have a followup (much) later on.

internal/driverconfig/config_test.go
```diff
@@ -221,6 +221,9 @@ func TestDefault(t *testing.T) {
 	assert.Equal(t, ":8080", d.BindAddress)
 	assert.Equal(t, device.CPU_DEVICE_MODE_GROUPED, d.CPUDeviceMode)
 	assert.Equal(t, device.GROUP_BY_NUMA_NODE, d.GroupBy)
+	// The kubelet root defaults to the standard location, so behavior is
+	// unchanged unless the kubelet --root-dir is relocated.
+	assert.Equal(t, "/var/lib/kubelet", d.KubeletRootDir)
```
ffromani: the default path should be a constant in driverconfig or driver packages.

internal/driverconfig/config_test.go
```diff
@@ -291,3 +294,74 @@ func TestLoad_ExcludedFieldInFileIsError(t *testing.T) {
...
+	_, err := driverconfig.Load(cfg, "", fs, testr.New(t))
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "must be an absolute path")
+}
+
+// TestLoad_EmptyKubeletRootDirIsDefaulted: an empty kubeletRootDir is
+// normalized to the default in the config layer, so the logged and gathered
+// config reports the value the driver actually uses.
+func TestLoad_EmptyKubeletRootDirIsDefaulted(t *testing.T) {
```
ffromani: I'm sorry if I missed prior conversation, but I think this should be rejected. It smells from a user config error passing through (e.g. bad templating) than else. And, in general, if the tunable expects a value, a value should be given. If there it was consensus reached for this behavior please link the prior conversation so I can catch up.

internal/driverconfig/config_test.go
ffromani: actually we do have the constant already, any reason not to use it? Constants are safe to reuse from production code in test code because... they are constant, with no logic whatsoever :)

internal/driverconfig/load.go
```diff
@@ -34,48 +35,71 @@ var flagToJSONKey = map[string]string{
...
 	}
 
-	logger.V(6).Info("config: after file", result.LogValues()...)
+	// Normalize the kubelet root so the logged config reports the value the
+	// driver actually uses. An empty value becomes the
+	// default; a non-canonical absolute path (with "." or ".." or repeated
+	// separators) is cleaned, matching filepath.Join on the driver side and
+	// clean on the chart side.
+	if result.KubeletRootDir == "" {
+		result.KubeletRootDir = Default().KubeletRootDir
```
ffromani: assuming we want to keep the normalization (see previous comment) can we use the constant directly?

pkg/driver/driver.go
```diff
@@ -146,6 +152,17 @@ func New(logger logr.Logger, providers Providers, config *Config) (*CPUDriver, e
...
+	// production path; this fallback covers direct callers of New that build a
+	// Config without setting KubeletRootDir. A non-empty relative path is
+	// rejected either way.
+	kubeletRootDir := config.KubeletRootDir
+	if kubeletRootDir == "" {
+		kubeletRootDir = driverconfig.DefaultKubeletRootDir
+	}
+	if !filepath.IsAbs(kubeletRootDir) {
+		return nil, fmt.Errorf("kubelet root directory must be absolute: %q", kubeletRootDir)
+	}
```
ffromani: didn't we do this check? is it needed in this flow? I'd expect that once the config is loaded, the value has to be validated

pkg/driver/driver.go
```diff
@@ -213,14 +231,32 @@ func New(logger logr.Logger, providers Providers, config *Config) (*CPUDriver, e
...
+	return filepath.Join(cp.kubeletRootDir, "plugins")
+}
+
+// pluginDataDir is the per-driver directory where the DRA socket is created. It
+// includes the driver name because the kubeletplugin contract requires it not
+// to be shared with other kubelet plugins.
+func (cp *CPUDriver) pluginDataDir() string {
+	return filepath.Join(cp.pluginsDir(), cp.driverName)
+}
+
```
ffromani, line 235: uhm, we are adding more, albeit trivial, methods to the CPUDriver demi-god object, which is not the best direction. OTOH there's not yet an obvious seam to split this code in its own object to compose in CPUDriver, but we need to look for a path there. Non blocking for this PR overall, but if you can avoid adding responsabilities to the CPUDriver object that would be great.

ffromani (commented):
thanks, some implementation comments but we're close to converge. Since @fmuyassarov reviewed the previous PR, I'd appreciate their review here if possible.

internal/driverconfig/config.go
thc1006: The helper does allow them independently, and the kubelet does not care where the plugin data directory sits as long as the path is the same inside and outside the container. For a relocated root they always move together, so one value keeps the chart honest: the hostPath mounts render from it too, and two independent knobs would let those drift apart. Happy to open it up later if a case turns up that needs it.

internal/driverconfig/config_test.go
thc1006: Done, using `driverconfig.DefaultKubeletRootDir`.

thc1006: No reason, I just missed it. Using the constant now.

internal/driverconfig/config_test.go
thc1006: No prior conversation to link, and you are right. An earlier review raised the same thing and I only fixed the wording rather than the behaviour, which was the wrong call: the way this arrives empty in practice is a chart that rendered nothing into it, and the hostPath mounts come from the same value, so falling back to the standard root puts the socket where the kubelet is not watching. It is refused now, including an empty value appended after a good one, since the flag package takes the last.

internal/driverconfig/load.go
thc1006: The normalization stays for the cleaning, but the empty branch is gone with the defaulting, so that use of the constant went with it.

pkg/driver/driver.go
thc1006: Removed. `New` has one caller and it loads the config first, so this was a second copy of a check that belongs in the config layer. It also drops `pkg/driver`s import of `internal/driverconfig`, which existed only for the default it no longer applies.

pkg/driver/driver.go
thc1006: Fair, and easy here: both are now plain functions taking the root, so nothing new hangs off CPUDriver.

fmuyassarov:
> thanks, some implementation comments but we're close to converge. Since @fmuyassarov reviewed the previous PR, I'd appreciate their review here if possible.

Absolutely, this one and https://github.com/kubernetes-sigs/dra-driver-cpu/pull/273 are on my TODO before EOD

thc1006:
Pushed as 695160a, with a reply on each thread.

The chart wiring is up as #274 now, so the two can be read together. One thing there is worth knowing while reviewing this one: adding the flag turned `extraArgs: ["--kubelet-root-dir=..."]` from an unknown-flag crash into a silent win over the chart value, with the mounts left behind. #274 refuses that at template time rather than leaving it open.

pkg/driver/driver.go
thc1006: Follow-up on this one: removing the check left two comments behind that described the fallback that went with it, one of them saying `Load` normalizes an empty root to the default when it now refuses it. Corrected in the push just now, since they said the opposite of what the code does.

thc1006:
Added a check this should have had from the start. The root is now refused when the registrar socket path underneath it would not fit `sun_path`: 108 bytes less the terminating NUL leaves 107 for the path, and the fixed suffix leaves 73 for the root, counted in bytes rather than characters. Without it the root was accepted, the directories were created, and the failure arrived from inside the kubeletplugin helper as `bind: invalid argument`, naming neither the root nor the limit it broke. A test binds a socket at the limit and one byte past it, so the constant answers to the kernel rather than to a header.

/hold

Holding for merge order rather than for anything in this branch. #273 should land first: until it does, a config file spelling the key `kubeletrootdir` still reaches `KubeletRootDir` and outranks an explicitly passed `--kubelet-root-dir`, which is the divergence this PR exists to prevent. I will rebase and add that regression case once it merges.

internal/driverconfig/load.go
```diff
@@ -61,45 +63,65 @@ func WarnDeprecatedFlags(fs *flag.FlagSet, logger logr.Logger) {
...
-		delete(confMap, jsonKey)
-	}
+		if err := applyMap(&result, confMap); err != nil {
+			return Config{}, fmt.Errorf("applying config file %q: %w", filePath, err)
+		}
 
-	result := base
-	if err := applyMap(&result, confMap); err != nil {
-		return Config{}, fmt.Errorf("applying config file %q: %w", filePath, err)
+		logger.V(6).Info("config: after file", result.LogValues()...)
```
fmuyassarov: nit, let's not confuse the reader

```suggestion
		logger.V(6).Info("config: after applying config file", result.LogValues()...)
```

thc1006:
Rebased onto main now that #273 has landed, and the interaction between the two is covered.

The one conflict was in `schema_internal_test.go`, where #273's `TestConfigDeclaresExplicitJSONNames` and this branch's `TestGenerateDriverConfigSchema_ExcludesKubeletRootDir` had both been added at the same place. Both are kept.

`TestLoad_FoldedKubeletRootDirCannotOverrideTheFlag` is the regression I said I would add once #273 was in. A config file with `kubeletrootdir: /wrong/root` alongside `--kubelet-root-dir=/correct/root` used to walk past the excluded-field check, which compares the canonical spelling, and then decode into the field over the value the flag had set. The mounts would still render from the chart's own value, so the driver would register where the kubelet is not watching. Removing #273's canonical-key pass makes that test fail, which is what says the two halves close it together rather than either alone; a control asserts an unrelated file still leaves the flag standing.

`go build`, `go vet` and the unit tests are green on `3771b3c`.

fmuyassarov (commented):
Looks much better. I was looking/searching for the Helm front bits as well but then realized that it is decoupled from this PR, which is better and easier. Let me know if my understanding is incorrect. 
I just had a single nit below

internal/driverconfig/load.go
thc1006: Taken, thanks. Applied in `d4a4760`.

The restructure moved that line into the file branch while the cleaning and validation stayed below it, so `after file` had started to read as the final config rather than one step of it.

fmuyassarov (approved):
Thanks @thc1006 , LGTM. Looking forward for the other part that exposed the flag to the user facing front. 
let's see what @pravk03 @ffromani think about current state.

thc1006:
/hold cancel

The hold was mine and it was for merge order: until #273 landed, a config file spelling the key `kubeletrootdir` still reached `KubeletRootDir` and outranked an explicitly passed `--kubelet-root-dir`. It has landed, this is rebased onto it, and `TestLoad_FoldedKubeletRootDirCannotOverrideTheFlag` covers that interaction. Removing #273's canonical-key pass makes it fail while a control with an unrelated config file keeps passing, so it pins the two halves together rather than either alone.

Thanks @fmuyassarov. The user-facing half is #274, stacked on this one and green.

internal/driverconfig/load.go
```diff
@@ -61,45 +63,65 @@ func WarnDeprecatedFlags(fs *flag.FlagSet, logger logr.Logger) {
...
+	// refuse.
+	if result.KubeletRootDir != "" {
+		result.KubeletRootDir = filepath.Clean(result.KubeletRootDir)
+	}
 
+	// Always validate, so a bad value from a CLI flag with no config file is
+	// rejected too, not only a bad value from the file. The file is named as an
+	// input rather than as the source: what is validated here is the merged
+	// result, and a field the file is not allowed to set -- kubeletRootDir --
+	// would otherwise send the operator to look for it there.
```
pravk03: nit, optional: I think this is self explanatory ?

general comment: the inline comments are a bit too verbose and are repeated is tests and code. We could condense them in some places. Self explanatory comments can be removed.

internal/driverconfig/config_test.go
```diff
@@ -486,3 +489,180 @@ func TestLoad_ExcludedFieldInFileIsError(t *testing.T) {
...
+
+	require.NoError(t, err)
+	assert.Equal(t, "/mnt/fast/k8s/kubelet", result.KubeletRootDir)
+}
+
+// TestLoad_KubeletRootDirRejectedInFile: the kubelet root has to match the
+// hostPath mounts the chart renders, so setting it in the config file (where
+// the chart cannot keep the two in step) is rejected, the same way
+// bindAddress and exposePCIeRoots are.
+func TestLoad_KubeletRootDirRejectedInFile(t *testing.T) {
```
pravk03: nit: I think this test can be added into `TestLoad_ExcludedFieldInFileIsError` ?

General comment: Let's try to add new test cases into existing tests wherever possible. Ideally it would be better to have (and maintain) 1 test with multiple cases within it per exported function.

pravk03 (commented):
Thanks a lot for this fix and iterating though multiple review @thc1006! . The core changes look great now, left a couple of minor/optional nits. 

/approve
/lgtm
/hold

internal/driverconfig/load.go
thc1006: TLDR: fixed across the PR, not only on this line. The comments it adds are down from 112 lines to 35.

This one is gone, along with the paragraph above it that repeated what the test below already asserts. What survives is the part the code cannot say on its own. The registrar path is the longer of the two sockets, so checking it covers both. The length check sits outside `Validate` because it needs the driver name, which the config does not carry, and the socket test is Linux-only because `sun_path` is not a fixed size across Unix.

Thanks for putting it as a general note as well as on the line. That is the version I could act on.

internal/driverconfig/config_test.go
thc1006: TLDR: done here, and applied to the rest of the block. Fourteen new test functions became seven.

`kubeletRootDir` is a row in `TestLoad_ExcludedFieldInFileIsError` now. Its `expectedError` became `wantErrs []string` so the row can assert both the refusal and the pointer to `--kubelet-root-dir`; the three existing rows still carry one string each.

The other eight collapsed into two tables. `TestLoad_KubeletRootDirFromFlag` holds five cases for the flag on its own, and `TestLoad_KubeletRootDirWithAConfigFile` holds three for the flag and a file together, including the control where an unrelated file leaves the flag alone.

Two things surfaced while folding them. `assert.NotEqual(t, "/wrong/root", result.KubeletRootDir)` could never have failed, since `Load` returns a zero `Config` on any error, so it is gone. And a comment describing `TestLoad_RelativeKubeletRootDirIsError` had ended up above a different function during the rebase onto #273, so it documented a test two functions away; the restructure puts that right.

thc1006:
TLDR: both nits are in c4f60eb, with a reply on each thread. The comments this PR adds are down from 112 lines to 35, and its new test functions from fourteen to seven. `make test-unit` and `golangci-lint run ./...` are clean.

@pravk03 `kubeletRootDir` is a row in `TestLoad_ExcludedFieldInFileIsError` as you suggested, and the eight other single-case tests became two tables, one for the flag on its own and one for the flag together with a config file.

Folding them turned up two things I had missed. One assertion could never have failed, because `Load` returns a zero `Config` on any error. A comment had also drifted onto the wrong function while rebasing onto #273.

@ffromani you had a number of threads open here in July and they have all been answered since. If you would like another pass before this goes in, I am happy to wait. @pravk03, the hold is yours whenever suits you.

I have also added the AI usage note the contributor guide asks for to the description, which I should have included from the start.

pravk03:
Thanks for addressing the comments !. 

> @ffromani you had a number of threads open here in July and they have all been answered since. If you would like another pass before this goes in, I am happy to wait.

@ffromani mentioned he would be OOO for the next few days, but I see all of his comments are addressed.

I also see that @fmuyassarov has lgtm'd in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/272#pullrequestreview-4844918159. 

I think we are good to go on this. Please feel free to un-hold as you see fit. 

/lgtm

thc1006:
/hold cancel

Thanks @pravk03, and thanks for checking on @ffromani's threads rather than leaving it to me to guess.

Un-holding. #274 is stacked on this one and carries a hold of mine purely for merge order, which I will lift once this lands; #282 follows after that and is the one that closes the issue.

PR #273, opened 2026-07-31
fix: refuse a config key spelled differently from the schema

thc1006:
#### What type of PR is this?

/kind bug

`encoding/json` matches a struct field without regard to case, so `reservedcpus` in a config file sets `ReservedCPUs` just as `reservedCPUs` does. Everything around the decoder compares exactly, and two places depend on that.

`rejectExcludedFields` looks its names up directly, so a folded spelling sets a field the schema excludes from the file: `bindaddress: ":9999"` reaches `BindAddress`, although `bindAddress` is refused.

`Load` deletes the keys of explicitly-passed flags so the file cannot override them, and that delete is by exact name too. A file with `reservedcpus: "2-3"` survives the precedence pass and overwrites `--reserved-cpus=0-1`, which is the opposite of what the flag help and the configuration guidelines say happens.

Folding both comparisons would close those two but leave a file setting `reservedCPUs` and `reservedcpus` decoding both into one field, with whichever the decoder applies last winning. A file that reads one way and loads another is refused instead, and the message says which spelling to use.

The canonical set comes from `Config`'s own json tags, so a field added later cannot be forgotten here. A name matching no field at all is left to the decoder, which already rejects it through `DisallowUnknownFields`.

None filed; found while reviewing #272. It should land before #272: that PR adds `kubeletRootDir` to the excluded fields, and until this one is in, a config file spelling it `kubeletrootdir` still reaches the field and outranks an explicitly passed `--kubelet-root-dir`, which is exactly the driver-versus-mounts divergence #272 exists to prevent. That path is a hand-written `--config` file; the chart's values schema rejects the key in either spelling.

This rejects config files that load today, in two ways. A file using a folded spelling currently works by accident, and a file setting the same key twice currently keeps one of them in an undefined order. Both now fail at startup, closed rather than silently applying a value the reader did not intend.

Helm normally catches a folded key through values schema validation (`additionalProperties: false` on `driverConfig`). The driver-side check still matters for a raw `--config` file, and for a Helm install or upgrade run with `--skip-schema-validation`: I checked, and `--set-string driverConfig.reservedcpus=2-3` under that flag puts `reservedcpus: 2-3` into the generated ConfigMap verbatim, where this change is what stops it.

The canonical set is Config's own explicit json tags on its direct fields. `encoding/json` resolves more than that, so a field that was embedded, untagged, or tagged `json:"-,"` would be settable through a file while this check had never heard of it. That is enforced by a test over `Config` rather than reimplemented here, so it fails at build time the moment `Config` grows a shape the helper does not cover.

#### Release note

```release-note
Config files now reject keys whose letter case differs from the schema, as well as duplicate YAML keys. A non-canonical key can no longer bypass excluded-field checks or override an explicitly passed command-line flag.
```

#### Documentation updates

NONE

### DIFF
--- a/internal/driverconfig/cfgfile.go
+++ b/internal/driverconfig/cfgfile.go
@@ -19,8 +19,13 @@ package driverconfig
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"maps"
 	"os"
+	"reflect"
+	"slices"
+	"strings"
 
 	"sigs.k8s.io/yaml"
 )
@@ -38,6 +43,12 @@ func buildConfMap(filePath string) (map[string]any, error) {
 	}
 	delete(confMap, "apiVersion")
 
+	// Before anything reads the map by name: the decoder folds case, the checks
+	// below do not.
+	if err := rejectNonCanonicalKeys(confMap); err != nil {
+		return nil, err
+	}
+
 	if err := rejectExcludedFields(confMap); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


internal/driverconfig/config_test.go
```diff
@@ -107,6 +107,202 @@ cpuDeviceMode: individual
...
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), `use "apiVersion"`)
+	assert.NotContains(t, err.Error(), "unknown field",
+		"a schema property spelled wrongly should be corrected, not reported as unknown")
+}
+
+// TestLoad_RejectsCaseFoldedDuplicateKeys: two keys that decode into one field.
+// Accepting both leaves whichever the decoder happens to apply last, which is
+// not something a reader of the file can predict.
+func TestLoad_RejectsCaseFoldedDuplicateKeys(t *testing.T) {
```
pravk03: nit: we can use table-driven tests for the new tests added.

internal/driverconfig/cfgfile.go
```diff
@@ -38,21 +43,112 @@ func buildConfMap(filePath string) (map[string]any, error) {
...
+// DisallowUnknownFields and reports it with its own wording.
+func rejectNonCanonicalKeys(confMap map[string]any) error {
+	canonical := slices.Sorted(maps.Keys(canonicalConfigKeys()))
+	var problems []string
+	// Sorted, and every key this check rejects reported rather than the first
+	// one found. Ranging a map gives a different key each run, so one file
+	// produced a different message on each node and each fix took another
+	// restart. Only this check batches: an excluded field, an unknown field, a
+	// bad apiVersion and a duplicate key are each found by a different pass, and
+	// each reports one problem and stops.
```
pravk03: nit: some of the comments can be shortened/removed as the code changes are self explanatory.

pravk03 (commented):
Thanks a lot for fixing this. 

/hold 
(for fixing the minor nits) 

/lgtm
/approve

internal/driverconfig/cfgfile.go
thc1006: Done, and you were right about the volume. The comments are down to about half: what is left on `rejectNonCanonicalKeys` and `canonicalConfigKeys` is the part the code cannot say for itself, and the rest is gone. Measured against `load.go` and `schema.go` afterwards, the added lines are much closer to the surrounding density than they were.

internal/driverconfig/config_test.go
thc1006: Folded into one table. Seven of the eight were the same shape, so they are now rows with `wantContains` and `wantNotContains`.

`TestLoad_ReportsEveryMiscasedKey` stayed on its own: it asserts the whole error string, including the temporary file path, which does not sit well in a static table and is the point of that particular test.

thc1006:
Both nits addressed in `a889ebd`, which drops the `lgtm`. Could you take another look when you get a chance?

One thing worth flagging while you are here: this should merge before #272. That PR adds `kubeletRootDir` to the excluded fields, and until this one is in, a config file spelling it `kubeletrootdir` still reaches the field and outranks an explicitly passed `--kubelet-root-dir`. #272 and #274 are on `/hold` for that reason.

fmuyassarov (commented):
/lgtm
Thanks @thc1006 

I guess we are good to continue reviewing https://github.com/kubernetes-sigs/dra-driver-cpu/pull/272 after this lands right?

Issue #255, opened 2026-07-25
Chart does not validate kubelet plugin hostPaths as directories


thc1006:
Found while reviewing #244; pre-existing chart behavior, and hardening rather than a confirmed bug.

The `device-plugin` and `plugin-registry` volumes in `deployment/helm/dra-driver-cpu/templates/daemonset.yaml` set only `path`, with no `hostPath.type`. An empty type means the kubelet does not check, before mounting, that the path exists and is a directory (the type is left empty for backward compatibility).

`type: Directory` would give earlier validation when a configured kubelet root is missing or is not a directory, instead of the problem surfacing later. I am not claiming a specific downstream symptom: the exact result of a wrong root depends on the CRI/runtime and on whether the path already exists, and I have not reproduced it. `DirectoryOrCreate` would be the wrong choice, since creating a fresh directory under a wrong root hides the fact that the kubelet is not watching it — and neither type catches a wrong-but-existing root. Worth considering as a fail-fast improvement.

PR #274, opened 2026-07-31
feat(helm): render the kubelet paths from one chart value

thc1006:
#### What type of PR is this?

/kind feature

The chart hardcodes `/var/lib/kubelet` in four places: the two hostPath volumes and the two mountPaths that pair with them. A cluster whose kubelet runs with a different `--root-dir` cannot be served by the chart at all, even with #272's flag available, because the mounts still point at the default.

Adds `kubeletRootDir`, defaulting to the current path so nothing changes for anyone who does not set it, and renders `--kubelet-root-dir` and all four paths from it. At the default the objects the API server receives are unchanged and no flag is emitted at all; the only textual difference is that the four paths are now quoted.

Setting the flag through `extraArgs` alone is not enough and never was: the mounts stay behind, which is how the socket ends up outside them.

Empty strings and relative values are refused rather than defaulted, at the values schema and again in the template for anyone running with `--skip-schema-validation`. An omitted key, and a key set to YAML `null`, both select `/var/lib/kubelet`: Helm drops a null while coalescing values, so nothing downstream can tell it apart from a release installed before this value existed. The driver refuses an empty `--kubelet-root-dir` for the same reason: a value that rendered to nothing means the mounts rendered from it are wrong too.

Part of #231. Second of three: #272 landed the flag this renders, and the third now carries all of the new coverage, so the issue stays open until that lands.

#273 closed one more way around this, though it was not a code dependency. A config file supplied through `extraArgs: --config=...`, which takes effect when `driverConfig` is empty and the chart appends no `--config` of its own, can spell the key `kubeletrootdir`: the exact-name exclusion misses it and it still decodes into `KubeletRootDir`, moving the driver root while the mounts stay put. #273 closes that generically.

The render check moved to #282. @fmuyassarov asked for it: keeping the new coverage for how the driver is installed in one place, and it was the last thing here he had not settled ("I don't have anything blocking this PR apart from the helm-render-check.sh"). It also gives the script its own review rather than reviewing it as an appendix to a chart change.

Worth naming rather than leaving to the diff, since it is the price of the move: while #282 is in flight, nothing on this PR checks that the chart's `dra.cpu-reg.sock` still follows `driverName` in `cmd/dracpu/app.go`, that the four copies of the default root agree with each other, or that the template's own refusals still hold under `--skip-schema-validation`. The default root does keep its end-to-end coverage, since `deployment/helm/**` triggers the helm e2e job and that installs the chart into kind and waits for ResourceSlices; it is the relocated root and the refusals that wait. The check itself is unchanged and still passes against this branch, so it arrives in #282 as it was, not as something to be repaired there.

The Go half keeps its tests here, because `config_test.go` and `schema_internal_test.go` cover what the config file tells you to do instead, which is not part of how the driver gets installed.

`values.schema.json` and the chart README are regenerated, not hand-edited.

Not included, deliberately: #255 asks for `type: Directory` on the two kubelet hostPaths so a wrong root fails early rather than silently creating directories. That is a good idea and it becomes more useful once the root is configurable, but it is its own change with its own failure modes, so I left it out rather than bundling.

#### Release note

```release-note
The Helm chart now takes a `kubeletRootDir` value, so clusters whose kubelet uses a non-default `--root-dir` can run the driver. It defaults to `/var/lib/kubelet`, and existing installations are unaffected.

Passing `--kubelet-root-dir` through `extraArgs` is now refused when the chart renders. Anyone who reached a relocated root that way, with a custom image and hand-written matching mounts, should move the value to the top-level `kubeletRootDir` and drop the duplicate mounts.
```

#### Documentation updates

`docs/user/configuration.md`: `--kubelet-root-dir` added to the non-deprecated flags, and a section on using a custom kubelet root directory.

#### AI usage


thc1006:
Rebased onto the latest #272. Three changes since the last round.

The template now checks the registrar socket path in bytes. The values schema's `maxLength` counts characters, so a 26-character UTF-8 root can be 76 bytes and pass it; the schema is also skipped by `--skip-schema-validation`. The driver makes the same check and is the one that has to be right, so this is the earlier failure rather than the guarantee.

Surrounding whitespace is refused rather than trimmed. Quoting all five interpolations removed the reason trimming was there, and a trailing space is legal in a Linux pathname, so trimming it would give mounts and a flag that agree with each other and with the kubelet on nothing.

The invalid-root render cases now run both with and without `--skip-schema-validation`. That is what showed the previous round was only exercising the values schema, and that two of the template guards were not there at all.

/hold

Same reason as #272: this contains it, so it inherits the #273 ordering.

thc1006:
Dropped `maxLength: 73` from the values schema. It was a second copy of a limit that disagreed with the real one in both directions: it counts characters, so a 76-byte UTF-8 root passed it, and it sees the raw value, so `/var/lib/kubelet/x/../x/../...` was refused even though it cleans to `/var/lib/kubelet`. The result was that the same file loaded or not depending on whether `--skip-schema-validation` was passed. The template checks the cleaned path in bytes and the driver makes the same check, so nothing is lost by removing it.

The relocated-root docs now say which image to install with. The flag is newer than the chart's `appVersion`, so a source checkout with a custom root needs `image.repository` and `image.tag` pointed at a build that knows it. A default install is unaffected, since no flag is emitted at the default root.

The render checks also report a failed render as its own failure now, rather than letting it read as a missing mount, and assert that a custom root leaves no `/var/lib/kubelet` plugin path behind. That last one is what would catch a half-migrated template that added the new mounts while keeping an old one.

thc1006:
TLDR: rebased onto the updated #272, and applied @pravk03's two comments from there to this PR as well, so they do not need making twice.

#272 was rewritten to fold its nine single-case tests into tables and to cut its comments. That produced one conflict here, in `TestLoad_ExcludedFieldInFileIsError`: this PR changes the remediation for `kubeletRootDir` to name the chart value first and the flag for direct binary use. Resolved by keeping the new table and moving that change into its row, so the row asserts all three strings.

Ahead of the same comments landing here, the block comment in `_helpers.tpl` is down from 49 lines to 25 and `hack/ci/helm-render-check.sh` from 76 comment lines to 55. What is left is the part that is not in the code: why an empty or whitespace-bearing root is refused instead of trimmed, where the 73-byte budget comes from and why it is not repeated in the values schema, and why the extraArgs guard matches the parsed flag name.

All 66 render checks pass, `helm lint` is clean, and so are `make test-unit` and `golangci-lint run ./...`.

The hold here is mine and is only for merge order behind #272. I have also added the AI usage note the contributor guide asks for to the description.

thc1006:
/hold cancel

#272 has landed, so the merge-order hold here is no longer needed. This one is based on it and needs no rebase, since #272 went in as a merge commit.

@fmuyassarov you mentioned on #272 that you were looking forward to the user-facing half. This is it: both hostPath mounts and the driver flag now render from one chart value, and `hack/ci/helm-render-check.sh` covers the ways those two could drift apart, without needing a cluster.

#282 follows this one and is where the issue is closed.

thc1006:
TLDR: second of the three, and the two comments you left on #272 are already applied here, so there should be nothing to repeat.

The chart renders both kubelet hostPath mounts and the driver flag from one `kubeletRootDir` value, defaulting to the current path, so an existing install is unaffected. `hack/ci/helm-render-check.sh` covers the ways those two could drift apart, without needing a cluster.

@pravk03 a pass whenever you have time would be very welcome. #282 is the last of the three and is where #231 gets closed; it is waiting behind this one.

docs/user/configuration.md
```diff
@@ -155,7 +155,6 @@ The remaining flags aren't part of that deprecation:
 - `--config`: path to the config file described above.
 - `--kubeconfig`: path to a kubeconfig file, for out-of-cluster use. Also settable via the `kubeconfig` config field.
 - `--bind-address`: address the metrics server listens on.
-- `--show-metrics`: print the metrics endpoint and exit.
```
fmuyassarov, line 158: let's not do it here, this change doesn't belong to the PR

internal/driverconfig/config.go
```diff
@@ -29,6 +29,11 @@ type Config struct {
 	GroupBy          string `json:"groupBy,omitempty"`
 	ExposePCIeRoots  bool   `json:"exposePCIeRoots,omitempty"`
 	SysFSOverlay     string `json:"sysfsOverlay,omitempty"`
+	// KubeletRootDir is the kubelet root directory. The plugin registration and
+	// plugins directories are derived from it as <root>/plugins_registry and
+	// <root>/plugins. Set it when the kubelet --root-dir is not the default
+	// /var/lib/kubelet.
```
fmuyassarov: nit, I guess there is enough explanation around this flag, this seems redundant. Can we remove it?

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
```diff
@@ -69,6 +69,18 @@ spec:
...
+          {{- $kubeletRoot := include "dra-driver-cpu.kubeletRootDir" . }}
+          {{- if ne $kubeletRoot "/var/lib/kubelet" }}
+          {{- /* Emitted only when the root has been moved. The flag is newer than
+                 the chart's appVersion, so passing it to that image would make the
+                 documented local-checkout install exit on an unknown flag on every
+                 node. Anyone moving the root needs a newer image anyway, and anyone
+                 who has not gets no flag, no unknown-flag exit, and no rollout for
+                 a value identical to the driver's own default. The literal is
+                 DefaultKubeletRootDir in internal/driverconfig/defaults.go; the
+                 chart default in values.yaml is the same string by design. */}}
```
fmuyassarov: this will never show up in any user facing output, so this might be only visible to the maintainers, which I think isn't that needed as we have it described well in the docs

docs/user/configuration.md
```diff
@@ -164,6 +163,16 @@ The remaining flags aren't part of that deprecation:
...
+- `--kubelet-root-dir`: the kubelet's own root directory, `/var/lib/kubelet` unless the
+  kubelet was started with a different `--root-dir`. Set it through the chart's
+  `kubeletRootDir` rather than directly, so the hostPath mounts move with it. The driver registers under
+  `<root>/plugins_registry` and creates its socket under `<root>/plugins/<driver-name>`,
+  and the kubelet only watches its own root, so a mismatch leaves the driver running with
+  nothing ever noticing it. Like `--expose-pcie-roots` it is not deprecated and has no
+  config file equivalent: the Helm chart renders the hostPath mounts from the same value
+  (`kubeletRootDir`), and a config file the chart does not write could not keep the two in
+  step. An empty value is refused rather than defaulted, since the way it arrives empty is
+  a chart that rendered nothing into it.
```
fmuyassarov: probably I'm being too picky about docs, if so sorry, but I think there is a little room to make it more easier for readers
```md
- `--kubelet-root-dir`: the kubelet's own root directory (default `/var/lib/kubelet`).
  Set via the chart's `kubeletRootDir` value — do not pass this flag directly or through
  `extraArgs`, since the hostPath mounts won't move with it, and a mismatch between the
  two leaves the driver registered somewhere the kubelet isn't watching, with nothing
  reporting the failure.

  Like `--expose-pcie-roots`, it is not deprecated and has no config file equivalent:
  the chart owns this value, not `driverConfig`.
```

docs/user/configuration.md
```diff
@@ -174,3 +183,46 @@ one can be enabled at a time on any given node. You need to disable the CPUManag
...
+two releases is not a way around that: the chart's `DeviceClass` is cluster-scoped under a
+fixed name, and Helm refuses to hand an object owned by one release to another
+(`invalid ownership metadata`). Until the chart can vary the root per node, a cluster whose
+nodes disagree needs the driver deployed by something other than a single release of this
+chart.
+
+Passing `--kubelet-root-dir` through `extraArgs` is refused at template time. The flag
+would win, because Go's flag package takes the last value and `extraArgs` is appended
+after, while the mounts would keep rendering from `kubeletRootDir` and stay behind. Set
+`kubeletRootDir` instead.
```
fmuyassarov: I think we can simply this too a little bit
```md
If the kubelet runs with `--root-dir` set to something other than `/var/lib/kubelet`, set
`kubeletRootDir` to that root's resolved absolute path (relative paths are refused, since a
hostPath can't follow the kubelet's working directory):

```yaml
# values.yaml
kubeletRootDir: /mnt/data/kubelet
```

That one value becomes the driver's `--kubelet-root-dir` and both hostPath mounts.

The flag is newer than the chart's `appVersion`, which still selects the v0.2.0 image, so a
source-checkout install with a relocated root also needs an image built from a revision that
knows the flag — otherwise every node exits on an unknown flag:

```bash
helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system \
  --set-string kubeletRootDir=/mnt/data/kubelet \
  --set-string image.repository=REGISTRY/REPOSITORY \
  --set-string image.tag=IMAGE_TAG
```

A released chart's `appVersion` already knows the flag, and a default install (root left
unset) emits no flag at all — neither needs this.

Once release's DaemonSet lands on every node it selects, so those nodes must agree on the
kubelet root; splitting them across releases fails with `invalid ownership metadata`, since
the chart's `DeviceClass` is cluster-scoped under a fixed name. A cluster whose nodes
disagree needs the driver deployed some other way until the chart can vary the root per
node.

Passing `--kubelet-root-dir` through `extraArgs` is refused at template time — set
`kubeletRootDir` instead, since `extraArgs` would win the flag while the mounts kept
rendering from the chart value.
```

File: docs/user/configuration.md [hunk 6]
```diff
@@ -174,3 +183,46 @@ one can be enabled at a time on any given node. You need to disable the CPUManag
 1. Make sure `cpuManagerPolicy: "none"` is set in the kubelet [configuration file](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/).
 1. If you changed the kubelet configuration, restart the kubelet to take effect. **NOTE:** you may need to [delete the CPUManager state file](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#changing-the-cpu-manager-policy).
 1. You may now proceed with deploying and configuring this DRA driver.
+
+### A relocated kubelet root directory
```
fmuyassarov: Maybe something like ?
`Using a custom kubelet root directory`

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
```diff
@@ -50,6 +50,77 @@ app.kubernetes.io/name: {{ include "dra-driver-cpu.name" . }}
...
+{{- if not (isAbs $root) -}}
+{{- fail (printf "kubeletRootDir must be an absolute path, got %q" $root) -}}
+{{- end -}}
+{{- $cleaned := clean $root -}}
+{{- $socket := clean (printf "%s/plugins_registry/dra.cpu-reg.sock" $cleaned) -}}
+{{- if gt (len $socket) 107 -}}
+{{- fail (printf "kubelet registrar socket path %q is %d bytes, over the 107-byte limit for a Unix socket path: kubeletRootDir has 73 bytes to spend and is using %d" $socket (len $socket) (len $cleaned)) -}}
+{{- end -}}
+{{- $cleaned -}}
+{{- end }}
```
fmuyassarov, line 103: I think we can simplify this a bit as well. Since Github doesn't show the diff very nicely here, I added it as a separate commit. https://github.com/kubernetes-sigs/dra-driver-cpu/commit/47ced2feffcfe3d34c6195307ed72cacf4057cd7

deployment/helm/dra-driver-cpu/values.yaml
```diff
@@ -94,6 +94,13 @@ healthzPath: /healthz
 # -- Port the HTTP server binds to; used for the container port and probes
 healthzPort: 8080 # @schema type:integer;minimum:1;maximum:65535
 
+# -- Kubelet root directory, matching the kubelet's own `--root-dir`. The driver
+# registers under `<root>/plugins_registry` and creates its socket under
+# `<root>/plugins/<driver-name>`, and the hostPath mounts come from this same
+# value, so the two cannot come apart. Only set this if the kubelet does not use
+# the default.
```
fmuyassarov: how do you feel about something like
```yaml
# -- Kubelet base directory (the kubelet's own `--root-dir`). Only set this if
# the kubelet does not use the default. The hostPath mounts and the driver flag
# both derive from this value, so they cannot come apart. See docs/user/configuration.md for details
```

fmuyassarov (commented):
This looks pretty close to be fully ready @thc1006  and thanks for working on it. I've left some comments, mostly asking for some cosmetic changes but not the implementation vice. Also, I didn't review the testing part yet.

docs/user/configuration.md
thc1006: Restored, thanks. The line is stale in main (the flag is gone, and `test/e2e_local/metrics_test.go` asserts it is rejected), so I will send that removal on its own.

internal/driverconfig/config.go
thc1006: Removed. None of the other fields in the struct carry a comment either, so it was the odd one out.

deployment/helm/dra-driver-cpu/templates/daemonset.yaml
thc1006: Gone, your cleanup commit took it out along the way.

docs/user/configuration.md
thc1006: Taken as written, and it does read better.

One consequence worth naming: between this, the values.yaml comment and the Go field comment, `<root>/plugins_registry` no longer appears anywhere outside the templates. I think that is fine, since it is the kubelet's own layout rather than something anyone sets, but it is easy to put back in one place if you would rather keep it.

docs/user/configuration.md
thc1006: Taken as written, reading `Once release's` as `One release's`.

docs/user/configuration.md
thc1006: Done.

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
thc1006: Cherry-picked as it is, so it stays yours, and it now sits on top of my commit.

Passing the validated root in rather than the whole context is better than what I had: the validation runs once per render instead of once at each mount. `helm template` comes out byte-identical before and after, at the default root, at a relocated one, and at `/`.

deployment/helm/dra-driver-cpu/values.yaml
thc1006: Taken as written. `make helm-docs` and `make helm-schema` re-run, so the chart README and values.schema.json carry the same wording.

thc1006:
TLDR: all eight taken, and your helpers commit is cherry-picked on top so it stays yours.

The docs are shorter for it. One follow-up rather than a change here: the `--show-metrics` line is restored, but it is stale in main, so I will send that removal on its own.

`make helm-lint`, `hack/ci/helm-render-check.sh`, `make helm-docs-check`, `make lint` and the unit tests are green, and `helm template` comes out byte-identical across your helpers change at the default root, at a relocated one, and at `/`.

docs/user/configuration.md
thc1006: Correction to what I said above: I had to join your two paragraphs into one. mdformat reads a list item with a blank line inside it as a loose list and then spaces out every other item, which would have reformatted the four bullets above this one. Same words, one paragraph.

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
thc1006: One line from the old comment I would keep somewhere, here or as a follow-up: the socket name in the byte check is spelled out because the chart cannot ask the binary for it, so renaming `driverName` in `cmd/dracpu/app.go:50` moves the 73-byte budget. Nothing catches that today, since `hack/ci/helm-render-check.sh` pins the boundary against the same literal and would stay green.

fmuyassarov:
also, I don't know how to mark it otherwise, maybe feature? but it is not really bug IMO

fmuyassarov:
Thanks @thc1006 for addressing the comments. I will do my second round of reviews soon and testing part. I would appreciate it, if you could also reduce the commit message a bit so that it is easier to follow in the future when going trough the git history. Doesn't have to be done now but as part of addressing comments

fmuyassarov:
> Thanks @thc1006 for addressing the comments. I will do my second round of reviews soon and testing part. I would appreciate it, if you could also reduce the commit message a bit so that it is easier to follow in the future when going trough the git history. Doesn't have to be done now but as part of addressing comments

and + I don't think we need a separate commit (my proposed changes for the helpers). Better and cleaner if you take them into your first commit.

thc1006:
All three done.

`/kind feature` fits and I have switched it. Worth knowing rather than a disagreement: #272 and #273 both went in as `kind/bug`, so the series reads mixed now.

The commit message is down from about 1500 words to under 300. What is left is why the chart could not serve a relocated root, that the default is unchanged, why the flag is only emitted once the root moves, and where the guards are checked. The rest belongs in the description and in the template comments, and one line of it had gone stale: it still said the `--show-metrics` entry went with this change, which it no longer does.

Your helpers change is folded into that commit with `Co-authored-by` and your sign-off kept. I rebased onto main while rewriting, so the pull request is one commit now instead of also carrying #272's.

The rewrite moved no content: the diff is identical to the last push, checked line by line. `make helm-lint`, `hack/ci/helm-render-check.sh`, `make helm-docs-check`, `make lint`, `pre-commit run --all-files` and the unit tests are green.

docs/user/configuration.md
```diff
@@ -174,3 +180,40 @@ one can be enabled at a time on any given node. You need to disable the CPUManag
...
+helm install dra-driver-cpu ./deployment/helm/dra-driver-cpu -n kube-system \
+  --set-string kubeletRootDir=/mnt/data/kubelet \
+  --set-string image.repository=REGISTRY/REPOSITORY \
+  --set-string image.tag=IMAGE_TAG
+```
+
+A released chart's `appVersion` already knows the flag, and a default install (root left
+unset) emits no flag at all — neither needs this.
+
+One release's DaemonSet lands on every node it selects, so those nodes must agree on the
```
fmuyassarov:
```suggestion
Once release's DaemonSet lands on every node it selects, so those nodes must agree on the
```

thc1006:
Pushed three follow-ups.

The root-wide `additionalProperties: false` is out. A parent declaring this chart with `condition: dra-driver-cpu.enabled` passes `enabled` into the child values, Helm validates those against the child schema, and `helm template` fails with `additional properties 'enabled' not allowed`. The kubelet root does not need the closure: the template refuses an empty, relative, whitespace-padded or over-long root itself, and names a key differing from `kubeletRootDir` only in case, all of which hold under `--skip-schema-validation`. The release note lost the sentence about unknown top-level values with it.

The render check now reads `driverName` and asserts the chart's registrar socket follows it, the way the default root is already checked across its three spellings, and it covers a root that cleans back to the default, where the flag has to stay off for the appVersion image.

The configuration page refused `--kubelet-root-dir` outright while the config error tells anyone running the binary to use it, so that refusal is scoped to Helm now.

thc1006:
Folded the follow-ups back into one commit and shortened the message, as you asked. The implementation is the same as the version you last saw, apart from one comment: the render check still described the closed root schema I removed, and the template is what refuses a miscased key in both modes now.

The release note gained the `extraArgs` half. Refusing `--kubelet-root-dir` there is a change anyone who reached a relocated root that way, with a custom image and hand-written mounts, will notice, so it should not be left to the diff.

#282 is on hold and still carries the older version of this branch, including the schema change. I will rebuild it from main once this lands rather than rebase it, so only the workflow and the kind config remain.

@fmuyassarov the testing pass should be ready whenever you have time.

thc1006:
One more, on upgrades of a release older than this value, and it is worse than a warning: the upgrade fails outright.

`helm upgrade --reuse-values` passes the values the release was installed with as the whole override set, so a chart's own defaults are never coalesced in for a key that release never had. A v0.2.0 release has no `kubeletRootDir`, the helper read the absent key as empty, and the render stopped at `kubeletRootDir must not be empty`. Reproduced by rendering a copy of the chart with the key taken out, which is the same state:

```
Error: execution error at (dra-driver-cpu/templates/daemonset.yaml:15:20): kubeletRootDir must not be empty
```

`d3ccaea` separates absent from empty. Absent reads as `/var/lib/kubelet`, which is how those releases already ran, and renders the standard mounts with no flag. Set and empty still fails, since that one was written down.

The render check covers it now, on a copy with the key removed, because `helm template` always coalesces the chart's defaults and cannot reach the state otherwise. Putting the default in the helper made a fourth place that has to agree with the driver constant, the chart value and the template, so the check that compares them compares four now. Reverting the helper fails both.

The release note still says existing installations are unaffected. That was true of what they run and not of how they upgrade, so it now says what to do about `extraArgs` as well. Let me know if you would rather it called out `--reuse-values` too.

thc1006:
Two edge cases from another pass over the fixtures, both in `a883ec1` (the single commit, amended).

Helm removes a key set to YAML `null` while coalescing values, before the schema is applied, so `kubeletRootDir: null` reaches the template exactly as a pre-field release under `--reuse-values` does. Nothing downstream can tell the two apart, so both take `/var/lib/kubelet`. That is now stated rather than implied: the helper comment, the configuration doc and the PR description say so, and the render suite pins it next to the absent-key case. An explicit empty string is still refused, which the existing `empty root` case covers in both schema modes.

The legacy fixture dropped the key with `sed -i '/^kubeletRootDir:/d'`. That form takes a backup suffix on BSD sed, so it does not run on macOS, and it is line-oriented: quote or indent the key in `values.yaml` one day and the edit matches nothing, the copy keeps the chart default, and the case passes without ever reaching the path it exists for. It now uses the `yq` the target already installs, and asserts `has("kubeletRootDir")` is false before rendering. Pointing the deletion at a key that is not there turns the case red instead of green.

`helm-lint`, `helm-render-check`, `helm-docs-check`, `helm-schema-check` and `test-unit` pass, and the render suite is 73 PASS on both Helm 3.20.2 and 4.2.3. `shellcheck --severity=warning` at the pinned v0.10.0, `shfmt -s`, `mdformat` and `codespell` are clean on the files touched.

@fmuyassarov this should be ready for the testing pass whenever it suits you.

pravk03:
Thanks for the changes. 
Took a pass at this. Did not have anything to add beyond what @fmuyassarov has already covered. Will wait for @fmuyassarov's  lgtm, happy to approve after.

docs/user/configuration.md
thc1006: One word here no longer matches yours either. It did when I replied, and changed with the rework two days later. I dropped `resolved` from `that root's resolved absolute path` and added a sentence saying the path is cleaned rather than resolved.

The template runs `clean`, so `..` and repeated separators collapse, but a symlinked root is left as it is for the kubelet to follow. Saying resolved would promise something the chart does not do. Everything else in that block is yours as written.

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
thc1006: This one is closed, and by something better than the comment I was asking to keep.

`hack/ci/helm-render-check.sh` now reads `driverName` out of `cmd/dracpu/app.go` and fails if the chart's registrar socket name stops following it. Renaming it moves the 73-byte budget and CI says so, instead of staying green because the check pinned the same literal the chart did. Nothing outstanding here from my side.

docs/user/configuration.md
thc1006: Flagging that this paragraph no longer reads as you wrote it. It did when I replied; it changed two days later when I reworked the branch, and I should have said so then.

Yours said not to pass the flag directly or through `extraArgs`. That is right under Helm and wrong outside it, since the flag is the only way to set the root when the binary is run directly or the manifests are written by hand, and #272 added it for exactly that. So the prohibition is now scoped to Helm and a sentence covers the direct case, together with the two mounts that have to match it. The rest of the paragraph is yours.

.github/workflows/helm-lint.yaml
```diff
@@ -5,7 +5,11 @@ on:
     paths:
       - 'deployment/helm/**'
       - 'internal/driverconfig/**'
+      - 'cmd/dracpu/app.go'
+      - 'pkg/driver/driver.go'
```
fmuyassarov: Do we really needs these to trigger Helm linter?

fmuyassarov (commented):
Thanks @thc1006 for the update. I have two last nits below and other than those I don’t have anything blocking this PR apart from the helm-render-check.sh

I think the additional testing here covers more than just the directory protection, which is definitely useful. That said, would you mind decoupling those tests from this PR and perhaps moving them to the e2e patch https://github.com/kubernetes-sigs/dra-driver-cpu/pull/282 ? My thinking is that it would be nice to keep all the new testing around the driver installation method together in one place. Also, I must admit I’m still trying to fully understand a few parts of the bash script. I will have some follow up questions there as I go through it.

fmuyassarov (commented):
/lgtm 
Thank you.

fmuyassarov:
@pravk03 I think we are good now here.

.github/workflows/helm-lint.yaml
thc1006: No, and this file is back to exactly what main has. Both entries were there for the render check, which has moved to #282 together with the step that ran it and the `make helm-render-check` target.

They were not equally justified either, so the two do not follow it equally. The check reads `driverName` out of `cmd/dracpu/app.go`, and renaming that is the change that silently moves the chart's byte budget for the root, so that entry earns its place and goes to #282 with the check. Nothing the job runs reads `pkg/driver/driver.go`. I had added it because the socket name is derived there, which is a reason to think about the coupling and not a reason to run the linter. That one is simply gone.

docs/user/configuration.md
thc1006: I have not taken this one as written, and I would rather say so than leave it quietly unapplied.

The sentence is about one release covering many nodes, so its subject is `One release's DaemonSet`. `Once release's` reads as a typo for that, and the nearest grammatical reading of `Once` would change the claim: the nodes have to agree from the install, not from the moment the rollout finishes landing.

You have flagged this spot twice, though, so the opener is clearly not doing its job. It now reads `A single release's DaemonSet lands on every node it selects`, which cannot be taken either way. If `Once` did mean something I have missed, tell me and I will write that instead.

thc1006:
TLDR: the render check is #282's now, this branch is down to eleven files, and both nits are answered above.

Taking the ask first. `hack/ci/helm-render-check.sh` moved to #282 unchanged, with the `make helm-render-check` target and the workflow step that ran it, so `.github/workflows/helm-lint.yaml` and the `Makefile` are back to what main has and drop out of this diff entirely. #282's description now reads the script block by block, since you said you were still working through parts of it. It may save you some of the questions, and it should at least make the rest easier to ask against something named.

One thing this reopens, which I would rather tell you than have you find. On the 9th I said the `driverName` coupling was closed and that nothing was outstanding from my side, because the check had started reading `driverName` out of `cmd/dracpu/app.go` instead of pinning the same literal the chart pins. That check is no longer on this branch, so on this branch the coupling is unguarded again: renaming `driverName` moves the 73-byte budget in the driver while the chart's spelled-out `dra.cpu-reg.sock` stays where it is. #282 closes it again.

What the move costs while #282 is in flight is worth being exact about, because it is less than it sounds. `deployment/helm/**` still triggers the helm e2e job, so the default root keeps its end-to-end coverage: the chart is installed into kind and the ResourceSlices are waited for. What is left unguarded here is the relocated root, the template's own refusals under `--skip-schema-validation`, the four-way agreement on the default root, and the socket name above. The Go half keeps its tests here, since those are about what the config file tells you to do rather than about installing anything.

The check itself is not being repaired on the way over. I ran it against this branch after the move and every case passes, so it arrives in #282 as the thing you have already started reading.

Also rebased onto main, which had moved nine commits, and one unrelated tidy in the same push: a comment line in `_helpers.tpl` had been left four words wide by a later edit of mine, and is rewrapped. `helm lint --strict`, `helm-docs-check`, `helm-schema-check`, `make test-unit`, `golangci-lint run ./...` and `pre-commit run --all-files` are green on the rebased tree, and `helm template` is byte-identical across the rewrap at the default root, at a relocated one, and at `/`.

thc1006:
> /lgtm Thank you.

And also thx for hard work reviewing!!!

pravk03:
@thc1006  @fmuyassarov Thanks a lot for the changes and the reviews. 

/lgtm 
/approve

PR #276, opened 2026-07-31
docs: add quickstart and use-case examples, consolidate user docs

pravk03:
#### What type of PR is this?
/kind documentation

- Adds a `Usage` section to the README two target usecases: topology-aware CPU allocation and resource alignment.
- Adds `docs/user/quickstart.md`, a guided install, run, and verify walkthrough. 
- Adds `docs/user/device-attributes.md`, an attribute reference per device mode with CEL selector examples. It absorbs `resourceslices-examples.md`.
- Folds `prerequisites.md` into `installation.md` and adds compatibility,security, upgrade, and uninstall guidance.
- Folds `cpu-manager-mapping.md` into `feature-support.md`, next to the featurematrix it belongs with.

N/A

Note:  This PR was drafted with AI assistance. I have reviewed the content. 

#### Release note

```release-note
NONE
```


pravk03:
@ffromani @fmuyassarov I got feedback that the driver's target use cases (resource alignment, fine-grained CPU allocation) have no examples documented. I started with adding those but made a few more changes (AI assisted) along the way. PTAL.

docs/user/device-attributes.md
```diff
@@ -0,0 +1,296 @@
+# Device Attributes and Selectors
+
+Every CPU device the driver publishes carries topology attributes that claims can select on
+with [CEL expressions](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#device-selectors).
```
fmuyassarov: nit, wrong link?

```suggestion
Every CPU device the driver publishes carries topology attributes that claims can select on
with [CEL expressions](https://kubernetes.io/docs/reference/using-api/cel/).
```

fmuyassarov (commented):
Thanks @pravk03 . I have a few comments though most if not all are probably just personal preferences. So I'm not pushing hard on any of these, but I still wanted to share them. Also, adding `# Device Attributes and Selectors` was good.

1. Should we move the `Usage` section out of the root README? That would help keep it concise and avoid it gradually growing again. 
2. This is not directly about your PR but to me it seems that we could hide the `How It works` short paragraph from the main README and keep it within `how-it-works.md`. In other words, drop the short description https://github.com/kubernetes-sigs/dra-driver-cpu/blob/main/README.md#how-it-works 
3. Maybe drop `What you need` section in from the `quickstart.md` and have a ref to compatibility ? https://github.com/pravk03/dra-driver-cpu/blob/cc2e4e1557c41e454954f8d42e42e9830f79f131/docs/user/installation.md#compatibility

docs/user/quickstart.md
```diff
@@ -0,0 +1,154 @@
+# Quickstart
+
+This guide takes you from an empty cluster to a pod running on exclusive, pinned CPUs, with a
+verification step after each stage. It should take about ten minutes.
+
+## What you need
+
+- **Kubernetes 1.34 or newer.** DRA (`resource.k8s.io/v1`) is GA and enabled by default from
+  1.34. The default `grouped` device mode additionally uses DRA consumable capacity, enabled
+  by default from 1.36 — on 1.34/1.35, enable the `DRAConsumableCapacity` feature gate or use
```
AutuSnow: The individual-mode fallback is not actionable in the current quickstart: the install command still uses the default grouped mode, and the example claim uses capacity. Could we either show the required Helm override with a count-based claim, or require DRAConsumableCapacity for this walkthrough?

pravk03:
> 1. Should we move the Usage section out of the root README? That would help keep it concise and avoid it gradually growing again.
> 2. This is not directly about your PR but to me it seems that we could hide the How It works short paragraph from the main README and keep it within how-it-works.md. In other words, drop the short description https://github.com/kubernetes-sigs/dra-driver-cpu/blob/main/README.md#how-it-works

Thanks @fmuyassarov!. Personally, I think there is a lot of value in answering **What**, **Why**, and **How** directly on the main README so first-time visitors can immediately get a high understanding about the project before diving into the sub-pages. Having a target use-case with a simple example brings immediate familiarity for anyone who uses DRA, and a couple of points on the core components answer those initial questions right away.  Infact, these exact questions came up recently from a user who could benefit from this project, but wasn't sure if it was the right fit. That said, I totally agree with your concern about the README growing. I have made the `How It Works` section a little more concise, and I'm happy to be convinced otherwise if you feel strongly about moving anything out. 

> 3. Maybe drop What you need section in from the quickstart.md and have a ref to compatibility ? https://github.com/pravk03/dra-driver-cpu/blob/cc2e4e1557c41e454954f8d42e42e9830f79f131/docs/user/installation.md#compatibility

Good idea. I made some adjustments and added a pointer to installation.md.

docs/user/quickstart.md
pravk03: Good point. removed the individual mode fallback.

docs/user/device-attributes.md
pravk03: Thanks> Fixed.

fmuyassarov (commented):
Thank you @pravk03 and as said those were just some personal taste. 
/lgtm

PR #277, opened 2026-08-01
fix: reject grouped allocations with missing CPU capacity

### LINKED ISSUE #271: Grouped claim preparation succeeds when CPU consumed capacity is missing
**What happened**:

In grouped mode, claim preparation reads the requested CPU count from the `dra.cpu/cpu` entry in `AllocationResult.ConsumedCapacity`.

When an allocation result matches `Driver == "dra.cpu"` but does not contain that entry, `claimCPUCount` remains zero. Socket and NUMA grouping then produce an empty CPU assignment, and preparation returns success through the existing no-CPU-assignment path. No CDI device is created and no CPU allocation is recorded.

The no-CPU-assignment path appears intentional when a multi-driver claim has no allocation results for this driver. However, a matching `dra.cpu` allocation result without its CPU capacity seems to be a different case.

**What you expected to happen**:

If a matching grouped CPU allocation is required to contain `dra.cpu/cpu`, claim preparation should reject an allocation result where that capacity entry is missing instead of treating it as a successful zero-CPU allocation.

**How to reproduce it (as minimally and precisely as possible)**:

1. Configure the driver in grouped mode with either socket or NUMA grouping.
2. Call claim preparation with an allocation result whose driver is `dra.cpu` and whose device identifies a grouped CPU device.
3. Leave `dra.cpu/cpu` out of `ConsumedCapacity`.
4. Observe that preparation succeeds with an empty result.

A unit test can reproduce this by adding such an allocation result to the existing grouped-mode preparation tests and expecting an error. With the current implementation, the error is nil.

**Context / setup details**:

This was found through code inspection and a focused unit test rather than on a running cluster. Grouped CPU devices set `AllowMultipleAllocations=true`. The Kubernetes API documentation says that capacity entries are included for devices which allow multiple allocations.

Related issue #246 and PR #247 validate zero, negative, and fractional CPU capacity values when the entry is present, but do not cover a missing entry.

**Relevant logs or outputs**:

```text
Expected an error for a missing dra.cpu/cpu capacity entry, but got nil.
```

**Question / what you want to verify**:

Is the presence of `dra.cpu/cpu` for every matching grouped CPU allocation considered an invariant that the node preparation path should validate?

If rejecting the missing entry is the desired behavior, I would be happy to submit a small PR that:

- returns an error when a matching grouped CPU allocation is missing `dra.cpu/cpu`;
- preserves the existing behavior for claims without any allocation for this driver;
- adds a unit test for the missing-capacity case.

**Diagnostics (`dracpu-gatherinfo`)**:

```text
Not applicable; found through code inspection and a unit test.
```

**Environment**:
- DRA driver version/tag (image tag or commit SHA): current `main`
- DRA driver mode (`grouped` or `individual`): `grouped`
- DRA grouped mode setting (`numanode` or `socket`, if applicable): both
- Container runtime and version (optional): N/A
- Kubernetes version (use `kubectl version`): N/A
- Cloud provider or hardware configuration: N/A
- OS: N/A
- Kernel: N/A
- Network plugin and version: N/A
- Others: N/A

alanhuangch:
#### What type of PR is this?

/kind bug

Grouped CPU devices allow multiple allocations and therefore matching allocation results are expected to include the consumed `dra.cpu/cpu` capacity.

When that entry was missing, socket and NUMA grouping treated the allocation as requesting zero CPUs and returned success through the no-CPU-assignment path. This PR rejects such malformed grouped CPU allocations with an explicit error instead.

Allocation results belonging to other drivers continue to be ignored, preserving multi-driver claim behavior.

The PR adds regression coverage for:

- missing CPU capacity in socket grouping;
- missing CPU capacity in NUMA grouping;
- allocation results from other drivers without consumed CPU capacity.

Fixes #271

This is node-side validation only and does not change the device API. The validation applies to matching allocations in all grouped modes, whose devices set `AllowMultipleAllocations=true`.

Validation completed with:

```text
go test ./pkg/... ./internal/... ./api/... ./cmd/...
go test -race -count=1 ./pkg/driver
```

#### Release note

```release-note
NONE
```

#### Documentation updates

NONE

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -125,21 +125,21 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 	var cpuAssignment cpuset.CPUSet
 	sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
 	for _, alloc := range claim.Status.Allocation.Devices.Results {
-		claimCPUCount := int64(0)
 		if alloc.Driver != cp.driverName {
 			continue
 		}
-		if quantity, ok := alloc.ConsumedCapacity[device.CPUResourceQualifiedName]; ok {
-			if quantity.Sign() <= 0 {
-				return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q must be positive, got %s", alloc.Device, quantity.String())}
-			}
-			count := quantity.Value()
-			if quantity.CmpInt64(count) != 0 {
-				return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q must be a whole number, got %s", alloc.Device, quantity.String())}
-			}
-			claimCPUCount = count
-			logger.V(4).Info("found CPU request", "numCPUs", count, "device", alloc.Device)
+		quantity, ok := alloc.ConsumedCapacity[device.CPUResourceQualifiedName]
+		if !ok {
+			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q is missing", alloc.Device)}
 		}
+		if quantity.Sign() <= 0 {
+			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q must be positive, got %s", alloc.Device, quantity.String())}
+		}
+		claimCPUCount := quantity.Value()
+		if quantity.CmpInt64(claimCPUCount) != 0 {
+			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q must be a whole number, got %s", alloc.Device, quantity.String())}
+		}
+		logger.V(4).Info("found CPU request", "numCPUs", claimCPUCount, "device", alloc.Device)
 
 		topo := cp.topology.cpuTopology
 
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -130,7 +130,7 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		}
 		quantity, ok := alloc.ConsumedCapacity[device.CPUResourceQualifiedName]
 		if !ok {
-			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q is missing", alloc.Device)}
+			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity %q for device %q is missing", device.CPUResourceQualifiedName, alloc.Device)}
 		}
 		if quantity.Sign() <= 0 {
 			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q must be positive, got %s", alloc.Device, quantity.String())}
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -1239,68 +1239,19 @@ func TestPrepareResourceClaimsGroupedMode(t *testing.T) {
 
 func TestPrepareResourceClaimsGroupedModeRejectsInvalidCPUCapacity(t *testing.T) {
 	testCases := []struct {
-		name          string
-		capacity      string
-		expectedError string
+		name            string
+		groupBy         string
+		device          string
+		capacity        string
+		includeCapacity bool
+		expectedError   string
 	}{
-		{name: "fractional below one", capacity: "500m", expectedError: "must be a whole number"},
-		{name: "fractional above one", capacity: "1500m", expectedError: "must be a whole number"},
-		{name: "zero", capacity: "0", expectedError: "must be positive"},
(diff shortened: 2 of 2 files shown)


pkg/driver/dra_hooks.go
```diff
@@ -125,21 +125,21 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
...
-			}
-			count := quantity.Value()
-			if quantity.CmpInt64(count) != 0 {
-				return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q must be a whole number, got %s", alloc.Device, quantity.String())}
-			}
-			claimCPUCount = count
-			logger.V(4).Info("found CPU request", "numCPUs", count, "device", alloc.Device)
+		quantity, ok := alloc.ConsumedCapacity[device.CPUResourceQualifiedName]
+		if !ok {
+			return kubeletplugin.PrepareResult{Err: fmt.Errorf("CPU capacity for device %q is missing", alloc.Device)}
```
pravk03: nit, optional: we can include the missing capacity key in the error message

pkg/driver/dra_hooks_test.go
```diff
@@ -1293,6 +1293,95 @@ func TestPrepareResourceClaimsGroupedModeRejectsInvalidCPUCapacity(t *testing.T)
 	}
 }
 
+func TestPrepareResourceClaimsGroupedModeRejectsMissingCPUCapacity(t *testing.T) {
```
pravk03: nitL I think these new tests can be folded into TestPrepareResourceClaimsGroupedModeRejectsInvalidCPUCapacity as additional test cases.

pravk03 (commented):
Thanks for the fix. It makes sense since we were already checking if quantity was non-zero. Left a couple of minor nits, otherwise LGTM.

pkg/driver/dra_hooks.go
alanhuangch: Done in the new commit c203968. The error now includes the missing dra.cpu/cpu capacity key. Thanks!

pkg/driver/dra_hooks_test.go
alanhuangch: Done in c203968. I folded the socket and NUMA missing-capacity cases into TestPrepareResourceClaimsGroupedModeRejectsInvalidCPUCapacity. And I keep the other-driver case separate because it verifies successful filtering rather than capacity rejection. Thanks!

AutuSnow:
@alanhuangch   Thanks for the fix. This makes the node-side handling more robust.
﻿
I'm curious about the concrete scenario in which a matching grouped allocation could be missing the `dra.cpu/cpu` entry. In Kubernetes v0.36, when `DRAConsumableCapacity` is enabled, the scheduler populates `ConsumedCapacity` for devices with `AllowMultipleAllocations` by iterating over all declared device capacities. Even if the ResourceClaim does not explicitly request `dra.cpu/cpu`, the default value or the device's full capacity should still be recorded. If the feature is disabled and the claim explicitly requests capacity, scheduling should fail earlier.
﻿
Is this mainly intended as fail-closed validation for malformed or version-skewed allocation results, or is there a supported path that can produce such an allocation?

alanhuangch:
@AutuSnow Thanks for checking the scheduler behavior.

I agree that I have not identified a supported Kubernetes v0.36 scheduler path that would normally produce a matching grouped allocation without the dra.cpu/cpu entry.

The intent here is primarily fail-closed validation at the node boundary. The allocation result could be malformed because of version-skewed state or a  non-standard allocator/controller, but I don't have a concrete scheduler-generated reproducer for that scenario.

Previously, the node plugin interpreted the missing entry as a zero-CPU allocation and returned success without creating CDI state or recording an allocation. Rejecting an allocation result that violates the grouped-device invariant seemed safer and makes the failure explicit.

So I would characterize this as defensive hardening rather than fixing a currently known supported scheduler path. I’m happy to adjust the PR framing if you think it should be described differently.

ffromani:
+1 for the fix, which I intend as general robustness fix worth adding, but also +1 about the "when this can happen?" question.

In general I'm willing to be a little more wary wrt the external API validation, or when the scope is big enough. E.g., this level of defense in depth vs API objects is fine to me; OTOH checking and re-validating input data across a function boundary in a purely internal flow is probably too much.

AutuSnow:
I agree that this is a reasonable defense-in-depth check at the external API boundary. As I understand it, this should not occur in the normal Kubernetes v0.36 scheduling path; it would require something abnormal, such as a scheduler or controller bug, or version/feature-gate skew between components. Is that a fair characterization?

alanhuangch:
@AutuSnow Yes, that is a fair characterization. In the normal Kubernetes v0.36 scheduling path, I would not expect this state to be produced. 

The check protects the node plugin against an allocation result that violates the grouped-device invariant. A scheduler/controller bug or an incompatible or skewed component could theoretically produce such a result, although I have not reproduced a specific version/feature-gate combination that does so.

So normal operation should not hit this error; it is intended to fail explicitly if abnormal API state reaches the node.

AutuSnow:
Okay, that sounds good.Thanks!
/lgtm

PR #278, opened 2026-08-03
fix: match configs for selected subrequests

AutuSnow:
#### What type of PR is this?

/kind bug

Uses `resourceclaim.ConfigForResult` when selecting an opaque allocation config. This preserves the existing empty-list and exact-match behavior while also applying a config for a parent request such as `foo` to the selected prioritized subrequest `foo/bar`.

Without this, machine-mode `NodePrepareResources` rejects prioritized-list allocations with zero matching configurations.

Follow-up to https://github.com/kubernetes-sigs/dra-driver-cpu/pull/269#issuecomment-5138403095

#### Release note

```release-note
Fixed machine-mode preparation for prioritized-list requests whose opaque CPU configuration targets the parent request.
```

#### Documentation updates

NONE

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"slices"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


ffromani:
/lgtm

leaving to @pravk03 the final approval

pravk03:
Thanks for the fix @AutuSnow ! 

/approve

PR #282, opened 2026-08-03
fix(helm): refuse a kubelet root the kubelet reinterprets in only one place

### LINKED ISSUE #231: Helm chart hardcodes /var/lib/kubelet; driver fails to register on a relocated kubelet --root-dir

thc1006:
#### What type of PR is this?

/kind bug

`kubeletRootDir` becomes three things: the driver's `--kubelet-root-dir` argument and the two kubelet hostPath mounts. The kubelet may reinterpret `$(` and `$$` in a container's argument, rewriting `$$` to a single `$` and substituting `$(VAR)` where that variable is defined, while `hostPath.path` and `volumeMount.mountPath` are taken as written. A root spelled that way can reach the driver as one directory and the volumes as another, which is the shape of #231.

The chart now refuses either sequence when it survives cleaning. Cleaning first matters both ways: `/var/lib/$(IGNORED)/../kubelet` renders as `/var/lib/kubelet` and is accepted, so a path that works today does not become an upgrade that fails. A lone `$` is accepted too, since the kubelet leaves it alone and it names the same directory everywhere.

Coverage is in two places. `hack/ci/helm-render-check.sh` renders the chart for each root worth checking and hands the DaemonSet to a validator in `test/render`, which asserts that both named mounts, both hostPaths behind them and the managed argument derive from one directory. The `helm e2e` workflow gains a lane whose kind cluster relocates the kubelet root, so the chart is installed against a kubelet that is not where the driver's default points.

Fixes #231

**Compatibility.** A release whose cleaned `kubeletRootDir` contains `$(` or `$$` will now fail to render rather than install. `$$` is always rewritten in a container's arguments; `$(VAR)` is rewritten only where that variable exists, and an unresolved reference is left alone. The chart cannot know what a Pod's environment will hold, so it refuses both rather than guess, and the refusal is a hard failure and not a warning.

**Out of scope.** Exporting the driver name is still worth doing, and the repo needs it for its own sake, since `cmd/dracpu/app.go` keeps `driverName` unexported while `test/e2e/e2e_suite_test.go` declares its own copy. The reduced validator no longer asks for the name, so there is nothing here to hang it on; keeping that refactor separate avoids growing this one again.

A lane-specific path assertion is also left out. Registration succeeding under the configured root is the same claim, and it stays small to add later, since the matrix row already carries the root and a check step would need one `env:` line to see it.

**Job names change.** `helm-lint` becomes `helm-lint (v3.20.1)` and `helm-lint (v4.2.4)`; `helm-e2e` becomes `helm-e2e (default-root)` and `helm-e2e (relocated-root)`. Both workflows now name their helm rather than taking whatever they are given, so a helm release cannot change what CI proves on its own. `v3.20.1` is the version the Makefile installs when the machine has no helm, which nothing was covering before: that helm assumes Kubernetes v1.20 when no cluster is named, so it refuses this chart's `kubeVersion` range outright. The render check therefore names a version too, the oldest stable release the range covers.

**Follow-up.** #308 covers the neighboring hole this does not close: a top-level values key misspelled any way other than by case is still accepted and ignored, and the chart renders the default root. #309 implements it, and has since narrowed to three files. The two tips now overlap in the `Makefile`, the generated `values.schema.json` and `docs/user/configuration.md`, and merging them locally has no conflicts. Whichever lands second rebases and regenerates the schema.

**How the cover was checked.** Each piece was broken rather than read:

- pointing the registry `hostPath` at the plugins directory, or giving a case a wrong expected root, fails the validator
- renaming the test out of Go's discovery, or skipping it early, fails the package
- emitting `-kubelet-root-dir=` or `--kubelet-root-dir=` at the default root fails the default case, while a form the flag package rejects is left alone
- appending a second DaemonSet, a wrong `apiVersion`, unrelated keys, an explicit `null` or an explicit `{}` each fails the render check, and only comment or whitespace documents are skipped
- dropping the template's `clean` is caught by the `/` case
- widening the guard to refuse a lone `$`, and turning the socket budget's `gt` into `ge`, are each caught by their boundary case
- disabling either half of the guard, the miscased-key guard or the socket budget fails the matching refusal case
- a refusal case whose message never appears fails and prints what the chart actually said, so an unrelated helm failure cannot satisfy one

The chart's lint, render, docs and schema checks pass under both pinned helm versions. `make lint`, `make modernize`, `make test-unit`, `go build ./...` and `go vet ./...` pass, along with shellcheck and shfmt at the versions the pre-commit hooks pin.

#### Release note

```release-note
The Helm chart now refuses a `kubeletRootDir` whose cleaned value contains `$(` or `$$`. The kubelet may reinterpret those in a container's arguments but not in `hostPath.path` or `volumeMount.mountPath`, which could otherwise give the driver a different root from the one mounted into the container.
```

#### Documentation updates

- `docs/user/configuration.md` documents the restriction on `kubeletRootDir`.
- `docs/dev/testing.md` gains the render check.

### DIFF
--- a/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
+++ b/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
@@ -79,6 +79,9 @@ check below for why that specific check exists.
 {{- if not (isAbs $root) -}}
 {{- fail (printf "kubeletRootDir must be an absolute path, got %q" $root) -}}
 {{- end -}}
+{{- if or (contains "$(" $root) (contains "$$" $root) -}}
+{{- fail (printf "kubeletRootDir must not contain %q or %q, got %q: the kubelet expands those in a container's arguments and not in its mount paths, so the driver would be given a root the volumes were not mounted at" "$(" "$$" $root) -}}
+{{- end -}}
 {{- /* Cleaned (not just trimmed) so this resolves to the same directory the
        driver's filepath.Join produces. */ -}}
 {{- $cleaned := clean $root -}}
--- a/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
+++ b/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
@@ -80,7 +80,7 @@ check below for why that specific check exists.
 {{- fail (printf "kubeletRootDir must be an absolute path, got %q" $root) -}}
 {{- end -}}
 {{- if or (contains "$(" $root) (contains "$$" $root) -}}
-{{- fail (printf "kubeletRootDir must not contain %q or %q, got %q: the kubelet expands those in a container's arguments and not in its mount paths, so the driver would be given a root the volumes were not mounted at" "$(" "$$" $root) -}}
+{{- fail (printf "kubeletRootDir must not contain %q or %q, got %q: the kubelet expands those in a container's arguments and not in its mount paths, so the spelling is refused wherever it appears rather than only where it would still diverge after cleaning" "$(" "$$" $root) -}}
 {{- end -}}
 {{- /* Cleaned (not just trimmed) so this resolves to the same directory the
        driver's filepath.Join produces. */ -}}
--- a/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
+++ b/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
@@ -80,8 +80,12 @@ check below for why that specific check exists.
 {{- fail (printf "kubeletRootDir must be an absolute path, got %q" $root) -}}
 {{- end -}}
 {{- /* Cleaned (not just trimmed) so this resolves to the same directory the
-       driver's filepath.Join produces. */ -}}
+       driver's filepath.Join produces. Cleaned before the check below, because
+       that value is what reaches the argument and the two mount paths. */ -}}
 {{- $cleaned := clean $root -}}
+{{- if or (contains "$(" $cleaned) (contains "$$" $cleaned) -}}
+{{- fail (printf "kubeletRootDir must not contain %q or %q once cleaned; %q cleans to %q: the kubelet may reinterpret those in a container's arguments but not in hostPath.path or volumeMount.mountPath" "$(" "$$" $root $cleaned) -}}
+{{- end -}}
 {{- /* The registrar socket under this root has to fit sun_path: 108 bytes minus
        the terminating NUL leaves 107 for the path, and this suffix, fixed while
        rolling updates are off, leaves 73 for the root. Checked here in bytes,
--- a/test/kubeletroot/registration_test.go
+++ b/test/kubeletroot/registration_test.go
@@ -1,0 +1,243 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package kubeletroot checks, on a running kind cluster, that the driver put
+// its sockets under the kubelet root the chart was given. The kubelet watches
+// plugins_registry under its own root, so a driver that registers anywhere else
+// is never seen. Rendering the chart proves none of that, which is why this is
+// separate from the render checks in test/render.
+package kubeletroot
+
+import (
+	"context"
+	"errors"
+	"os"
+	"os/exec"
+	"path"
+	"slices"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/kubernetes-sigs/dra-driver-cpu/internal/driverconfig"
+	testclient "github.com/kubernetes-sigs/dra-driver-cpu/test/pkg/client"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8swait "k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// test/render pins this against cmd/dracpu and the chart, so a rename turns
+	// that test red rather than moving these paths.
+	driverName = "dra.cpu"
+
+	sliceTimeout = 2 * time.Minute
+
+	// Bounds one inspection, so a node that stops answering names the socket.
+	inspectTimeout = 30 * time.Second
+)
+
+func envOr(name, fallback string) string {
+	if v := os.Getenv(name); v != "" {
+		return v
+	}
+	return fallback
+}
+
+// requireVariant holds the run to the job it was launched as, since the root
+// alone cannot say which of the two the workflow meant. Absent is refused rather
+// than defaulted, because defaulting is how it would go missing.
+func requireVariant(t *testing.T, root string) {
+	t.Helper()
+	relocated := root != driverconfig.DefaultKubeletRootDir
+	switch variant := os.Getenv("DRACPU_E2E_VARIANT"); variant {
+	case "default-root":
+		if relocated {
+			t.Fatalf("the default-root variant is running against %q", root)
+		}
+	case "relocated-root":
+		if !relocated {
+			t.Fatalf("the relocated-root variant is running against the default root %q", root)
+		}
+	default:
+		t.Fatalf("DRACPU_E2E_VARIANT is %q, want default-root or relocated-root", variant)
+	}
+}
+
+func TestDriverRegistersUnderTheConfiguredRoot(t *testing.T) {
+	// Cleaned first, since the paths below are joined onto it and joining cleans:
+	// left raw, a non-canonical spelling of the default root would ask for one
+	// socket to be present and then to be missing.
+	root := path.Clean(envOr("DRACPU_E2E_KUBELET_ROOT_DIR", driverconfig.DefaultKubeletRootDir))
+	requireVariant(t, root)
+	namespace := envOr("DRACPU_E2E_NAMESPACE", "kube-system")
+	name := envOr("DRACPU_E2E_DAEMONSET", "dracpu")
+
+	if _, err := exec.LookPath("docker"); err != nil {
+		t.Fatalf("this test reads sockets inside the kind nodes and needs docker: %v", err)
+	}
+	cs, err := testclient.NewK8SClientset()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx := t.Context()
+
+	// From the cluster, not from the driver's own pods: a set derived from the
+	// pods would agree with itself if the DaemonSet stopped covering a node.
+	nodes := clusterNodes(t, ctx, cs)
+	t.Logf("checking %d node(s) against kubelet root %q", len(nodes), root)
+
+	if got := driverNodes(t, ctx, cs, namespace, name); !slices.Equal(got, nodes) {
+		t.Fatalf("driver pods run on %v, the cluster has %v", got, nodes)
+	}
+	waitForSliceCoverage(t, ctx, cs, nodes)
+
+	for _, node := range nodes {
+		t.Run(node, func(t *testing.T) {
+			// Asserted present, not merely "not under the default": without this a
+			// mistake in the paths built above would leave both sides looking clean.
+			if got := statOf(t, node, registrarSocket(root)); got != "socket" {
+				t.Errorf("registrar socket under %s is %s, want socket", root, got)
+			}
+			if got := statOf(t, node, draSocket(root)); got != "socket" {
+				t.Errorf("DRA socket under %s is %s, want socket", root, got)
+			}
+			if root == driverconfig.DefaultKubeletRootDir {
+				return
+			}
+			// Another kubelet plugin may legitimately keep a registrar socket under
+			// the default root, so these are named rather than globbed.
+			if got := statOf(t, node, registrarSocket(driverconfig.DefaultKubeletRootDir)); got != "missing" {
+				t.Errorf("registrar socket under the default root is %s, want missing", got)
+			}
+			if got := statOf(t, node, draSocket(driverconfig.DefaultKubeletRootDir)); got != "missing" {
+				t.Errorf("DRA socket under the default root is %s, want missing", got)
+			}
+		})
+	}
+}
+
+func registrarSocket(root string) string {
+	return path.Join(root, "plugins_registry", driverName+"-reg.sock")
+}
+
+// draSocket is what kubeletplugin binds while rolling updates are off, which is
+// how this driver runs.
+func draSocket(root string) string {
+	return path.Join(root, "plugins", driverName, "dra.sock")
+}
+
+func clusterNodes(t *testing.T, ctx context.Context, cs kubernetes.Interface) []string {
+	t.Helper()
+	list, err := cs.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	var names []string
+	for _, n := range list.Items {
+		names = append(names, n.Name)
+	}
+	if len(names) == 0 {
+		t.Fatal("the cluster reported no nodes, so nothing below would have been checked")
+	}
+	slices.Sort(names)
+	return names
+}
+
+// driverNodes returns the nodes the driver's pods run on, through the
+// DaemonSet's own selector so it cannot drift from how the release was installed.
+func driverNodes(t *testing.T, ctx context.Context, cs kubernetes.Interface, namespace, name string) []string {
+	t.Helper()
+	ds, err := cs.AppsV1().DaemonSets(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	selector, err := metav1.LabelSelectorAsSelector(ds.Spec.Selector)
+	if err != nil {
+		t.Fatal(err)
+	}
+	pods, err := cs.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: selector.String()})
+	if err != nil {
+		t.Fatal(err)
+	}
+	var names []string
+	for _, p := range pods.Items {
+		if !slices.Contains(names, p.Spec.NodeName) {
+			names = append(names, p.Spec.NodeName)
+		}
+	}
+	slices.Sort(names)
+	return names
+}
+
+// A rolled out DaemonSet does not mean every driver has published, and reading
+// the list once would pass on the first slice from the first node.
+func waitForSliceCoverage(t *testing.T, ctx context.Context, cs kubernetes.Interface, nodes []string) {
+	t.Helper()
+	var last []string
+	err := k8swait.PollUntilContextTimeout(ctx, 2*time.Second, sliceTimeout, true, func(ctx context.Context) (bool, error) {
+		list, err := cs.ResourceV1().ResourceSlices().List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return false, err
+		}
+		last = nil
+		for _, s := range list.Items {
+			// A slice without a node name is published for the whole cluster, and
+			// says nothing about which kubelet accepted the driver.
+			if s.Spec.Driver != driverName || s.Spec.NodeName == nil {
+				continue
+			}
+			if !slices.Contains(last, *s.Spec.NodeName) {
+				last = append(last, *s.Spec.NodeName)
+			}
+		}
+		slices.Sort(last)
+		return slices.Equal(last, nodes), nil
+	})
+	if err != nil {
+		t.Fatalf("%s ResourceSlices cover %v, the cluster has %v: %v", driverName, last, nodes, err)
+	}
+}
+
+// An inspection that could not run is a failure of this test, never the same
+// answer as the socket not being there.
+const statScript = `
+set -eu
+p=$1
+if [ -S "$p" ]; then echo socket
+elif [ -e "$p" ]; then echo other
+else echo missing
+fi
+`
+
+func statOf(t *testing.T, node, socketPath string) string {
+	t.Helper()
+	return onNode(t, node, statScript, socketPath)
+}
+
+func onNode(t *testing.T, node, script, arg string) string {
+	t.Helper()
+	ctx, cancel := context.WithTimeout(t.Context(), inspectTimeout)
+	defer cancel()
+	out, err := exec.CommandContext(ctx, "docker", "exec", node, "sh", "-c", script, "sh", arg).Output()
+	if err != nil {
+		if exit, ok := errors.AsType[*exec.ExitError](err); ok {
+			t.Fatalf("reading %s on %s: %v: %s", arg, node, err, strings.TrimSpace(string(exit.Stderr)))
+		}
+		t.Fatalf("reading %s on %s: %v", arg, node, err)
+	}
+	return strings.TrimSpace(string(out))
+}
--- a/test/render/kubelet_root_test.go
+++ b/test/render/kubelet_root_test.go
@@ -488,6 +488,10 @@ func TestUnusableRootIsRefused(t *testing.T) {
 		// mount paths, so a root carrying one is mounted at one place and given to
 		// the driver as another.
 		{"a variable reference", "/mnt/$(KUBERNETES_SERVICE_HOST)/kubelet", "expands those in a container's arguments"},
+		// Cleaning would take this one back to /mnt/kubelet, where nothing diverges.
+		// Refused anyway, because the rule is on the spelling: which sequences
+		// survive cleaning is not something a reader of the value can tell.
+		{"a variable reference cleaning would remove", "/mnt/$(discarded)/../kubelet", "refused wherever it appears"},
 		{"an escaped dollar", "/mnt/$$kubelet", "expands those in a container's arguments"},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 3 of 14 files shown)


thc1006:
/hold

Keeping this on hold until #274 merges; this branch is still stacked on it.

Consolidating the current state here:

- `helm-e2e` now runs both the existing default-root case and a kind cluster whose kubelets use `/var/lib/custom-kubelet`.
- The relocated-root check fails closed: it uses `set -euo pipefail`, requires a non-empty node set, checks that driver Pods cover every node once the DaemonSet has rolled out, waits for `dra.cpu` ResourceSlices to cover every node, verifies real Unix sockets under the relocated root, and confirms that no `dra.cpu` sockets remain under `/var/lib/kubelet`.
- The workflow triggers now include the relevant image, Go, driver, chart, kind-config, and test-helper inputs.

A ResourceClaim/Pod workload is deliberately out of scope here. #231 is the registration/path mismatch, which this test covers; exercising `NodePrepareResources` would be a broader follow-up.

After #274 merges, I'll rebase so this PR contains only the workflow change and the relocated kind config.

pravk03:
Sorry for the delay. This is on my list, will take a look by tomorrow.

thc1006:
> Sorry for the delay. This is on my list, will take a look by tomorrow.

It's OK. Review is really a hard work more than creating a PR.
Take ur time. Really very thank you!!!

thc1006:
/unhold

#274 has merged, which was the only reason this was held.

Rebased onto current main while I was here, so the presubmits run against what is there now rather than a base from a week ago. The rebase was clean: this branch only touches `helm-e2e.yaml`, `helm-lint.yaml`, `Makefile`, `hack/ci/helm-render-check.sh` and `hack/ci/kind-ci-relocated-root.yaml`, and none of those overlap with what #274, #289 or #291 changed, so nothing of theirs is disturbed.

@pravk03 no rush at all, just flagging that the dependency is gone and this is now reviewable on its own.

This comment was written in part with the assistance of generative AI. Bcuz my English might be sucks.

ffromani (commented):
thanks for this PR. Adding tests is welcome and needed *especially* for a rarely-used (or expected rarely used) feature. The concerns I have are about how we implement the tests.
Can we perhaps avoid large inline shell scripts in the manifests? having the test/setup code packaged as its own indepedent script, invoked by the manifests, helps testability, maintainership and versioning
Can we have helpers, perhaps partially, written in higher level languages, e.g. golang so the testing logic is more explicit and less muddled by the quirks of shell scripting?

thc1006:
/hold

Holding this behind #294 rather than leaving the ordering to a note in the description.

The two disagree about one thing: the case here asserts that `extraArgs` accepts `--kubelet-root-directory` as a near-match of the managed flag, and #294 removes `extraArgs` from the chart entirely, so the schema refuses the whole key. Either order is fine on its own, but nothing serialises them. `make helm-render-check` runs in the Helm Lint workflow, and a GitHub Actions `pull_request` run is against each pull request's own merge ref; no Prow job runs it, so the required jobs on a combined ref would not exercise that test at all. Two green pull requests could merge into a red main.

#294 already has an lgtm, so it is the cheaper one to let through. Once it lands I will rebase and rewrite that block against the chart API it leaves behind, then unhold.

thc1006:
/hold cancel

#294 has landed, which was the only reason this was held. Rebased on it and reworked the block that depended on it.

The rebase was clean, which is the part worth flagging. The render suite still asserted the deny guard #294 removed from `_helpers.tpl`, so four of its cases then failed against main, reporting `accepted=true` for spellings the chart no longer refuses. Nothing conflicted; the meaning went out from under the test.

That guard is obsolete rather than broken. With no `extraArgs` in the chart there is nothing to splice in after the managed flag, so the divergence it protected against cannot be written any more. The block now checks what replaced it: nothing written under `extraArgs` reaches the container, and the root the driver starts on still comes from `kubeletRootDir`.

One thing went the other way from what I assumed. The key is ignored rather than refused, since the top-level values schema sets no `additionalProperties`, so the check reads the rendered arguments instead of the render's exit status. Re-adding the DaemonSet append fails all five cases and names the two `kubelet-root-dir` flags the driver would have been handed.

Ran it under helm 3.20.1 and 4.2.3 both, because `HELM` in the Makefile prefers whatever `helm` is on PATH and only falls back to the pinned build when there isn't one, so the pin is softer than it looks.

---

*Written in part with the assistance of generative AI. I ran the tests and the mutation myself.*

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package render checks the DaemonSet the chart produces, for the things lint
+// and schema validation cannot see: the kubelet root has to produce both mount
+// paths and the driver flag from one value, or the driver registers on a path
+// the kubelet never watches. No cluster needed.
```
ffromani: so it's a semantic evaluation essentially, right?

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+		if unquoted, err := strconv.Unquote(lit.Value); err == nil {
+			name = unquoted
+		}
+		return false
+	})
+	if name == "" {
+		t.Fatalf("no driverName constant found in %s", appSourcePath)
+	}
+	return name
+}
```
ffromani: I'm fine adding extra testing, and I'm still convinced that golang (or python) test helpers are a better and healthier approach than a large shell script blob; but this doesn't mean we can't use a mixed approach to simplify the testing code. In this case the best approach is to move the constant to an importable package - `pkg/driver` seems like a fine place, and we can rename it so it reads `driver.Name`

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+	"github.com/kubernetes-sigs/dra-driver-cpu/internal/driverconfig"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	chartPath     = "../../deployment/helm/dra-driver-cpu"
```
ffromani: the effort spent to make the test self sufficient is admirable, but we can also solve this problem with better orchestration: the chart path can and should be a test parameter (passed however it is simple and reasonnable, from flags to env vars) the calling orchestration (github action, Makefile... perhaps Makefile) can pass

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+
+// What the chart renders when no root is given, spelled out rather than joined
+// onto the driver's constant by the same code being checked.
+const (
+	defaultPlugins  = "/var/lib/kubelet/plugins"
+	defaultRegistry = "/var/lib/kubelet/plugins_registry"
+)
+
+func helmBin(t *testing.T) string {
+	t.Helper()
```
ffromani: I'd prefer all the helpers to drop the `t *testing.T` parameter and just return `err`, bubbling up as needed, and letting the top-level caller handle the error, for example with `t.Fatalf`.

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+	base := []string{"template", "t", chart, "--kube-version", kubeVersion}
+	cmd := exec.CommandContext(ctx, helmBin(t), append(base, args...)...) //nolint:gosec // the helm binary is whichever one the caller put on PATH or in HELM
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("%w: %s", err, strings.TrimSpace(stderr.String()))
+	}
+	return stdout.Bytes(), nil
+}
```
ffromani: likewise: the rendering logic is pretty simple and straightforward so it can be kept as shell helper or done in the Makefile. In general, we don't _need_ to do it programmatically in the test code, and actually doing it in the test orchestration code should lead to less and simpler code overall. Likewise, if we can use an existing tool to split the rendered YAML and just focus on the daemonset, it's even better.

The intent of my earlier suggestion was not to avoid the shell entirely, but to move the complex testing logic into a readable and maintainable artifact, which is _likely_ golang (or python, probably better golang).

I think the best and simplest path forward involves a mix of shell/commands and custom testing logic in golang.

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+	for i := range ds.Spec.Template.Spec.Containers {
+		if c := &ds.Spec.Template.Spec.Containers[i]; c.Name == driverContainer {
+			found = append(found, c)
+		}
+	}
+	if len(found) != 1 {
+		t.Fatalf("%d containers named %q, want one", len(found), driverContainer)
+	}
+	return found[0]
+}
```
ffromani: OTOH once we get the daemonset YAML finding the right container belongs in the golang testing code. IOW, we probably want to keep this

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+func commandLine(c *corev1.Container) ([]string, error) {
+	argv := append(slices.Clone(c.Command), c.Args...)
+	if len(argv) == 0 {
+		return nil, errors.New("the container carries no command line")
+	}
+	if strings.HasPrefix(argv[0], "-") {
+		return nil, fmt.Errorf("the command line starts with %q, so there is no program to run", argv[0])
+	}
+	return argv, nil
+}
```
ffromani: this seems correct, but is it relevant for the relocated root testing we care about?

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+		return ""
+	}
+	return c.Value.String()
+}
+
+// renderedRoot answers what the driver would run on, or why it would not get
+// that far. It registers what cmd/dracpu registers and parses from the same
+// offset, so the answer comes from the parser the render will meet rather than
+// a second implementation of Go's flag grammar.
+func renderedRoot(c *corev1.Container) (root string, sent bool, err error) {
```
ffromani: so we are reimplementing production logic in the test, which leads to more bugs and coupling down the road.
Can we avoid this part at the cost of a less robust test? the best shape of the solution is IMO add an introspect subcommand (or, if makes sense, augment an existing one) to report the data we need so we can test the built artifact vs reimplementing it.

This is another example of what I tried (and failed, sorry about it) to convey: the testing logic proper should be in a high enough language so it's maintainable, but the testing logic can and should be fed by the test runner/orchestrator. The test runner can for example run `dracpu inspect <whatever>` and feed the test runner

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+	if !strings.Contains(string(helpers), socket) {
+		t.Errorf("the chart does not spell the registrar socket %q, which driverName %q gives", socket, name)
+	}
+}
+
+// Four places have to agree on the default root: the driver's constant, the
+// chart value, the template's test for whether to send the flag, and the
+// helper's answer when the value is absent. No flag is sent at the default, so
+// a one-sided change leaves nothing in the render to say the two disagree.
+func TestDefaultRootAgreesEverywhere(t *testing.T) {
```
ffromani: I'm not sure how much value this test add. Meaning: why should these agree? as long as each layer is consistent, they can drift. The builtin code can have a default which needs to be safe and correct; the helm manifest can legitimately use a different, albeit safe and correct, override path.

test/render/kubelet_root_test.go
```diff
@@ -0,0 +1,841 @@
...
+			if diff := cmp.Diff(tc.want, b.notWhole); diff != "" {
+				t.Errorf("mounts reported as not carrying the whole volume (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+// Every shape below is one line in a template, and they part company only at
+// the driver.
+func TestRenderedArgumentsAreReadAsTheDriverReadsThem(t *testing.T) {
```
ffromani: it seems to me this is basically retesting the behavior of the flag processing. I think we should start with a simpler test, perhaps postponed till we have config introspection

test/kubeletroot/registration_test.go
```diff
@@ -0,0 +1,243 @@
...
+	case "relocated-root":
+		if !relocated {
+			t.Fatalf("the relocated-root variant is running against the default root %q", root)
+		}
+	default:
+		t.Fatalf("DRACPU_E2E_VARIANT is %q, want default-root or relocated-root", variant)
+	}
+}
+
+func TestDriverRegistersUnderTheConfiguredRoot(t *testing.T) {
```
ffromani: the intent of this test is fine. We want to actually start a kubelet with a relocated root and a matching driver and check everything works.
But can't we make it simpler?
1. create a lane with a relocated root
2. run the driver
3. run the existing (or a selected meaningful subset) of the existing tests - if they pass, the driver is communicating fine with the kubelet, so the relocated root is validated
4. optionally, add a lane-specific check which verifies all the path look as we expect them to look; we can use ginkgo filters and labels to make sure this test runs only on the relocated root lane

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
```diff
@@ -79,6 +79,9 @@ check below for why that specific check exists.
 {{- if not (isAbs $root) -}}
 {{- fail (printf "kubeletRootDir must be an absolute path, got %q" $root) -}}
 {{- end -}}
+{{- if or (contains "$(" $root) (contains "$$" $root) -}}
+{{- fail (printf "kubeletRootDir must not contain %q or %q, got %q: the kubelet expands those in a container's arguments and not in its mount paths, so the driver would be given a root the volumes were not mounted at" "$(" "$$" $root) -}}
```
ffromani: I think this test is worth adding but it will need rearchitected once the comments against the prior commits are addressed

test/render/kubelet_root_test.go
```diff
@@ -880,3 +884,46 @@ func TestTheRelocatedKindConfigDiffersOnlyByTheKubeletRoot(t *testing.T) {
 		t.Errorf("the two kind configs differ by more than the kubelet root (-default,+relocated):\n%s", diff)
 	}
 }
+
+// The cluster verifier can only judge a row that runs, so deleting the relocated
+// entry leaves every remaining check green and takes the only proof of issue 231
+// with it. Read the workflow rather than the rows it happened to produce.
+func TestBothKubeletRootsStayInTheMatrix(t *testing.T) {
```
ffromani: this is a smell: a test need to be aware of how is run, the dependency chain is backwards. The test runner must pass the test parameters as needed, and the test should adapt.
Furthermore as this test stands it seems to add little value so I'd probably drop it and re-evaluate later

ffromani (commented):
I want to first and foremost acknowledge the effort and the work spent in this iteration, which are truly commendable. Impressive initiative.
The problem, however, is that it seems the pendulum swong too far now and there is too much work done at golang layer and reimplemented in golang. This is a different but equally problematic place maintainability-wise.

The fix is downscope quite aggressively: let's keep the *validation* logic in golang, because this is custom, semantic and better expressed into higher level language. but the test runner and orchestration can and should be done using shell script with (I believe) simple and focused well known commands.

Likewise, if some targeted refactoring can make the test more robust and simpler (vs reading back and parsing CI-specific config or source code!) then we should do it as predepenendency.

Some tests can and should be postponed once that refactoring happens; let's focus on the key max-ROI validation for this PR, so we can wrap up this work.

I'd recommend to get in touch, perhaps with github/review comments or perhaps on slack, to fully align about expectations before to embark in such massive (and again, impressive and commendable) rework. I actually meant for a much smaller change than what was implemented

ffromani:
possible introspect config support: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/262

thc1006:
You were right, and thanks for saying it plainly. After the first round I moved the orchestration into Go as well as the validation, and the result was a second model of the repository.

Cut back and rebased on main. 1353 lines to 371.

`hack/ci/helm-render-check.sh` now owns helm, the case list, the chart path and the expected values. `--show-only templates/daemonset.yaml` is the existing tool that splits the stream. Go keeps only what shell reads badly: the two named mounts and the volumes behind them have to be the same directory, and the managed argument has to be that directory or absent at the default. Helpers return errors; the test boundary decides.

Gone: the flag registration and argv parsing, the AST read of `cmd/dracpu`, the kind-config diff, the workflow matrix read, the four-layer default comparison, and the cases that tested the validator against a hand-edited oracle.

Also gone is `test/kubeletroot`. I went looking for what the relocated lane already proves rather than assuming: readiness reads `/healthz`, which stays 503 until `ready.Store(true)` in `cmd/dracpu/app.go`, which only runs after `CPUDriver.Start` returns, and `Start` blocks on `waitForRegistration(ctx, d, registrarDir(cp.kubeletRootDir), ...)` in `pkg/driver/driver.go`. A root the kubelet is not watching never becomes Ready, so `rollout status` on that lane is the regression test for #231. The docker exec, the node-name assumption and the socket walk were all buying something already bought.

Two of your suggestions I did not take, and I would rather say why than quietly skip them.

`driver.Name`: agreed, and the repo needs it for its own sake, since `cmd/dracpu` keeps the constant unexported while `test/e2e/e2e_suite_test.go` carries its own copy and a `"dra.cpu/cpu"` literal beside it. But the reduced validator no longer asks for the name at all, so there is nothing here to hang it on. You said as predependency, so I would rather send it separately than grow this one back.

The lane-specific path check: I left it out for now. Registration succeeding under the configured root is the same claim, and the suite has no labels yet, so this would be the first. If you want it anyway it is small, since the workflow already passes `DRACPU_E2E_KUBELET_ROOT_DIR` per matrix row and only the consumer went away.

One thing I found while rewriting the docs: the wording said the kubelet expands `$(` and `$$` in arguments, which is not quite it. `$$` becomes one `$`, and `$(VAR)` is substituted only where that variable is defined. Reworded to say the kubelet may reinterpret them.

`make lint`, `make modernize`, `make helm-lint`, `make helm-render-check`, `make helm-docs-check`, `make helm-schema-check`, `make test-unit`, `go build ./...` and `go vet ./...` all pass. I also checked the render validator catches a broken chart rather than passing vacuously: pointing the registry hostPath at the plugins directory fails it, and disabling the template guard fails the refusal cases.

thc1006:
Two additions since that comment, both from re-reading which tests you had actually asked to drop.

The guard-refusal cases were not among them, so `hack/ci/helm-render-check.sh` covers the miscased key and the sun_path budget again alongside the two this change adds, and each now asserts the message so the values schema cannot pass one by refusing it first.

Green on `ee698c0`.

thc1006:
One more case since that comment: a `/` root, which is the reason the chart cleans whole paths rather than appending. Under a mutation that drops the `clean`, the default, relocated and trailing-slash cases all still pass and only that one catches the `//plugins` it produces.

The description was still describing the shape before the cut, so it is rewritten too.

thc1006:
Rebuilt the history while I was in here. It was six commits, three of which built the things the last one deleted, and this repo keeps individual commits through the merge. Two now: the guard with its docs, then the coverage.

Two more from going back over it:

A `/` root is the case the chart cleans whole paths for, and it was missing. Under a mutation that drops the `clean`, the default, relocated and trailing-slash cases all still pass and only that one catches the `//plugins` it produces.

A trailing space is the only one of the six guards the values schema does not also refuse. I probed each input to see which layer answered rather than reasoning about it, and empty, relative, leading-space and blank all stop at the schema. So that one had no net and no case; it has both now.

Every guard has a case, and each was checked by disabling the guard rather than by reading it.

Also dropped a `.dockerignore` path filter I had added with a comment explaining its effect on the image. This repo has no `.dockerignore`.

thc1006:
The runner was failing open and I had not noticed. `go test -run` exits 0 when its pattern matches nothing, so a renamed, deleted or split test, or a typo in the pattern, would have left all four positive cases green with no assertion run. Checked against this package rather than in the abstract:

```
$ go test -count=1 ./test/render/ -run '^NoSuchTestName$' -args -manifest /dev/null
ok  github.com/kubernetes-sigs/dra-driver-cpu/test/render  [no tests to run]
$ echo $?
0
```

`TestMain` now fails the package when a manifest was passed and nothing read it, and the script no longer needs `-run` at all. Renaming the test out of Go's discovery turns `make helm-render-check` red with `no test read /tmp/.../default.yaml: the validator did not run`, where before it stayed green. A bare `go test ./test/render/` still passes, since the sentinel only applies when a manifest was given.

Worth having in a change whose whole point is coverage that cannot degrade quietly.

test/render/kubelet_root_test.go
thc1006: Yes, and that is now all the Go side does: it is handed a rendered DaemonSet and judges it. Rendering and the case list live in the shell.

test/render/kubelet_root_test.go
thc1006: Dropped rather than moved. The socket-name assertion that needed the constant is gone, so nothing parses it out of the source any more. The importable-constant change is better made with the driver-name refactor than here.

test/render/kubelet_root_test.go
thc1006: Done. The validator takes `-manifest`, `-expected-root` and `-expect-root-flag`, and `hack/ci/helm-render-check.sh` supplies them per case.

test/render/kubelet_root_test.go
thc1006: Done. Every helper returns `error`; the one test function is the only thing holding a `*testing.T`.

test/render/kubelet_root_test.go
thc1006: Done. `helm template --show-only templates/daemonset.yaml` does the splitting, and Go only decodes what it was handed. The file went from 936 lines to 209.

test/render/kubelet_root_test.go
thc1006: Kept, in `driver()`.

test/render/kubelet_root_test.go
thc1006: It is not. Dropped.

test/render/kubelet_root_test.go
thc1006: Dropped. Nothing re-implements the flag parser now: the check is that the flag is present exactly when the root is not the default, and that it carries that root.

test/render/kubelet_root_test.go
thc1006: Agreed, they can drift. Dropped.

test/render/kubelet_root_test.go
thc1006: Dropped, and left for config introspection as you suggest.

test/kubeletroot/registration_test.go
thc1006: Done as 1 to 3. The file is gone, and the relocated lane runs the existing e2e suite against a kind cluster with a moved root. Readiness is already the check you describe: `/healthz` stays 503 until `Start` returns, and `Start` blocks on registering under the configured root, so `rollout status` fails on a root the kubelet is not watching. 4 is not in this revision, since you marked it optional.

deployment/helm/dra-driver-cpu/templates/_helpers.tpl
thc1006: Reworked. The guard reads the cleaned value now, not the raw one, so a sequence in a segment the cleaning removes no longer fails a render that would have been correct.

test/render/kubelet_root_test.go
thc1006: Dropped, and the parameters come from the runner now.

thc1006:
@ffromani reworked along the lines you asked for, and every thread above is answered on this head.

`hack/ci/helm-render-check.sh` owns helm, the case list, the chart path and the expected values, and `--show-only templates/daemonset.yaml` is the existing tool that splits the stream. Go keeps only what shell reads badly: the two named mounts and the volumes behind them have to be one directory, and the managed argument has to be that directory or absent at the default. Helpers return errors. The Go side is 245 lines, against 841 for the file it replaced.

Gone: the flag registration and argv parsing, the AST read of `cmd/dracpu`, the kind-config and workflow reads, the four-layer default comparison, and `test/kubeletroot`. On that last one I went looking for what the relocated lane already proves rather than assuming. `/healthz` stays 503 until `ready.Store(true)` in `cmd/dracpu/app.go`, which runs only after `CPUDriver.Start` returns, and `Start` blocks on `waitForRegistration(ctx, d, registrarDir(cp.kubeletRootDir), ...)`. A root the kubelet is not watching publishes no slices and never becomes Ready, so that lane is the regression test for #231.

Two of your suggestions I did not take, and I would rather say why than skip them quietly.

`driver.Name`: agreed, and the repo needs it for its own sake, since `cmd/dracpu` keeps the constant unexported while `test/e2e/e2e_suite_test.go` carries its own copy. But the reduced validator no longer asks for the name at all, so there is nothing here to hang it on. You said as a predependency, so I would rather send it separately than grow this one back.

The lane-specific path check: left out for now. Registration succeeding under the configured root is the same claim, and the suite has no labels yet. It stays small to add if you want it anyway, since the matrix row already carries the root and a check step would need one `env:` line to see it.

The render names a Kubernetes version rather than letting helm assume one. The helm the Makefile installs when none is present assumes a version below the chart's `kubeVersion` range and refuses the chart, so the target failed before reaching a case. The one named is the oldest stable release that range covers. Both workflows name their helm for the same reason: they took whatever they were given, which was never the one the Makefile installs, and the lint job now runs against both.

`make lint`, `modernize`, `helm-lint`, `helm-render-check`, `helm-docs-check`, `helm-schema-check`, `test-unit`, `go build ./...` and `go vet ./...` pass, under that helm as well as the one on my PATH. Each guard was checked by disabling it, and each case by breaking what it reads, rather than by reading them.

Rebased onto current main after #310, so the relocated lane now runs against Kubernetes v1.37.0 and the Go checks build against the final `k8s.io v0.37.0` modules. Prow reports `BaseSHA:8db2cd8` on the new head.

Green on f0f1e17: both helm-lint versions, both root lanes, Go, pre-commit, E2E Local, modernize, and the six Prow jobs.

@ffromani could you take another look?

ffromani (commented):
/approve
/lgtm

thanks for the updates! I think we may need a followup to add more e2e tests, but this will likely depend on #262 which needs to wait a bit more.
I think this is a good start and a worthy addition.

PR #286, opened 2026-08-06
feat: add per-allocated-CPU metadata entries for grouped modes

### LINKED ISSUE #285: Add per-allocated-CPU metadata entries for grouped modes
Follow-up from #264.

In grouped modes, after `TakeByTopologyNUMAPacked` selects specific CPUs, publish per-allocated-CPU device entries in the KEP-5304 metadata alongside the group device entry. Each allocated CPU gets its own metadata entry with full topology attributes (`cpuID`, `numaNodeID`, `socketID`, `coreID`, `cacheL3ID`, `coreType`, `smtEnabled`).

This enables topology-aware consumers to map allocated CPUs to NUMA nodes without reading host sysfs. The group device metadata (#265) provides the NUMA/socket-level view; per-CPU entries provide the individual CPU topology.

Individual mode is unchanged since each allocation result already corresponds to a single CPU with full attributes.

Also extracts a shared `CPUAttributes()` function to eliminate attribute construction duplication between `builder.go` (ResourceSlice) and `dra_hooks.go` (metadata).

/kind feature

johnahull:
## Summary

- In grouped modes, add per-allocated-CPU device entries to KEP-5304
  metadata alongside the group device entry
- Each allocated CPU gets its own metadata entry with full topology
  attributes (cpuID, numaNodeID, socketID, coreID, cacheL3ID, coreType,
  smtEnabled)
- Extract shared `CPUAttributes()` function to eliminate duplication
  between `builder.go` and `dra_hooks.go`
- Individual mode unchanged

Fixes #285

## Example

Grouped-by-NUMA claim requesting 4 CPUs on a 4-NUMA Dell XE9680 (2x Intel Xeon Gold 6448Y, 128 CPUs). The metadata contains 1 group device + 4 per-CPU entries:

```json
{
  "requests": [
    {
      "name": "req-cpu",
      "devices": [
        {
          "driver": "dra.cpu",
          "name": "cpudevnuma000",
          "attributes": {
            "dra.cpu/allocatedNumCPUs": {"int": 4},
            "dra.cpu/numCPUs": {"int": 32},
            "dra.cpu/numaNodeID": {"int": 0},
            "dra.cpu/socketID": {"int": 0},
            "dra.cpu/smtEnabled": {"bool": true}
          }
        },
        {
          "driver": "dra.cpu",
          "name": "cpudev000",
          "attributes": {
            "dra.cpu/cpuID": {"int": 0},
            "dra.cpu/coreID": {"int": 0},
            "dra.cpu/socketID": {"int": 0},
            "dra.cpu/numaNodeID": {"int": 0},
            "dra.cpu/cacheL3ID": {"int": 0},
            "dra.cpu/coreType": {"string": "standard"},
            "dra.cpu/smtEnabled": {"bool": true}
          }
        },
        {
          "driver": "dra.cpu",
          "name": "cpudev024",
          "attributes": {
            "dra.cpu/cpuID": {"int": 24},
            "dra.cpu/coreID": {"int": 1},
            "dra.cpu/numaNodeID": {"int": 0},
            "..."
          }
        },
        {
          "driver": "dra.cpu",
          "name": "cpudev064",
          "attributes": {
            "dra.cpu/cpuID": {"int": 64},
            "dra.cpu/coreID": {"int": 0},
            "dra.cpu/numaNodeID": {"int": 0},
            "..."
          }
        },
        {
          "driver": "dra.cpu",
          "name": "cpudev088",
          "attributes": {
            "dra.cpu/cpuID": {"int": 88},
            "dra.cpu/coreID": {"int": 1},
            "dra.cpu/numaNodeID": {"int": 0},
            "..."
          }
        }
      ]
    }
  ]
}
```

CPUs 0 and 64 are SMT siblings (coreID 0). CPUs 24 and 88 are SMT siblings (coreID 1). All on NUMA node 0, socket 0, L3 cache 0.

## Test plan

- [x] Unit tests for per-CPU entries in all grouped modes
- [x] E2e test verifies per-CPU device count and attributes
- [x] `golangci-lint` passes with zero issues
- [x] Verified on Dell XE9680 (4 NUMA nodes, grouped-by-NUMA and
      grouped-by-socket modes)

### DIFF
--- a/pkg/device/attributes.go
+++ b/pkg/device/attributes.go
@@ -17,6 +17,7 @@ limitations under the License.
 package device
 
 import (
+	"github.com/kubernetes-sigs/dra-driver-cpu/pkg/cpuinfo"
 	resourceapi "k8s.io/api/resource/v1"
 )
 
@@ -50,6 +51,21 @@ const (
 	// claim from a grouped device's capacity.
 	AttributeAllocatedNumCPUs resourceapi.QualifiedName = "dra.cpu/allocatedNumCPUs"
 )
+
+// CPUAttributes builds the per-CPU device attributes from a CPUInfo and SMT state.
+func CPUAttributes(cpu cpuinfo.CPUInfo, smtEnabled bool) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute {
+	attrs := map[resourceapi.QualifiedName]resourceapi.DeviceAttribute{
+		AttributeNUMANodeID: {IntValue: new(int64(cpu.NUMANodeID))},
+		AttributeSocketID:   {IntValue: new(int64(cpu.SocketID))},
+		AttributeSMTEnabled: {BoolValue: new(smtEnabled)},
+		AttributeCacheL3ID:  {IntValue: new(int64(cpu.UncoreCacheID))},
+		AttributeCoreType:   {StringValue: new(cpu.CoreType.String())},
+		AttributeCoreID:     {IntValue: new(int64(cpu.CoreID))},
+		AttributeCPUID:      {IntValue: new(int64(cpu.CpuID))},
+	}
+	SetCompatibilityAttributes(attrs, int64(cpu.NUMANodeID))
+	return attrs
+}
 
 // SetCompatibilityAttributes add attributes to enable compatibility (e.g. alignment) with other
 // DRA resource drivers leveraging attributes which are not kubernetes standard.
--- a/pkg/device/builder.go
+++ b/pkg/device/builder.go
@@ -246,16 +246,7 @@ func createCPUDeviceSlices(deviceInfos []cpuDeviceInfo, pcieRootMapper *store.PC
 	var allDevices []resourceapi.Device
 	for _, deviceInfo := range deviceInfos {
 		cpu := deviceInfo.cpu
-		deviceAttrs := map[resourceapi.QualifiedName]resourceapi.DeviceAttribute{
-			AttributeNUMANodeID: {IntValue: new(int64(cpu.NUMANodeID))},
-			AttributeSocketID:   {IntValue: new(int64(cpu.SocketID))},
-			AttributeSMTEnabled: {BoolValue: new(smtEnabled)},
-			AttributeCacheL3ID:  {IntValue: new(int64(cpu.UncoreCacheID))},
-			AttributeCoreType:   {StringValue: new(cpu.CoreType.String())},
-			AttributeCoreID:     {IntValue: new(int64(cpu.CoreID))},
-			AttributeCPUID:      {IntValue: new(int64(cpu.CpuID))},
-		}
-		SetCompatibilityAttributes(deviceAttrs, int64(cpu.NUMANodeID))
+		deviceAttrs := CPUAttributes(cpu, smtEnabled)
 		addPCIeRootsAttribute(pcieRootMapper, deviceAttrs, cpu.CpuID)
 
 		cpuDevice := resourceapi.Device{
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -302,6 +302,37 @@ func (cp *CPUDriver) prepareDevices(logger logr.Logger, claim *resourceapi.Resou
 		preparedDevices = append(preparedDevices, preparedDevice)
 	}
 
+	if cp.cpuDeviceMode == device.CPU_DEVICE_MODE_GROUPED && claimCPUSet.Size() > 0 {
+		var requestNames []string
+		for _, allocResult := range claim.Status.Allocation.Devices.Results {
+			if allocResult.Driver == cp.driverName && allocResult.Request != "" {
+				requestNames = append(requestNames, allocResult.Request)
+			}
+		}
+		for _, cpuID := range claimCPUSet.List() {
+			cpuInfo, ok := cp.topology.cpuTopology.CPUDetails[cpuID]
+			if !ok {
+				logger.Error(nil, "CPU not found in topology", "cpuID", cpuID)
+				continue
+			}
+			attrs := device.CPUAttributes(cpuInfo, cp.topology.cpuTopology.SMTEnabled)
+
+			metadataAttrs := make(map[string]resourceapi.DeviceAttribute, len(attrs))
+			for k, v := range attrs {
+				metadataAttrs[string(k)] = v
+			}
+			preparedDevices = append(preparedDevices, kubeletplugin.Device{
+				PoolName:   cp.nodeName,
+				DeviceName: fmt.Sprintf("%s%03d", device.CPUDevicePrefix, cpuID),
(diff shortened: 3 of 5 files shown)


pravk03:
Thanks @johnahull. I started looking at this PR and I have a few questions worth discussing:

1. The resolved cpuset is already available to the container via the `DRA_CPUSET_<claimUID>` env var, and the group topology attributes are already published via the downward API metadata added in #265. A combination of these should give all the information ?. Maybe we can also expose the allocatedCPUSet as a downward attribute. 

2. [KEP-5304](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/5304-dra-attributes-downward-api/README.md) it is recommended that the devices in the metadata file are the devices allocated to the request. Since we add synthetic devices here that does not exist in ResourceSlice or allocated devices, I am wondering if there is it is ok, if there is any validation ?

3. Adding more devices in PrepareResult has side effect that Kubelet records every prepared device from the response. With [ResourceHealthStatus](https://github.com/kubernetes/enhancements/issues/4680) we would end up exposing one entry per allocated CPU in `pod.status.allocatedResourcesStatus`. With large requests, this would end up increasing the pod object size.

4. In the current implementation, a multi-request claim would end up with all the claim's CPUs listed under every request, I think each request should only include the device information specific to that request. 

5. To me, it seems like most workloads may not need this level of detailed information, and I think we may want a config option to enable this. 

Let's also wait for @ffromani to be back from vacation and provide input on the general direction here.

johnahull:
> Thanks @johnahull. I started looking at this PR and I have a few questions worth discussing:
> 
> 1. The resolved cpuset is already available to the container via the `DRA_CPUSET_<claimUID>` env var, and the group topology attributes are already published via the downward API metadata added in [feat: publish device attributes as KEP-5304 metadata #265](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/265). A combination of these should give all the information ?. Maybe we can also expose the allocatedCPUSet as a downward attribute.

I agree that for NUMA-grouped mode (the common case), the group device's numaNodeID plus the CDI env var cpuset covers the consumer's needs. Per-CPU entries are mainly valuable for socket/machine modes where a group spans multiple NUMA nodes, and for SMT sibling awareness. The question is whether those cases justify the approach. Not sure it does, but I thought I'd post anyway. 

> 2. [KEP-5304](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/5304-dra-attributes-downward-api/README.md) it is recommended that the devices in the metadata file are the devices allocated to the request. Since we add synthetic devices here that does not exist in ResourceSlice or allocated devices, I am wondering if there is it is ok, if there is any validation ?

I checked KEP-5304 — it doesn't require metadata devices to match allocated devices. Drivers control what they publish. But I take the point that it's unconventional.

> 3. Adding more devices in PrepareResult has side effect that Kubelet records every prepared device from the response. With [ResourceHealthStatus](https://github.com/kubernetes/enhancements/issues/4680) we would end up exposing one entry per allocated CPU in `pod.status.allocatedResourcesStatus`. With large requests, this would end up increasing the pod object size.

Agree this is a big concern and probably a blocker. 

> 4. In the current implementation, a multi-request claim would end up with all the claim's CPUs listed under every request, I think each request should only include the device information specific to that request.

That's definitely a bug, I'll fix if we proceed.

> 5. To me, it seems like most workloads may not need this level of detailed information, and I think we may want a config option to enable this.

This makes sense

Given point 3, I think the per-CPU device approach may not be the right path. Not sure how we'd work around that. I like the idea of the allocatedCPUSet string attribute on the group device (the resolved cpuset, like allocatedNumCPUs but with the actual CPU list). That gives consumers the allocation info in metadata without synthetic devices or pod status bloat. Consumers that need per-CPU NUMA topology in socket/machine mode could fall back to sysfs.

AutuSnow:
I agree with the concerns raised above, especially points 2-4. PrepareResult.Devices is not a metadata-only channel: kubelet persists every returned device in the prepared claim state, and ResourceHealthStatus may also expose each synthetic CPU through Pod status.
There is also a request-scoping issue here. requestNames is collected from all allocation results without deduplication, while every synthetic CPU comes from the claim-wide claimCPUSet. A multi-request claim therefore exposes the full CPU union to every request, and a single request allocated from multiple grouped devices can receive duplicate CPU entries.
I don't think the per-CPU synthetic-device approach should merge in its current form. Keeping the real grouped devices and attaching request-scoped allocation information to their metadata seems like a cleaner direction, although a single cpuset string would also need to account for the DeviceAttribute length limit.

ffromani:
I may be wrong, and I'm back from like 48hours so I can obviously be missing a lot of context, but still:
it seems to me there's an interesting commonality to explore between this work and the issue raised in #290 

In both cases, it seems to me there's a need for a 3rd party entity to learn highly detailed node topology data.
The gap, however, is that what's missing is a node topology information source, which happened to be incidentally addressed by the resource slice in individual mode (#290) and which we are trying to fit (partially?) in the metadata, but doesn't seem to belong here for the reason already raised.

In #264 I argued (https://github.com/kubernetes-sigs/dra-driver-cpu/issues/264#issuecomment-5126988340) that pods running on the same node already have access to the topology discovery API - linux sysfs at the cost of reimplementing the parsing, and I called it fair game, but I stated I'm open to make consumer's life easier _provided the change are cheap_. I think that comment aged well and this convo further reinforces that direction: providing more data is costly, doesn't scale, and likely adding extra information in the metadata is the wrong approach.
As it stands now, the only real gap I see for node-local consumers is to make sure the mapping the driver does and the mapping THEY do matches; IOW, we miss _a_ way to reconcile the representations the actors take from the same source (linux sysfs). I think **this** (make sure the representations match) is a fair ask we should address. **A** possible way which I want to mention is to force the driver to adhere to a particular representation, and I think we can get there extending the linux sysfs overlay work we begun earlier in the cycle.

It seems now from #290 that control plane components (which we cannot assume they run on the node or have access to the node sysfs!) have a similar need: they need a topology discovery API. As it stands today, I do *not* think that resource slice can be that API, nor any existing DRA-related (core kubernetes) API can do that. It seems it's emerging the desire to have a new complementary API, which may or may not somehow be under the DRA umbrella.
As it stands today, that topology discovery API seems however related but not a direct responsability of this driver, that should be a consumer more than a producer.

Comments and thoughts welcome to check if this train of thought can and should be expanded.

johnahull:
Thanks everyone for the feedback.

I agree the per-CPU synthetic device approach isn't the right path. The pod status bloat (point 3), request scoping issues (point 4), and ffromani's point about metadata not being the right place for detailed topology are all convincing.

ffromani's framing makes sense to me. The real gap is topology representation consistency between the driver and consumers reading sysfs, not shipping topology data through metadata. The sysfs overlay direction sounds like the right way to address that.

Closing this PR. The base metadata from #265 (group device attributes + allocatedNumCPUs) covers the common case.

PR #291, opened 2026-08-10
fix: report why kubelet did not register the plugin

### LINKED ISSUE #254: Registration timeout omits the kubelet's last rejection reason
Found while reviewing #244.

The driver waits for registration in `pkg/driver/driver.go` with a 30-second poll:

```go
status := d.RegistrationStatus()
if status == nil {
    return false, nil
}
return status.PluginRegistered, nil
```

`status.Error` is never surfaced, so when the kubelet keeps reporting a rejected status the caller only sees the poll timeout, not the reason the kubelet gave.

A safer fix than returning early is to keep the retry and attach the reason only at the end: while `PluginRegistered` is false, record the latest non-empty `status.Error` but keep polling (registration can retry from scratch after a transient failure, so a single `false` is not necessarily terminal); on timeout, wrap the error with that last recorded reason; and if no status was ever received, say so and include the registrar path.

Worth noting: a wrong kubelet root usually does *not* produce a `status.Error`. If the socket sits in a directory the kubelet does not watch, the kubelet never discovers it and `RegistrationStatus()` stays `nil` (the shape seen in #231). The cases this change would actually improve are kubelet-side rejections — unsupported API/service version, driver-name or metadata rejection, or a handler registration failure. Glad to pick this up.

thc1006:
#### What type of PR is this?

/kind bug

`Start` waits up to 30 seconds for kubelet to acknowledge the plugin, polling `RegistrationStatus()` and looking only at `PluginRegistered`. The same status carries the reason kubelet gives when it turns the plugin down, and that field was dropped, so a driver that kubelet keeps rejecting fails with a bare `context deadline exceeded` and nothing about why.

Bailing out as soon as a rejection appears would be wrong: kubelet retries registration from scratch, so a single unregistered status is not the end of the attempt. The wait keeps polling and only remembers the newest non-empty reason, which it attaches if the timeout is reached.

Three outcomes are worth telling apart, and the error now says which one happened:

- kubelet rejected the plugin and said why, and the reason is quoted;
- kubelet reported the plugin as unregistered but gave no reason;
- kubelet never reported at all, and the message names the registrar directory.

The third case is the shape seen in #231: when the socket sits outside the directory kubelet watches, kubelet never discovers it, `RegistrationStatus()` stays nil, and no reason is ever produced, so pointing at the registrar path is the only useful thing left to say. The cases where a reason does show up are kubelet-side rejections, such as an unsupported plugin API version, a driver name or metadata rejection, or a handler registration failure.

Only a wait that actually ran out gets explained, keyed off `context.DeadlineExceeded` the way the DRA scheduler plugin keys its own binding timeout. A cancelled context means the caller is shutting down and already carries its reason, so it passes through untouched rather than collecting advice about kubelet that does not apply.

Fixes #254

The wait moved into `waitForRegistration` so a fake registrar can drive it, following how `runNRIPluginWithRetry` is structured and tested. The poll interval and timeout became parameters so the tests do not have to sit through the real 30 seconds; the call site passes the existing 1s and 30s values, so nothing about the timing changes.

I deliberately left out logging the rejection as it arrives. Kubelet can report the same refusal on every one-second poll, so a log line per poll would be thirty repeats of one fact, and the failure already surfaces through `driver failed to start`, which is the shape `docs/dev/logging.md` asks for: fail the flow, and let the error point at the fix. If you would rather see it sooner than the timeout, a single `V(2)` line the first time a reason appears would be easy to add.

Naming `*registerapi.RegistrationStatus` in the small interface promotes `k8s.io/kubelet` from an indirect to a direct requirement in `go.mod`. It is the same v0.36.0 already in the module graph and matches the other `k8s.io` modules, so nothing new is pulled in. If you would rather not take the direct requirement, I am glad to keep the wait inline in `Start` instead, at the cost of the unit test.

#### Release note

```release-note
The CPU DRA driver now reports the reason kubelet gave when plugin registration times out, rather than only the timeout itself.
```

#### Documentation updates

NONE

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -18,6 +18,7 @@ package driver
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand/v2"
 	"os"
@@ -39,12 +40,17 @@ import (
 	drametadatav1alpha1 "k8s.io/dynamic-resource-allocation/api/metadata/v1alpha1"
 	"k8s.io/dynamic-resource-allocation/kubeletplugin"
 	"k8s.io/dynamic-resource-allocation/resourceslice"
+	registerapi "k8s.io/kubelet/pkg/apis/pluginregistration/v1"
 	"k8s.io/utils/cpuset"
 )
 
 const (
 	// maxAttempts indicates the number of times the driver will try to recover itself before failing
 	maxAttempts = 5
+	// registrationPollInterval and registrationTimeout bound the wait for kubelet to
+	// acknowledge the plugin at startup
+	registrationPollInterval = 1 * time.Second
+	registrationTimeout      = 30 * time.Second
 )
 
 const opIDLen = 8
@@ -52,6 +58,7 @@ const opIDLen = 8
 // KubeletPlugin is an interface that describes the methods used from kubeletplugin.Helper.
 type KubeletPlugin interface {
 	PublishResources(context.Context, resourceslice.DriverResources) error
+	RegistrationStatus() *registerapi.RegistrationStatus
 	Stop()
 }
 
@@ -293,14 +300,7 @@ func (cp *CPUDriver) Start(ctx context.Context) (<-chan error, error) {
 		return asyncErr, fmt.Errorf("start kubelet plugin: %w", err)
 	}
 	cp.draPlugin = d
-	err = wait.PollUntilContextTimeout(ctx, 1*time.Second, 30*time.Second, true, func(context.Context) (bool, error) {
-		status := d.RegistrationStatus()
-		if status == nil {
-			return false, nil
-		}
-		return status.PluginRegistered, nil
-	})
-	if err != nil {
+	if err := waitForRegistration(ctx, d, registrarDir(cp.kubeletRootDir), registrationPollInterval, registrationTimeout); err != nil {
 		return asyncErr, err
 	}
 
@@ -354,6 +354,42 @@ func getDeviceAttributes(deviceSlices [][]resourceapi.Device, deviceName string)
 func (cp *CPUDriver) Shutdown(ctx context.Context) {
 	logger := ctxlog.FromContext(ctx)
 	logger.Info("runtime shutting down")
+}
+
+// waitForRegistration waits for kubelet to report the plugin as registered. On timeout
+// it reports the last reason kubelet gave for refusing, or that it never reported at all.
+func waitForRegistration(ctx context.Context, p KubeletPlugin, registrarPath string, interval, timeout time.Duration) error {
+	logger := ctxlog.FromContext(ctx)
+	var lastRejection string
+	var sawStatus bool
+	err := wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(context.Context) (bool, error) {
+		status := p.RegistrationStatus()
+		if status == nil {
+			return false, nil
+		}
+		sawStatus = true
+		// Kubelet retries from scratch, so keep the newest reason but do not stop here.
+		// Only the newest survives to the error, so an earlier one is logged instead of lost.
+		if status.Error != "" && status.Error != lastRejection {
+			if lastRejection != "" {
+				logger.Info("kubelet gave a new reason for refusing the plugin", "previous", lastRejection, "current", status.Error)
+			}
+			lastRejection = status.Error
+		}
+		return status.PluginRegistered, nil
+	})
+	// Only a timeout is worth explaining; a cancelled context is the caller shutting down.
+	if !errors.Is(err, context.DeadlineExceeded) {
+		return err
+	}
+	switch {
+	case lastRejection != "":
+		return fmt.Errorf("kubelet did not register the plugin, last rejection was %q: %w", lastRejection, err)
+	case sawStatus:
+		return fmt.Errorf("kubelet did not register the plugin and reported no reason: %w", err)
+	default:
+		return fmt.Errorf("kubelet never reported a registration status, check that it watches %s: %w", registrarPath, err)
+	}
 }
 
 type nriRunner interface {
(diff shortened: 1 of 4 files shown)


pkg/driver/driver.go
```diff
@@ -356,6 +355,41 @@ func (cp *CPUDriver) Shutdown(ctx context.Context) {
...
+	err := wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(context.Context) (bool, error) {
+		status := p.RegistrationStatus()
+		if status == nil {
+			return false, nil
+		}
+		sawStatus = true
+		// Kubelet retries from scratch, so keep the newest reason but do not stop here.
+		if status.Error != "" {
+			lastRejection = status.Error
+		}
```
ffromani: changing the rejection from "" to <anything> is fine. Can the lastRejection ever change between two non-empty errors? if this ever happens (not sure) we should log the previous error before to overwrite

pkg/driver/driver.go
```diff
@@ -356,6 +355,41 @@ func (cp *CPUDriver) Shutdown(ctx context.Context) {
 	logger.Info("runtime shutting down")
 }
 
+type pluginRegistrar interface {
+	RegistrationStatus() *registerapi.RegistrationStatus
+}
```
ffromani: there's no need for a new interface, I reckon. We can add this method to the existing KubeletPlugin interface, so `waitForRegistration` can take a `p KubeletPlugin`. This will simplify the code and the mocks - we can extending the existing one.

Minimal interface is usually a noble goal, but in this case the extension seems a natural fit of the existing code shape, and leads to an overall code reduction, which seems a good tradeoff.

pkg/driver/driver_test.go
```diff
@@ -102,6 +104,110 @@ func TestRunNRIPluginWithRetry_SuccessfulRunNoRetry(t *testing.T) {
...
 }
 
+type mockPluginRegistrar struct {
+	statusFunc func(call int32) *registerapi.RegistrationStatus
+	calls      atomic.Int32
+}
+
+func (m *mockPluginRegistrar) RegistrationStatus() *registerapi.RegistrationStatus {
+	return m.statusFunc(m.calls.Add(1))
+}
```
ffromani: please extend `mockKubeletPlugin` to absorb this code

pkg/driver/driver_test.go
```diff
@@ -102,6 +104,110 @@ func TestRunNRIPluginWithRetry_SuccessfulRunNoRetry(t *testing.T) {
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			registrar := &mockPluginRegistrar{
+				statusFunc: func(call int32) *registerapi.RegistrationStatus {
+					return tc.status(cancel, call)
+				},
+			}
+
+			err := waitForRegistration(ctx, registrar, registrarPath, time.Millisecond, 200*time.Millisecond)
```
ffromani, line 184: every time we test private code/functions/structs we should pause and look hard and long if we can avoid it. Testing private functions is usually the lesser evil, and there's no hard rule about avoiding it, but a strong preference to do it.
In this case, the pragmatic choice is to _keep_ this test, because the real tension is that `driver.Start` does a lot and does not allow dependency injection. A future (probably far future) should address this issue, and enable us to test that portion in isolation. Up until now, adding tests *and documenting why we have this exception and we do test private function* is fine.
Please add a comment stating the above in the test code, so we can move on.

ffromani (commented):
thanks for the fix, some comments inline.

pkg/driver/driver.go
ffromani: note that you IMO **correctly** choose not to log to default the rejection reason after each poll. This part is fine. I'm looking at a much narrower, hopefully infrequent and therefore unexpected case: the rejection reason _changes_. That change, which should happen way less than the N attempt max, is IMO worth logging, not the 30 or so identical rejections.

thc1006: Agreed, and it can. Kubelet retries registration from scratch, so a plugin that fails an early handshake and then hits a different problem gives two different non-empty reasons, and only the newest reached the error. It now logs the previous one when the reason changes, and the comparison also stops it logging the same reason on every poll.

This comment was written in part with the assistance of generative AI. Bcuz my English might be sucks.

pkg/driver/driver.go
thc1006: You are right, and it did come out shorter. `pluginRegistrar` is gone, `KubeletPlugin` carries `RegistrationStatus()`, and `waitForRegistration` takes a `KubeletPlugin`. The real plugin already had the method, so only the interface changed.

This comment was written in part with the assistance of generative AI. Bcuz my English might be sucks.

pkg/driver/driver_test.go
thc1006: Done, `mockPluginRegistrar` is gone and `mockKubeletPlugin` absorbed it. `statusFunc` is left nil by the tests that do not exercise registration, and `RegistrationStatus` returns nil in that case, which is the same thing kubelet reports before it has looked.

This comment was written in part with the assistance of generative AI. Bcuz my English might be sucks.

pkg/driver/driver_test.go
thc1006: That is a fair rule and I would rather it were written down than assumed. Added above the test:

```go
// TestWaitForRegistration covers an unexported function, which we would normally
// reach through its caller instead. CPUDriver.Start does a lot and takes no
// injectable dependencies, so there is no seam to reach this behaviour from
// outside the package today. Testing it directly is the exception rather than the
// pattern, and it can move behind Start once Start is separable.
```

This comment was written in part with the assistance of generative AI. Bcuz my English might be sucks.

ffromani (commented):
thanks for addressing my comments quickly.

@thc1006 I appreciate the full disclosure about using gen AI, this is very transparent and open. Please be aware that it is unnecessary: to use gen AI to translate from languages, or to improve a human-created text, is totally fine and doesn't require such level of transparency. You are welcome to keep adding the notes if you wish so, but I just want to clarify that adding those is _not_ expected, let alone required.

PR #292, opened 2026-08-10
refactor: extract layered config resolution pipeline

ffromani:
#### What type of PR is this?

/kind cleanup

The major design items of our current config management flow are:
- we have a set of possible settings, represented by a struct with possibly nested substructs holding only the tunable values.
- we have a set of possible config sources - config file, flags...
- each config source can mutate a subset of tunables, and doesn't mutate what is not set explicitly
- we can therefore compose the configuration loading/resolution stage as a pipeline on which we start from a default state
  and we iteratively apply layers of mutations, each potentially partial, each coming from a config source.

This PRs extracts an explicit `Source` interface and we add a new Resolve([]Source) pipeline, so each source (defaults, config file, CLI flags) becomes a composable, testable unit, as demonstrated in the `gatherinfo` changes.

The goal is to make the existing structure fully explicit and fully surfaced, which makes the code more composable and testable

N/A

This work started from explorations I did in #229 and #262 ; I think it earned its own PR and is worth to be added. #262 will be rebased on top of it and finally I'll use (albeit, admitedly, marginally) #262 to implement testing coverage for #229

Worth stressing that this PR is all about remixing and repackaging existing concepts; there are not new ideas here, but "just" a different combination of concepts we already discussed and implemented.

AI usage disclosure: the idea and the first 2 drafts of this PR were entirely human-generated. After that, I used AI to self-review, polish, rebase.

#### Release note

```release-note
NONE
```

#### Documentation updates

N/A

### DIFF
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -111,7 +111,14 @@ func main() {
 		return
 	}
 
-	cfg, err := driverconfig.Load(driverFlags, configFile, flag.CommandLine, logger)
+	cfg, err := driverconfig.Resolve(logger, []driverconfig.Source{
+		// the ordering of Sources matters and determines the
+		// precedence of layers (aka overrides).
+		// Per our current policy, the priority (rightmost wins)
+		// is defaults -> config -> flags.
+		driverconfig.FromFile(configFile),
+		driverconfig.FromFlags(flag.CommandLine),
+	})
 	if err != nil {
 		logger.Error(err, "failed to load configuration")
 		os.Exit(1)
--- a/cmd/dracpu/app.go
+++ b/cmd/dracpu/app.go
@@ -55,7 +55,7 @@ const (
 )
 
 var (
-	driverFlags = driverconfig.Default()
+	driverFlags driverconfig.Config
 	configFile  string
 
 	ready atomic.Bool
@@ -96,12 +96,10 @@ func main() {
 			fmt.Fprintln(os.Stderr, "dracpu: root flags cannot be combined with subcommands")
 			os.Exit(1)
 		}
-
 		opts := subcommands.Options{
-			DriverConfig: driverFlags,
-			Logger:       logger,
-			Stdout:       os.Stdout,
-			Stderr:       os.Stderr,
+			Logger: logger,
+			Stdout: os.Stdout,
+			Stderr: os.Stderr,
 		}
 
 		if err := subcommands.Run(flag.Args(), opts); err != nil {
--- a/internal/driverconfig/cfgfile.go
+++ b/internal/driverconfig/cfgfile.go
@@ -17,8 +17,6 @@ limitations under the License.
 package driverconfig
 
 import (
-	"bytes"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
@@ -27,8 +25,35 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/go-logr/logr"
 	"sigs.k8s.io/yaml"
 )
+
+type FileSource struct {
+	confPath string
+}
+
+func FromFile(confPath string) FileSource {
+	return FileSource{
+		confPath: confPath,
+	}
+}
+
+func (fs FileSource) Name() string {
+	return fs.confPath
+}
+
+func (fs FileSource) Apply(logger logr.Logger, cfg *Config) error {
+	if fs.confPath == "" {
+		return nil // nothing to do
+	}
+	overrides, err := buildConfMap(fs.confPath)
+	if err != nil {
+		return err
+	}
+	logger.V(6).Info("overrides", "stage", fs.Name(), "values", overrides)
+	return applyMap(cfg, overrides)
+}
 
 // buildConfMap loads the config file at filePath, validates and strips
 // "apiVersion", and returns the resulting map.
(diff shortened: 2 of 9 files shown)


cmd/dracpu/app.go
```diff
@@ -111,7 +111,11 @@ func main() {
 		return
 	}
 
-	cfg, err := driverconfig.Load(driverFlags, configFile, flag.CommandLine, logger)
+	cfg, err := driverconfig.Resolve(logger, []driverconfig.Source{
+		// overrides are applied in the given order
```
pravk03: nit: Since this ordering is important for the design, maybe add a comment explicitly indicating the order must not be modified? For future reference.

internal/driverconfig/load.go
```diff
@@ -62,61 +63,43 @@ func WarnDeprecatedFlags(fs *flag.FlagSet, logger logr.Logger) {
...
-		}
+func Resolve(logger logr.Logger, sources []Source) (Config, error) {
+	cfg := Default()
+	logger.WithValues("stage", "default").V(6).Info("config", cfg.LogValues()...)
 
-		if err := applyMap(&result, confMap); err != nil {
-			return Config{}, fmt.Errorf("applying config file %q: %w", filePath, err)
+	for _, src := range sources {
+		if err := src.Apply(logger, &cfg); err != nil {
+			return Config{}, err
```
pravk03: nit: Maybe we can include what is the source of the error (flags or configfile)

cmd/dracpu/app.go
```diff
@@ -91,14 +91,19 @@ func main() {
...
+	})
+	if err != nil {
+		logger.Error(err, "failed to load configuration")
+		os.Exit(1)
+	}
 
+	if len(flag.Args()) > 0 {
 		opts := subcommands.Options{
-			DriverConfig: driverFlags,
+			DriverConfig: cfg,
```
pravk03: nit: Can be dropped I think. I do not see this being consumed.

pravk03 (commented):
Thanks a lot for this cleanup! 

/lgtm 
/hold 
(in case we want to address the nits)

cmd/dracpu/app.go
ffromani: Absolutely, good idea!

cmd/dracpu/app.go
ffromani: yes, correct, this is a leftover from the extraction from #262 . Unnecessary here.

internal/driverconfig/load.go
ffromani: it seems a good idea to augment the error there. Done.

ffromani:
/unhold

updated, so LGTM is gone

internal/driverconfig/config_test.go
```diff
@@ -215,31 +216,28 @@ groupby: socket
...
-	_, err := driverconfig.Load(cfg, cfgFile, fs, testr.New(t))
+	_, err := driverconfig.Resolve(testr.New(t), []driverconfig.Source{
+		driverconfig.FromFile(cfgFile),
+	})
 
 	require.Error(t, err)
 	assert.EqualError(t, err,
-		`config file "`+cfgFile+`": `+
-			`field "cpudevicemode" is spelled differently from the schema; use "cpuDeviceMode"; `+
+		`field "cpudevicemode" is spelled differently from the schema; use "cpuDeviceMode"; `+
```
AutuSnow, line 234: The latest head now wraps source errors with `cannot apply <source>:`, but this assertion still expects the unwrapped message. This makes `make test-unit` fail both locally and in the current build check. Could we update the expected error to include `cfgFile` while preserving the exact canonical-key assertions?

cmd/dracpu/app.go
```diff
@@ -92,16 +92,10 @@ func main() {
 	driverconfig.WarnDeprecatedFlags(flag.CommandLine, logger)
 
 	if len(flag.Args()) > 0 {
```
AutuSnow, line 94: Could we keep rejecting driver configuration flags when dispatching a subcommand, or scope the relaxed behavior to flags that subcommands actually consume? With this guard removed, `dracpu --config=/does/not/exist introspect metrics` succeeds and silently ignores `--config`.

internal/driverconfig/config_test.go
ffromani: yes, absolutely. Had a fix locally and forgot to push. Should be addressed now.

cmd/dracpu/app.go
ffromani: good point, let me think about it

ffromani: after a closer inspection, this turned to be another leftover from #262. We will rediscuss this specific change in the context of #262 - but maybe I found a possible alternative. Anyway, not relevant for this PR.

internal/driverconfig/cfgfile.go
```diff
@@ -27,9 +25,36 @@ import (
...
+func (fs FileSource) Name() string {
+	return fs.confPath
+}
+
+func (fs FileSource) Apply(logger logr.Logger, cfg *Config) error {
+	if fs.confPath == "" {
+		return nil // nothing to do
+	}
+	overrides, err := buildConfMap(fs.confPath)
+	logger.V(6).Info("overrides", "stage", fs.Name(), "values", overrides)
```
fmuyassarov: maybe we move this after we check that the we didn't get an error from buildConfMap()?

fmuyassarov (commented):
one question, otherwise LGTM

internal/driverconfig/cfgfile.go
ffromani: right! it should be after the if guard. Fixing!

fmuyassarov (commented):
Assuming all the comments are addressed according the thread, I'm adding my LGTM
/lgtm
Thanks @ffromani

PR #293, opened 2026-08-12
chore: drop NET_ADMIN & SYS_ADMIN from daemonset

fmuyassarov:
#### What type of PR is this?
/kind cleanup

These capabilities are redundant because the driver applies cpuset changes via the NRI which the container runtime enforces on its behalf. So the driver does not really need `NET_ADMIN` or `SYS_ADMIN`.  I don't recall where exactly, but I do remember @pravk03 has asked in the past about similar changes. 

I've done very simple testing with exclusive CPUs allocation just to double check on top of e2e, which I hope would/should catch in case with face some unexpected bumps. 

#### Release note

```release-note
manifests: drop NET_ADMIN & SYS_ADMIN from daemonset
```

#### Documentation updates
- README updates

### DIFF
--- a/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
+++ b/deployment/helm/dra-driver-cpu/templates/daemonset.yaml
@@ -101,9 +101,6 @@ spec:
           limits:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-        securityContext:
-          capabilities:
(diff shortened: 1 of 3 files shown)


ffromani:
/approve
/lgtm

these were most likely relics of when we had the setup script as init container.
The driver should NOT require these privileges. If we need them again, we better have a strong and documented reason to, so I'm fine with a quick merge.

Thanks for this important cleanup!

Issue #256, opened 2026-07-25
extraArgs can override --bind-address without updating health probes


thc1006:
Found while reviewing #244; pre-existing behavior, low priority.

The chart renders `--bind-address=:{{ .Values.healthzPort }}` and then appends `extraArgs`, and Go's `flag` package takes the last value for a repeated flag. So:

```yaml
extraArgs:
  - --bind-address=:9999
```

makes the driver listen on `:9999` while the readiness and liveness probes still target the `healthz` port (`:8080`). The pod never becomes Ready, and since the liveness probe also targets `:8080`, it then keeps getting restarted.

This is the general `extraArgs` escape hatch, but `--bind-address` is the case that produces a concrete chart-internal inconsistency; I have not shown the same for every chart-owned flag. A minimal, non-breaking fix mirrors the existing `--config` guard: reject `--bind-address` / `-bind-address` (with or without `=`) in `extraArgs` and point users at `healthzPort`, with a small Helm render test asserting the template fails. Happy to take whichever direction the maintainers prefer.

AutuSnow:
@fmuyassarov @thc1006 I saw that extraArgs has been deleted in the submission of #294. Can this lssue be closed and reopened if necessary

/close

fmuyassarov:
Yes I think we are good to close it now

thc1006:
Thanks @AutuSnow and @fmuyassarov help!

PR #294, opened 2026-08-12
chore: remove extraArgs from helm chart

fmuyassarov:
#### What type of PR is this?
/kind cleanup

`extraArgs` allows override any helm chart managed flags (which is unintentional I believe). It only existed as a side effect of generalizing the old sysfsOverlay and most of the flags are already covered by args or driverConfig. For example, helm chart already exposes its own flag to control the sysfs overlay with `--set driverConfig.sysfsOverlay` and this is what's used also in the example here https://github.com/kubernetes-sigs/dra-driver-cpu/blob/4fc973652a5a301f7077970a16f56ba35273658d/hack/examples/sysfs-overlay/values.yaml#L5
The other issue is, it forces extra precaution whenever we are touching config options/flags, kubeletRootDir is one example. 

Also, there is another minor side effect https://github.com/kubernetes-sigs/dra-driver-cpu/issues/256. Unless we have another use for it (which I would be interested to hear about), I’m not sure we would get much value from keeping this flag.

#### Release note

```release-note
NONE
```

#### Documentation updates
- README updates
- Helm chart docs

### DIFF
--- a/deployment/helm/dra-driver-cpu/README.md
+++ b/deployment/helm/dra-driver-cpu/README.md
@@ -42,7 +42,6 @@ helm install dra-driver-cpu oci://registry.k8s.io/dra-driver-cpu/charts/dra-driv
 | args.logLevel | int | `4` | Log verbosity level passed as `--v` |
 | args.reservedCPUs | string | `""` | **Deprecated:** folded into the generated `driverConfig` ConfigMap and takes priority over it; use `driverConfig.reservedCPUs` instead. CPUs reserved for the OS and kubelet, excluded from DRA management (e.g. `"0-1"`). |
 | driverConfig | object | `{}` | Driver config file contents. When non-empty, or when a deprecated `args.*` field below is set, a ConfigMap is created and mounted into the driver container as /etc/dracpu/config.yaml. `args.*` fields that mirror a deprecated CLI flag (`cpuDeviceMode`, `groupBy`, `reservedCPUs`, `hostnameOverride`) are folded into the generated config automatically and take priority over the same field set here. |
-| extraArgs | list | `[]` | Extra command-line arguments appended to the driver arguments |
 | extraVolumeMounts | list | `[]` | Extra volume mounts for the driver container |
 | extraVolumes | list | `[]` | Extra volumes for the DaemonSet pod |
 | fullnameOverride | string | `""` | Override the full release name |
--- a/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
+++ b/deployment/helm/dra-driver-cpu/templates/_helpers.tpl
@@ -62,13 +62,6 @@ check below for why that specific check exists.
 {{- fail (printf "value %q differs from kubeletRootDir only in case, and the chart reads the exact name: it would be ignored and the kubelet paths would render at the default" $key) -}}
 {{- end -}}
 {{- end -}}
-{{- /* Matched on the parsed flag name: the flag package treats -name and --name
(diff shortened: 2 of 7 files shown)


ffromani:
/lgtm

let's give a chance to @killianmuldoon to have a look

thc1006:
Heads up on a small interaction with #282, which adds a render check for the kubelet root wiring.

That PR has a case asserting this chart's `extraArgs` guard: `--kubelet-root-dir` and `-kubelet-root-dir` are refused, and `--kubelet-root-directory` is accepted as a different flag. Once `extraArgs` is gone the whole key is refused by the schema, so the accepted half stops being true.

Either order works. Whichever of the two lands second updates that block, and I'm happy for that to be mine.

thc1006:
Correcting myself: I said above that once `extraArgs` is gone the whole key is refused by the schema. It isn't, and I should have checked before writing that.

The top-level schema has no `additionalProperties: false`, so a key missing from `properties` is accepted and ignored. A known key at the wrong type still fails, so validation is running, it just lets the extra one through.

With your change applied on top of main, a values file carrying

```yaml
extraArgs:
  - --v=9
  - --this-flag-is-unique-to-the-test
```

renders with no error and neither flag reaches the container args. On current main the same file puts both of them there. So an upgrade that reuses an existing values file drops them quietly.

Your call whether that deserves an explicit error, given the release note is NONE today. Either way #282 has to rewrite that block, and I would rather assert a real message than the silence. Happy to send a tombstone guard as a follow-up if you want one.

fmuyassarov:
Had to push due to the rebase.

fmuyassarov:
> Correcting myself: I said above that once `extraArgs` is gone the whole key is refused by the schema. It isn't, and I should have checked before writing that.
> 
> The top-level schema has no `additionalProperties: false`, so a key missing from `properties` is accepted and ignored. A known key at the wrong type still fails, so validation is running, it just lets the extra one through.
> 
> With your change applied on top of main, a values file carrying
> 
> ```yaml
> extraArgs:
>   - --v=9
>   - --this-flag-is-unique-to-the-test
> ```
> 
> renders with no error and neither flag reaches the container args. On current main the same file puts both of them there. So an upgrade that reuses an existing values file drops them quietly.
> 
> Your call whether that deserves an explicit error, given the release note is NONE today. Either way #282 has to rewrite that block, and I would rather assert a real message than the silence. Happy to send a tombstone guard as a follow-up if you want one.

`extraArgs` isn't part of any release since it was introduced just recently. So there is no released version whose values file could have `extraArgs` and suddenly it becomes a problematic during upgrade. I don't think explicit error is really needed here.

killianmuldoon:
Thanks for fixing this - definitely a bug because of the way my sysfs overlay PR evolved. This looks good to me and doesn't interfere with the sysfs overlay behaviour.

/lgtm

fmuyassarov:
@ffromani @killianmuldoon 
Seems lgtm got dropped due to required rebase. Would you mind to re-stamp it once more?

killianmuldoon:
/lgtm

But not sure if you need an /approve from someone with rights.

fmuyassarov:
> /lgtm
> 
thanks
> But not sure if you need an /approve from someone with rights.

Yes, let's wait for Francesco

pravk03:
/approve 

Since @ffromani had added lgtm in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/294#issuecomment-5276851623 and there aren't any open review comments.

PR #297, opened 2026-08-13
fix: reject containers when shared CPU pool is empty

### LINKED ISSUE #295: Empty shared CPU pool is not safely enforced for shared containers
/kind bug

## What happened

I have recently been reading the project code and validating its DRA behavior in a Kind cluster.

When a DRA ResourceClaim consumes all driver-managed CPUs on a node, `CPUAllocation.GetSharedCPUs()` returns an empty CPUSet. The NRI hook then passes an empty `Cpus` value to the runtime.

With NRI v0.11, an empty `Cpus` value does not apply an empty CPUSet. As a result, a shared container can receive or retain a CPUSet that overlaps with CPUs exclusively assigned to a DRA Claim.

## Observed result

The DRA Claim was assigned CPUs `1-3`.

The exclusive container reported:

```text
allocation.cpus=1-3
runtimeinfo.affinity=1-3
```

A shared container created after the Claim was prepared reported:

```text
allocation.cpus=0-9
runtimeinfo.affinity=0-9
```

The driver log contained:

```text
"no guaranteed CPUs found, using shared CPUs" sharedCPUs=""
"updating CPU allocation for containers without guaranteed CPUs" sharedCPUs=""
```

Therefore, the shared container was allowed to run on CPUs `1-3`, which were exclusively assigned to the DRA Claim.

In a staged reproduction where the shared container was created before the exclusive Claim, it retained its previous CPUSet after the shared pool became empty.

## Expected behavior

When an exclusive DRA allocation empties the shared CPU pool, the driver should handle this state safely.

A shared container should not receive or retain a CPUSet that overlaps with a DRA-exclusive CPUSet.

## Reproduction

1. Run the driver in grouped mode with `groupBy=numanode`, `reservedCPUs=0`, and `cpuManagerPolicy=none`.
2. Create a shared Pod without a ResourceClaim.
3. Create a grouped ResourceClaim requesting all driver-managed CPUs.
4. Create a Pod consuming that Claim.
5. Create another shared Pod without a ResourceClaim.
6. Compare the reported cgroup CPU sets and inspect the driver log for `sharedCPUs=""`.

The Pods should use `nodeSelector` instead of `spec.nodeName` so that the Claim goes through the normal scheduler and DRA allocation flow.

## Relevant implementation

The shared CPUSet is calculated in `pkg/store/cpu_allocation.go` as:

```go
availableCPUs.Difference(allocatedCPUs)
```

The shared-container path in `pkg/driver/nri_hooks.go` then does:

```go
sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
adjust.SetLinuxCPUSetCPUs(sharedCPUs.String())
```

When `sharedCPUs` is empty, the resulting runtime update does not safely enforce CPU isolation.

## Environment

- DRA driver commit: `2e6c4737598a85790343634dd1bcef6fa2564d1c`
- Kubernetes: `v1.36.0`
- CPU allocation mode: `grouped`
- Grouping: `numanode`
- Reserved CPUs: `0`
- CPU Manager policy: `none`
- Container runtime: `containerd 2.3.1`
- NRI: `v0.11.0`
- Architecture: `arm64`
- Environment: Kind on OrbStack

alanhuangch:
## Summary

- Fail closed when a prepared DRA allocation leaves no CPUs for shared containers.
- Reject a shared container instead of passing an empty CPUSet to NRI.
- Reject an exclusive container when its allocation would exhaust the shared pool while shared containers still exist.
- Roll back newly recorded claim ownership when container creation is rejected.
- Apply the same empty-pool validation during container restart and NRI synchronization.
- Add unit and focused E2E coverage for the failure paths.

## Background

An empty CPUSet is serialized through NRI as an empty Cpus value. The runtime treats that value as no CPUSet update, so a shared container may retain or receive unrestricted CPU affinity and overlap CPUs assigned exclusively through a DRA claim.

The scheduler-side accounting changes prevent this during normal scheduling when node allocatable resource mapping is enabled. This change adds a node-local fail-closed check so the driver does not apply an unsafe empty CPUSet if the condition is still encountered.

A node with no driver-managed CPUs and no prepared DRA allocation keeps the existing kubelet-managed behavior.

## Testing

```text
go test ./pkg/... ./internal/...
go test ./test/e2e -run "^$"
```

Focused E2E test against an amd64 Kind cluster with Kubernetes v1.36.0, 24 CPUs, Node.Status.Allocatable.cpu=23, reservedCPUs=0, and driver-managed CPUs 1-23:

```text
Ran 1 of 17 Specs in 54.764 seconds
SUCCESS! -- 1 Passed | 0 Failed
```

The E2E test consumes all driver-managed CPUs through DRA claims and verifies that the container which exhausts the shared pool and a subsequent shared container are rejected without applying an overlapping CPUSet.

Fixes #295.

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -103,6 +103,13 @@ func getCDIDeviceName(uid types.UID) string {
 	return fmt.Sprintf("claim-%s", uid)
 }
 
+// reserveResourceClaimAllocation records a new claim allocation while applying
+// the shared-pool guard for currently running shared containers.
+func (cp *CPUDriver) reserveResourceClaimAllocation(logger logr.Logger, claimUID types.UID, cpus cpuset.CPUSet) error {
+	hasSharedContainers := len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0
+	return cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claimUID, cpus, hasSharedContainers)
+}
+
 func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *resourceapi.ResourceClaim) kubeletplugin.PrepareResult {
 	logger.V(4).Info("preparing grouped resource claim")
 
@@ -193,15 +200,8 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		return kubeletplugin.PrepareResult{}
 	}
 
-	// Refuse to exhaust the shared pool while shared containers are running: NRI
-	// cannot represent an empty CPUSet, so an emptied pool would leave shared
-	// containers with unsafe affinity (see #295).
-	if len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0 && allocatableCPUs.Difference(cpuAssignment).IsEmpty() {
-		return kubeletplugin.PrepareResult{Err: fmt.Errorf("claim %q would exhaust the shared CPU pool while shared containers are running", claim.UID)}
-	}
-
 	// Reserve before CDI I/O so concurrent Prepare calls cannot select the same CPUs.
-	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, cpuAssignment); err != nil {
+	if err := cp.reserveResourceClaimAllocation(logger, claim.UID, cpuAssignment); err != nil {
 		return kubeletplugin.PrepareResult{Err: err}
 	}
 	result := cp.prepareDevices(logger, claim, cpuAssignment)
@@ -262,15 +262,8 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 		}
 	}
 
-	// Refuse to exhaust the shared pool while shared containers are running: NRI
-	// cannot represent an empty CPUSet, so an emptied pool would leave shared
-	// containers with unsafe affinity (see #295).
-	if len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0 && allocatableCPUs.Difference(claimCPUSet).IsEmpty() {
-		return kubeletplugin.PrepareResult{Err: fmt.Errorf("claim %q would exhaust the shared CPU pool while shared containers are running", claim.UID)}
-	}
-
 	// Reserve before CDI I/O so concurrent Prepare calls cannot select the same CPUs.
-	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, claimCPUSet); err != nil {
+	if err := cp.reserveResourceClaimAllocation(logger, claim.UID, claimCPUSet); err != nil {
 		return kubeletplugin.PrepareResult{Err: err}
 	}
 	result := cp.prepareDevices(logger, claim, claimCPUSet)
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -193,6 +193,13 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		return kubeletplugin.PrepareResult{}
 	}
 
+	// Refuse to exhaust the shared pool while shared containers are running: NRI
+	// cannot represent an empty CPUSet, so an emptied pool would leave shared
+	// containers with unsafe affinity (see #295).
+	if len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0 && allocatableCPUs.Difference(cpuAssignment).IsEmpty() {
+		return kubeletplugin.PrepareResult{Err: fmt.Errorf("claim %q would exhaust the shared CPU pool while shared containers are running", claim.UID)}
+	}
+
 	// Reserve before CDI I/O so concurrent Prepare calls cannot select the same CPUs.
 	if err := cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claim.UID, cpuAssignment); err != nil {
 		return kubeletplugin.PrepareResult{Err: err}
@@ -253,6 +260,13 @@ func (cp *CPUDriver) prepareResourceClaim(logger logr.Logger, claim *resourceapi
 		return kubeletplugin.PrepareResult{
 			Err: fmt.Errorf("claim %s/%s has overlapping device assignment with other claims", claim.Namespace, claim.Name),
 		}
+	}
+
+	// Refuse to exhaust the shared pool while shared containers are running: NRI
+	// cannot represent an empty CPUSet, so an emptied pool would leave shared
+	// containers with unsafe affinity (see #295).
+	if len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0 && allocatableCPUs.Difference(claimCPUSet).IsEmpty() {
+		return kubeletplugin.PrepareResult{Err: fmt.Errorf("claim %q would exhaust the shared CPU pool while shared containers are running", claim.UID)}
 	}
 
 	// Reserve before CDI I/O so concurrent Prepare calls cannot select the same CPUs.
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -104,7 +104,9 @@ func getCDIDeviceName(uid types.UID) string {
 }
 
 // reserveResourceClaimAllocation records a new claim allocation while applying
-// the shared-pool guard for currently running shared containers.
+// the shared-pool guard for currently running shared containers. A shared
+// container from the same pod may not have been created yet when this DRA hook
+// runs, so that case is detected later by the NRI CreateContainer check.
 func (cp *CPUDriver) reserveResourceClaimAllocation(logger logr.Logger, claimUID types.UID, cpus cpuset.CPUSet) error {
 	hasSharedContainers := len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0
 	return cp.cpuAllocationStore.ReserveResourceClaimAllocation(logger, claimUID, cpus, hasSharedContainers)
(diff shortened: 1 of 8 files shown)


ffromani (commented):
/ok-to-test

alanhuangch:
This PR has been updated. It now includes the driver-level fail-closed protection discussed in #295, together with unit tests and focused E2E coverage. The focused E2E test passed, and the updated CI jobs are now running.

pkg/driver/nri_hooks.go
```diff
@@ -239,12 +264,23 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
...
 		// A new owner means this is the first CreateContainer after Prepare. On restart,
 		// the owner already exists and the shared cpuset has not changed.
 		if len(newOwners) > 0 {
-			updates = cp.getSharedContainerUpdates(logger, containerId)
+			updates, err = cp.getSharedContainerUpdates(logger, containerId)
+			if err != nil {
+				cp.claimTracker.Cleanup(newOwners...)
+				return nil, nil, err
+			}
+		} else {
```
pravk03: I don't think we need this else part as it would block an existing container restart. 

I think we should have the checks only for new container creations.

test/e2e/cpu_assignment_test.go
```diff
@@ -241,6 +243,101 @@ var _ = ginkgo.Describe("CPU Allocation", ginkgo.Serial, ginkgo.Ordered, ginkgo.
 				verifySharedPoolMatches(ctx, fxt, shrPod2, expectedSharedCPUs)
 			})
 
+			ginkgo.It("should reject a shared container when the shared pool becomes empty", ginkgo.Label("negative"), func(ctx context.Context) {
+				if cpuDeviceMode == "grouped" && groupBy == "machine" {
```
pravk03, line 246: We should also skip when `publishNodeAllocatableMapping` is true. The  scheduler side accounting  should prevent this.

pkg/driver/nri_hooks.go
```diff
@@ -213,10 +233,15 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
 
 	if len(claimAllocations) == 0 {
 		// This is a shared container.
+		sharedCPUs := cp.cpuAllocationStore.GetSharedCPUs()
```
pravk03, line 231: I think we can catch this issue even earlier in PrepareResourceClaims() (in dra_hooks.go).  If Prepare rejects instead, the shared pool never empties and existing and new shared containers would continue to use it. Only the pod that empties the shared pool remains blocked.

pravk03 (commented):
Thanks a lot for catching the issue and fixing it :) 

The NRI changes mostly look good, but I think it would be better to catch this earlier in DRA claim preparation (and also have the NRI checks). I have not reviewed the e2e test yet, since it would likely change once the review comment is addressed.

test/e2e/cpu_assignment_test.go
alanhuangch: Done in b63c1bc.

The empty-shared-pool E2E test now skips when publishNodeAllocatableMapping is enabled, because scheduler-side accounting prevents this exhaustion scenario.

pkg/driver/nri_hooks.go
alanhuangch: Done in b63c1bc.

The restart-only validation has been removed. NRI shared-container checks are now performed for new container creation, while an existing guaranteed container can restart without being blocked by an empty shared pool.

pkg/driver/nri_hooks.go
alanhuangch: Done in b63c1bc.

Both individual and grouped claims now use a guarded reservation during PrepareResourceClaims. If shared containers exist, a claim that would exhaust the shared CPU pool is rejected before CDI preparation.

The NRI checks are still kept as a defensive fallback for inconsistent or legacy state. I also fixed the rollback path so a repeated guarded reservation cannot remove an existing allocation.

Added unit coverage for both grouped and individual claims.

pkg/driver/nri_hooks.go
```diff
@@ -173,10 +176,27 @@ func validateSynchronizedClaimAllocation(logger logr.Logger, uid types.UID, cpus
...
+	// An empty CPUSet is serialized by NRI as Cpus="", which means "do not
+	// change the current CPUSet" rather than "clear the CPUSet". Never emit
+	// that update while a prepared DRA allocation has exhausted the pool and
+	// shared containers still exist. An empty pool with no prepared allocation
+	// is valid when the node has no driver-managed CPUs.
+	if sharedCPUs.IsEmpty() && !preparedCPUs.IsEmpty() && len(sharedCPUContainers) > 0 {
+		return fmt.Errorf("cannot update shared containers: no shared CPUs available")
+	}
+	return nil
+}
```
ffromani: this is basically a `if` used only once - so it should be inline in the call site, with the comment of course

pkg/store/cpu_allocation.go
```diff
@@ -65,19 +65,46 @@ func NewCPUAllocation(cpuTopology *cpuinfo.CPUTopology, reservedCPUs cpuset.CPUS
...
+// affinity. The check and the reservation happen atomically so that concurrent
+// Prepare calls cannot jointly exhaust the pool.
+func (s *CPUAllocation) ReserveResourceClaimAllocationGuarded(logger logr.Logger, claimUID types.UID, cpus cpuset.CPUSet, sharedContainerCount int) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	inserted, err := s.reserveResourceClaimAllocationLocked(logger, claimUID, cpus)
+	if err != nil {
+		return err
+	}
+	if sharedContainerCount > 0 && s.availableCPUs.Difference(s.preparedCPUs).IsEmpty() {
```
ffromani: ok, so the idea is that that is not legit to exhaust the shared pool _if there are existing container accessing it_. Fair. But I think we don't need a `ReserveResourceClaimAllocationGuarded` variant but rather to push this check in the calling site.
IOW, we should detect this scenario in the caller and reject to reserve resource claims if we detect this state. We know the size of the requested CPUs and the shared pool set already in the caller - we pass them to this function!

however, if the current code shape is the result of a prior discussion, please point me to that before to make changes; I haven't reviewed initially and I may miss context.

pkg/driver/nri_hooks.go
alanhuangch: Done in 4fa7ea7.

I inlined the empty-shared-pool check into getSharedContainerUpdates and kept the comment explaining the NRI empty CPUSet behavior.

Sorry, I introduced unnecessary encapsulation for a check that is only used at this call site. Thanks for pointing it out.

pkg/store/cpu_allocation.go
alanhuangch: Done in 4fa7ea7.

There was no prior discussion requiring the guarded store API; it was an implementation choice on my side.

I moved the shared-pool exhaustion check into both Prepare call sites, using the current shared pool and the requested CPUSet, and restored the original

ReserveResourceClaimAllocation API. The existing driver-level tests still cover both grouped and individual modes.

Thanks for the suggestion.

pkg/driver/dra_hooks.go
```diff
@@ -193,6 +193,13 @@ func (cp *CPUDriver) prepareGroupedResourceClaim(logger logr.Logger, claim *reso
 		return kubeletplugin.PrepareResult{}
 	}
 
+	// Refuse to exhaust the shared pool while shared containers are running: NRI
```
pravk03: nit: consider adding this check in ReserveResourceClaimAllocation() ?. The advantage is that the check would happen with the lock, and we do not have to repeat this code in multiple places.

pravk03 (commented):
Left one nit. Otherwise looks good. Thanks. 

/lgtm

pkg/driver/dra_hooks.go
alanhuangch: Done in a2edf46.

I moved the shared-pool exhaustion check into `ReserveResourceClaimAllocation()`, so the check and CPU reservation now run under the allocation lock.

The grouped and individual Prepare paths share a driver-level helper for detecting running shared containers, so the check is no longer duplicated. `Synchronize()` continues to restore existing allocations without applying the new-reservation guard.

Added store-level coverage for rejected and allowed reservations. The focused unit tests and race tests pass.

Please take another look when you have a chance.

pkg/driver/dra_hooks.go
```diff
@@ -103,6 +103,13 @@ func getCDIDeviceName(uid types.UID) string {
 	return fmt.Sprintf("claim-%s", uid)
 }
 
+// reserveResourceClaimAllocation records a new claim allocation while applying
+// the shared-pool guard for currently running shared containers.
+func (cp *CPUDriver) reserveResourceClaimAllocation(logger logr.Logger, claimUID types.UID, cpus cpuset.CPUSet) error {
+	hasSharedContainers := len(cp.podConfigStore.GetContainersWithSharedCPUs()) > 0
```
pravk03, line 111: Hmm.. the obvious gap here is that if the pod that is taking up all the remaining CPUs also has the shared container, we would miss this check here and is only caught in the NRI. But I do not see any way around it. So, its a not a blocker.
Optional: Maybe good to leave a comment ?

pravk03 (commented):
Thanks !!. The changes look good and all the comments are addressed. I see @ffromani's comments resolved as well, but let's give it a little more time in case there are any follow-ups before final lgtm.

/approve

pkg/driver/dra_hooks.go
alanhuangch: Good point, agreed.

I added a comment in `38353df` documenting this limitation: the DRA-side guard only covers shared containers that are already known when `PrepareResourceClaims()` runs. If a shared container from the same Pod has not been created yet, the condition is detected later by the NRI `CreateContainer` check.

Thanks for pointing this out.

pravk03:
Thanks. The changes look good. 

/lgtm

PR #298, opened 2026-08-13
feat: publish standard resource.kubernetes.io/numaNode attribute

### LINKED ISSUE #219: Publish standard resource.kubernetes.io/numaNode attribute

ffromani:
#### What type of PR is this?

/kind feature

After we rebased on top of kubernetes 1.37, we can now expose the standard kubernetes NUMANode attribute,
keeping our legacy ones as deprecated backward compatibility. We will remove around the 0.4.0/0.5.0 timeframe.

Fixes: #219 

N/A

#### Release note

```release-note
The driver now publishes the standard NUMA node ID attribute, as available since kubernetes 1.37
```

#### Documentation updates

N/A

### DIFF
--- a/pkg/device/attributes.go
+++ b/pkg/device/attributes.go
@@ -37,7 +37,6 @@ const (
 )
 
 const (
-	AttributeNUMANodeID resourceapi.QualifiedName = "dra.cpu/numaNodeID"
 	AttributeSocketID   resourceapi.QualifiedName = "dra.cpu/socketID"
 	AttributeSMTEnabled resourceapi.QualifiedName = "dra.cpu/smtEnabled"
 	AttributeCacheL3ID  resourceapi.QualifiedName = "dra.cpu/cacheL3ID"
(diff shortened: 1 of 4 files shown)


pravk03:
I wonder if we want to deprecate the old attributes sooner. It might get more disruptive with time, considering containers can start relying on this since this is plumbed through downward API now. 

Non-blocking for this PR. We can reconsider before 0.3.0.

pravk03:
Thanks a lot. 

/lgtm
/approve

ffromani:
> I wonder if we want to deprecate the old attributes sooner. It might get more disruptive with time, considering containers can start relying on this since this is plumbed through downward API now.
> 
> Non-blocking for this PR. We can reconsider before 0.3.0.

absolutely, filing a tracker issue and adding to the 0.3.0 tracker: https://github.com/kubernetes-sigs/dra-driver-cpu/issues/299

PR #300, opened 2026-08-13
cleanup: address follow-up review comments from PR #267

### LINKED ISSUE #296: docs: stop inlining tunable references in the README.md, link to a stable doc

Currently, we try to be terse but still comprehensive in our docs and we add examples of config-only tunables in the `config` section of the README.md. We should just add a generic reference and defer to a stable less contrained (spacewise) user documentation

**Why is this needed**:
Doesn't scale! as we (hopefully) gain more important tunables, listing them BOTH tersely but in a useful manner quickly hits the limits. Xref: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/267/commits#diff-d24b7543e12af34beeff0b6c105e655c1d8576c57b7dd1ee6e7995a3e3b7290dR44

pravk03:
#### What type of PR is this?
/kind cleanup

Address follow up comments from https://github.com/kubernetes-sigs/dra-driver-cpu/pull/267
- Removes inline config examples in README and values schema
- Updates e2e tests to use pod matchers to check status

Fixes https://github.com/kubernetes-sigs/dra-driver-cpu/issues/296

I have restricted the refactoring to the NodeAllocatableClaimStatus matcher. I think more e2e assertions could be updated to follow the same pattern. We can target a wider e2e test refactor in a separate PR.

#### Release note

```release-note

```


pravk03:
xref: https://github.com/kubernetes/test-infra/pull/37682 

The e2e lanes now set the node allocatable mapping explicitly.

Makefile
```diff
@@ -115,8 +115,8 @@ DRACPU_E2E_RESERVED_CPUS ?= 0
 # publishNodeAllocatableResourceMapping driverConfig option.
 # Also requires a cluster (KIND_K8S_VERSION node image) with the
 # DRANodeAllocatableResources feature gate enabled.
-DRACPU_E2E_NODE_ALLOCATABLE ?= true
-HELM_NODE_ALLOCATABLE_ARGS := $(if $(filter true,$(DRACPU_E2E_NODE_ALLOCATABLE)),--set driverConfig.publishNodeAllocatableResourceMapping=true,)
+DRACPU_E2E_NODE_ALLOCATABLE_MAPPING ?= false
+HELM_NODE_ALLOCATABLE_ARGS := $(if $(filter true,$(DRACPU_E2E_NODE_ALLOCATABLE_MAPPING)),--set driverConfig.publishNodeAllocatableResourceMapping=true,)
 # Extra arguments passed to golangci-lint in the lint target.
```
fmuyassarov, line 119: I would simplify this and just rely on `DRACPU_E2E_NODE_ALLOCATABLE_MAPPING` without intermediate var.

and then here https://github.com/pravk03/dra-driver-cpu/blob/3b5f6b64ce436ca7d94663b86ef7431e23189f17/Makefile#L216 we could just use


```shell
--set args.exposePCIeRoots=true \
--set driverConfig.publishNodeAllocatableResourceMapping=$(DRACPU_E2E_NODE_ALLOCATABLE_MAPPING)
```

pravk03: This is simpler. Thanks. Done.

fmuyassarov: @pravk03 can we squash the last commit to the prior ones just to keep the git history clean ?
and then lets merge it

pravk03: Sorry, my bad. Done.

Issue #299, opened 2026-08-13
deprecate legacy DRA attributes sooner


ffromani:
The standardization of the `numaNode` attribute (see #219 ) brought the question: should we deprecate/remove the existing driver-specific attributes earlier than 0.4.0?

ffromani:
xref: https://github.com/kubernetes-sigs/dra-driver-cpu/pull/298#issuecomment-5283839217

alanhuangch:
I'd like to take a small scoped follow-up for this issue:

- update user-facing examples to prefer the standard `resource.kubernetes.io/numaNode` attribute;
- document `dra.cpu/numaNodeID` and `dra.net/numaNode` as deprecated compatibility attributes;
- extend the metadata e2e test to assert the standard attribute;
- keep publishing and testing the legacy attributes for now.

This would not change the removal timeline or delete any attributes. Would this scope be welcome?

ffromani:
> I'd like to take a small scoped follow-up for this issue:
> 
>     * update user-facing examples to prefer the standard `resource.kubernetes.io/numaNode` attribute;
> 
>     * document `dra.cpu/numaNodeID` and `dra.net/numaNode` as deprecated compatibility attributes;
> 
>     * extend the metadata e2e test to assert the standard attribute;
> 
>     * keep publishing and testing the legacy attributes for now.
> 
> 
> This would not change the removal timeline or delete any attributes. Would this scope be welcome?

Thanks for the contribution. I think we end up doing pretty much the same as side effect of #316 
All in all, we converged on waiting 0.4.0 for deprecations.

PR #301, opened 2026-08-18
docs: prefer standard NUMA device attribute

alanhuangch:
## Summary

- Update user-facing examples to prefer resource.kubernetes.io/numaNode.
- Document dra.cpu/numaNodeID and dra.net/numaNode as deprecated compatibility attributes.
- Extend the metadata e2e test to assert the standard NUMA attribute.
- Keep publishing and testing the legacy attributes for compatibility.

This does not remove any attributes or change the removal timeline.

## Testing

- make lint
- go test ./pkg/device ./pkg/driver
- go test -run ^$ ./test/e2e
- go vet ./...

Related to #299

### DIFF
--- a/docs/user/device-attributes.md
+++ b/docs/user/device-attributes.md
@@ -19,11 +19,12 @@ CPU group, `individual` one device per CPU.
 
 | Attribute                         | Type    | Description                                                                                                    |
 | --------------------------------- | ------- | -------------------------------------------------------------------------------------------------------------- |
-| `dra.cpu/numaNodeID`              | int     | NUMA node of the group (published when grouping by NUMA node)                                                  |
+| `resource.kubernetes.io/numaNode` | int     | Standard NUMA node of the group (published when grouping by NUMA node)                                         |
+| `dra.cpu/numaNodeID`              | int     | Legacy driver-specific NUMA node attribute (NUMA grouping)                                                     |
 | `dra.cpu/socketID`                | int     | CPU socket of the group (published when grouping by NUMA node or socket)                                       |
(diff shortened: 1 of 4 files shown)


ffromani (commented):
/ok-to-test

docs/user/device-attributes.md
```diff
@@ -19,11 +19,12 @@ CPU group, `individual` one device per CPU.
 
 | Attribute                         | Type    | Description                                                                                                    |
 | --------------------------------- | ------- | -------------------------------------------------------------------------------------------------------------- |
-| `dra.cpu/numaNodeID`              | int     | NUMA node of the group (published when grouping by NUMA node)                                                  |
+| `resource.kubernetes.io/numaNode` | int     | Standard NUMA node of the group (published when grouping by NUMA node)                                         |
+| `dra.cpu/numaNodeID`              | int     | Legacy driver-specific NUMA node attribute (NUMA grouping)                                                     |
```
ffromani: we should probably (in a followup PR likely) split this table in "currently supported" table and "legacy, deprecated, removed in a future version" table

ffromani (commented):
@alanhuangch thanks for this PR. If you can add a followup commit to split the tables per inline comment that would be great. Otherwise we can do in a followup PR.

docs/user/device-attributes.md
alanhuangch: Done in 8fb270f. 

I split the tables for both grouped and individual modes into currently supported and legacy/deprecated attributes, and clarified that the compatibility attributes will be removed in a future version. 

Thanks for the suggestion!

ffromani (commented):
/approve
/lgtm

thanks for the updates

PR #309, opened 2026-08-26
helm: reject unknown top-level values

### LINKED ISSUE #308: Helm chart accepts unknown top-level values; needs a migration-safe policy
Noticed while reviewing #282. Reproduced on main at f8753af with Helm v4.2.3.

`values.schema.json` sets `additionalProperties: false` on every nested object it describes, but not at the top level, so an unknown top-level key is accepted and ignored:

```console
$ helm template test deployment/helm/dra-driver-cpu --set-string args.bogusKey=x
Error: values don't meet the specifications of the schema(s) in the following chart(s):
dra-driver-cpu:
- at '/args': additional properties 'bogusKey' not allowed

$ helm template test deployment/helm/dra-driver-cpu --set-string bogusTopLevelKey=x
---
# Source: dra-driver-cpu/templates/daemonset.yaml
apiVersion: apps/v1
...
```

The nested objects that already refuse unknown keys are `args`, `image`, `rbac`, `resources`, `resources.requests`, and the generated `driverConfig` schema. The top level is the only one left open, and that looks less like a decision than like the one place nobody had to write it down.

I would like it to have a decided policy: the same `additionalProperties: false`, an allowlist that keeps named legacy keys accepted, or a documented decision to leave it open.

**Why is this needed**:

A misspelled top-level key reads as "unset". For `kubeletRootDir` that means the chart renders the default while the operator believes the kubelet was relocated:

```console
$ helm template test deployment/helm/dra-driver-cpu \
    --set-string kubeletRootDirectory=/var/lib/custom-kubelet
...
      - hostPath:
          path: /var/lib/kubelet/plugins_registry
          type: DirectoryOrCreate
        name: plugin-registry
```

That is #231's failure mode reached another way: the driver registers under a root the kubelet is not watching. #282 adds a template guard for the one spelling Helm cannot distinguish, a key differing from `kubeletRootDir` only in case, but that guard cannot see `kubeletRootDirectory` or `kubelet-root-dir`, and a per-value guard does not generalize to the other 18 top-level keys.

What makes this a policy question rather than a one-line schema change is what the strict version does to values carried forward:

```console
$ cat carried-forward.yaml
someKeyThisChartNoLongerHas: true

$ helm template test ./chart-with-additionalProperties-false -f carried-forward.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
dra-driver-cpu:
- at '': additional properties 'someKeyThisChartNoLongerHas' not allowed
```

`helm upgrade --reuse-values` merges the values a release was installed with, so a key this chart has since dropped would fail validation on upgrade instead of being ignored as it is today. Keys do move here: `args.cpuDeviceMode` is already marked deprecated in `values.yaml` in favor of `driverConfig.cpuDeviceMode`. Keys a user added for their own tooling are in the same position.

So the work is deciding what happens to those, not the one line.

/kind feature

Filed with AI assistance. The output above is from real runs, not reconstructed.

thc1006:
#### What type of PR is this?

/kind feature

`args`, `image`, `rbac`, `resources` and the bundled `driverConfig` schema already refuse unknown keys. The values root was the one object left open, so a top-level value the chart never had was accepted and ignored, and the chart rendered as if it were unset. For `kubeletRootDir` that means the kubelet paths render at the default while the operator believes they were moved, which is #231 reached by a different route.

This is the Helm-side counterpart of what #213 did for the config file, where `applyMap` was changed to `DisallowUnknownFields()` so a misspelled key fails startup instead of silently falling back to defaults. @fmuyassarov noted there that Helm schema validation was the intended place to catch the same mistake at install time.

Fixes #308

**Root only, not `--no-additional-properties`.** That option closes every object that does not say otherwise. It would close the maps and item shapes whose keys come from Kubernetes rather than from this chart, among them `podLabels`, `podAnnotations`, `serviceAccount.annotations`, `nodeSelector`, `affinity`, `tolerations`, `resources.limits`, `extraVolumes` and `extraVolumeMounts`. It would also put a closed wrapper around the `driverConfig` `$ref`, whose real contract lives in the bundled definition. I generated both and diffed them before choosing.

**`global` needs no hand-written entry.** Helm gives a dependency its parent's values under that name, which a closed root would reject. The pinned generator adds the property itself once the root disallows additional properties, so Helm's reserved `global` table keeps working. The nine lines in `values.schema.json` are that plus the root setting; nothing in the schema is edited by hand.

That is the `global` table specifically, not every way a parent can wrap a chart. A parent using `condition: dra-driver-cpu.enabled` puts an `enabled` key in the subchart's values, and a closed root refuses it. Whether to expose `enabled` as a chart value is a separate decision, so this PR does not make one.

**No values are removed here.** `args.cpuDeviceMode`, `args.groupBy`, `args.reservedCPUs` and `args.hostnameOverride` stay valid with their current precedence.

**Dropped since the first revision.** An earlier version added a shell check under `hack/ci` that rendered the chart for a handful of values and asserted which ones the schema refuses. @fmuyassarov pointed out that matching on stderr text and juggling temp files is awkward to keep alive, and that helm-unittest is where those cases belong. It is gone, and with it the workflow trigger it needed, which leaves this PR at the generator flag, the regenerated schema and the documentation. I would rather put the cases back under helm-unittest than carry a second shape of them.

**Overlap with #282.** Both touch the `Makefile`, the generated `values.schema.json` and `docs/user/configuration.md`. I merged them locally: no conflicts, both sets of changes survive, the merged schema comes out identical to a fresh `make helm-schema`, and the chart checks pass on the merged tree.

#### Release note

```release-note
The Helm chart now rejects unknown top-level values. A misspelled value name fails validation before install or upgrade instead of being silently ignored. Existing deprecated chart values stay accepted, and Helm's reserved `global` values remain available for dependency use.
```

#### Documentation updates

- `docs/user/configuration.md` describes what the root now rejects, and what to do when an upgrade still carries a value the chart has since removed.

### DIFF
--- a/docs/user/configuration.md
+++ b/docs/user/configuration.md
@@ -141,6 +141,21 @@ hand exposes individual fields as explicit Helm values.
   single configuration mechanism. The driver logs the effective configuration at startup so you
   can verify which values are active.
 
+### Helm values validation and compatibility
+
+The chart rejects unknown top-level values, so a misspelled value name fails validation before
+anything is rendered rather than being ignored. Helm's reserved `global` table stays accepted,
+and values whose keys you pick, such as `podLabels` and `resources.limits`, are untouched.
+
+If a chart value is removed in a later release, an upgrade that carries the previous release's
+values forward can fail validation. Helm carries them when no new values are supplied at all,
+with `--reuse-values`, and with `--reset-then-reuse-values`. Validation runs before anything is
+rendered or applied, so the running release is untouched: drop the stale key and upgrade again,
+or pass `--reset-values` with the values you want in full.
+
+Values that only your own tooling reads belong outside the file you hand to Helm. They were
+never chart values, so they are rejected rather than ignored.
+
 ### Command-line flags
 
 **NOTE:** Command-line flags are kept mainly for backward compatibility. Prefer the
--- a/Makefile
+++ b/Makefile
@@ -276,6 +276,7 @@ helm-schema: driverconfig-schema ## regenerate values.schema.json from values.ya
 	go run github.com/losisin/helm-values-schema-json/v2@v$(HELM_SCHEMA_VERSION) \
 		-f ${HELM_CHART}/values.yaml \
 		-o ${HELM_CHART}/values.schema.json \
+		--schema-root.additional-properties=false \
 		--use-helm-docs \
 		--bundle \
 		--bundle-root ${HELM_CHART} \
--- a/deployment/helm/dra-driver-cpu/values.schema.json
+++ b/deployment/helm/dra-driver-cpu/values.schema.json
@@ -76,6 +76,14 @@
       "description": "Override the full release name",
       "type": "string"
     },
+    "global": {
+      "description": "Global values shared between all subcharts",
+      "$comment": "Added automatically by 'helm schema' to allow this chart to be used as a Helm dependency, as the `additionalProperties` setting would otherwise collide with Helm's special 'global' values key.",
+      "type": [
+        "object",
+        "null"
+      ]
+    },
     "healthzPath": {
       "description": "Path for liveness and readiness probes",
       "type": "string"
@@ -210,6 +218,7 @@
       }
     }
   },
+  "additionalProperties": false,
   "$defs": {
     "driverconfig.schema.json": {
       "$schema": "https://json-schema.org/draft/2020-12/schema",


.github/workflows/helm-lint.yaml
```diff
@@ -6,6 +6,12 @@ on:
       - 'deployment/helm/**'
       - 'internal/driverconfig/**'
       - 'tools/gen-driverconfig-schema/**'
+      # The steps below run 'make', which pins the generator, decides whether
+      # the check runs, and builds the driverConfig generator from the module.
+      - 'hack/ci/helm-values-schema-check.sh'
+      - 'go.mod'
+      - 'go.sum'
+      - 'Makefile'
```
fmuyassarov: Why should these trigger the Helm linter?

docs/user/configuration.md
```diff
@@ -141,6 +141,26 @@ hand exposes individual fields as explicit Helm values.
...
+notes, normally for one or two minor releases.
+
+Once a value is gone, an upgrade carrying the previous release's values forward can fail
+validation. Helm carries them when no new values are supplied at all, with `--reuse-values`,
+and with `--reset-then-reuse-values`. Validation runs before anything is rendered or applied,
+so the running release is untouched: drop the stale key and upgrade again, or pass
+`--reset-values` with the values you want in full.
+
+Values that only your own tooling reads belong outside the file you hand to Helm. They were
+never chart values, so they are refused outright rather than given a deprecation window.
```
fmuyassarov: I see the intention behind adding this, which is good. I’m just not sure how much value it provides to the user in this particular place. 

For example, this part makes sense, but I’d probably leave it out here since it feels more like a note/contract for developers rather than something the user needs to see.
```
Chart values are deprecated before they are removed, and stay valid meanwhile. The deprecation
is called out in the values documentation, the generated schema description and the release
notes, normally for one or two minor releases.
```

fmuyassarov (commented):
Thanks @thc1006. One thought is that hack/ci/helm-values-schema-check.sh is a new shell scripton the CI and patterns like matching on stderr text and using temp files/traps can be a little fragile to evolve over time. Should we rather think of expressing these cases in go test ? I have a plan for some time now to build it with helm-unittest but didn't get a chance yet. So, if we could reduce the scope of the PR to `additionalProperties` as the title says, we should avoid the possible helm install situations we discussed on the issue

.github/workflows/helm-lint.yaml
thc1006: They shouldn't. The block was there for the check script, which is gone, so this file is back to exactly what main has. You asked the same thing on #274 and you were right then too.

docs/user/configuration.md
thc1006: Agreed, that paragraph was a note to ourselves. Removed. What is left is what the root rejects and what to do when an upgrade still carries a value the chart has since removed, which is the part a user hits.

thc1006:
Dropped it. Matching on stderr text and carrying temp files around is a poor thing to keep alive, and I would rather wait for helm-unittest than leave a second shape of it in the tree.

On Go: what those cases exercise is Helm validating values against the schema, so a Go test would still have to shell out to helm or reimplement that validation. helm-unittest is the right home. When you get to it I am glad to bring them back there, or to send them separately against it.

So this is now the generator flag, the regenerated schema and the documentation, three files and 25 lines. The workflow is back to what main has, and the deprecation note has come out of the user doc.

One thing worth recording, since the coverage went with the script: `helm-schema-check` only diffs the generated file, so regenerating with a different flag and committing the result still passes. The schema is small and the diff is readable, so that shows up in review rather than silently, but it is the gap those cases were closing.

I also merged this with #282 locally to check they do not collide: no conflicts, and the merged schema comes out identical to a fresh `make helm-schema`.

fmuyassarov (commented):
/lgtm
thank you
/cc @ffromani

pravk03:
/approve

Thanks!

PR #312, opened 2026-08-27
Fix machine grouped CPU capacity over-reporting

### LINKED ISSUE #311: Avoid over-reporting CPU capacity in machine grouped mode
## Problem

When topology discovery skips CPUs with incomplete topology information, the driver still publishes those CPUs in `groupBy=machine` mode.

`GetCPUInfos()` omits CPUs whose topology cannot be read, and the allocation store is built from the filtered `CPUDetails`. However, `groupedCPUDeviceInfos()` builds the machine device from the original `onlineCPUs` set. This can make the ResourceSlice report CPU capacity that the driver cannot actually allocate.

## Expected behavior

ResourceSlice capacity and allocation store should use the same validated CPU set. In machine grouped mode, CPUs skipped during topology discovery must not be advertised as allocatable.

AutuSnow:
Fixes #311

## What changed

Use the topology-validated `CPUDetails` CPU set when building the machine-grouped device. This keeps ResourceSlice capacity aligned with the allocation store when topology discovery skips an online CPU.

Add a regression test covering an online CPU missing from `CPUDetails`.

### DIFF
--- a/pkg/device/builder.go
+++ b/pkg/device/builder.go
@@ -136,7 +136,10 @@ func groupedCPUDeviceInfos(groupBy string, topo *cpuinfo.CPUTopology, onlineCPUs
 			})
 		}
 	case GROUP_BY_MACHINE:
-		allocatableCPUs := onlineCPUs.Difference(reservedCPUs)
+		// Use the topology-validated CPU set. GetCPUInfos filters online CPUs
+		// whose topology is incomplete, and the allocation store is built from
+		// the same CPUDetails map.
+		allocatableCPUs := topo.CPUDetails.CPUs().Difference(reservedCPUs)
 		devices = append(devices, groupedCPUDeviceInfo{
 			name: CPUDeviceMachineGrouped,
 			cpus: allocatableCPUs,
--- a/pkg/device/builder_test.go
+++ b/pkg/device/builder_test.go
@@ -136,3 +136,22 @@ func TestDeviceBuilderNodeAllocatableResourceMapping(t *testing.T) {
 		})
 	}
 }
+
+func TestMachineGroupedUsesTopologyValidatedCPUs(t *testing.T) {
+	topo := fakeTopology()
+	// CPU 4 is online but was omitted from CPUDetails because topology
+	// discovery could not validate it.
+	online := cpuset.New(0, 1, 2, 3, 4)
+
+	devices, _ := device.BuildGrouped(
+		logr.Discard(), device.GROUP_BY_MACHINE, topo, online, cpuset.New(),
+		store.NewPCIeRootMapper(), false,
+	)
+	require.Len(t, devices, 1)
+
+	capacity := devices[0].Capacity[resourceapi.QualifiedName(device.CPUResourceQualifiedName)]
+	require.Equal(t, int64(4), capacity.Value.Value())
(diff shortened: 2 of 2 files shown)


pkg/device/builder_test.go
```diff
@@ -136,3 +136,22 @@ func TestDeviceBuilderNodeAllocatableResourceMapping(t *testing.T) {
...
 }
+
+func TestMachineGroupedUsesTopologyValidatedCPUs(t *testing.T) {
+	topo := fakeTopology()
+	// CPU 4 is online but was omitted from CPUDetails because topology
+	// discovery could not validate it.
+	online := cpuset.New(0, 1, 2, 3, 4)
+
+	devices, _ := device.BuildGrouped(
+		logr.Discard(), device.GROUP_BY_MACHINE, topo, online, cpuset.New(),
```
ffromani, line 147: please use `testr.New(t)` in test code

pkg/device/builder.go
```diff
@@ -136,7 +136,10 @@ func groupedCPUDeviceInfos(groupBy string, topo *cpuinfo.CPUTopology, onlineCPUs
 			})
 		}
 	case GROUP_BY_MACHINE:
-		allocatableCPUs := onlineCPUs.Difference(reservedCPUs)
+		// Use the topology-validated CPU set. GetCPUInfos filters online CPUs
+		// whose topology is incomplete, and the allocation store is built from
+		// the same CPUDetails map.
+		allocatableCPUs := topo.CPUDetails.CPUs().Difference(reservedCPUs)
```
ffromani, line 142: while the fix makes sense, it seems the root issue is that we have 2 sources:
1. online CPUs from sysfs
2. the `cpuinfo` package
and they can drift

in which cases can they drift? both data source should ultimately take data from the kernel through sysfs, so it's important to see why we have a drift in the first place

long term (0.4.0 and beyond) we probably want to redesign `cpuinfo` to take into account `onlineCPUs` and integrate properly, and take into account the sub-NUMA allocator needs. Actually, we could do this in stages: simple refactoring incorporating onlineCPUs and rounding some rough eddges in 0.4.0, redesigning the package once the sub-NUMA allocator has a more defined shape.

ffromani (commented):
thanks for the fix. I agree when the drift happens is bad, we should rectify and the proposed solution look simple enough.

But I'd like to deep dive to understand where and why we have a drift in the first place.
Also, please make sure to reflect the purpose of each commit on its commit message: yes github provides context, but the git history should be self-sufficient and independent from github (also considering the recent outage streak).

No need to write essays: few concise lines are sufficient for a good start.

ffromani:
/approve
/lgtm

I prefer merging and improving the nits later. We need to review the online cpu tracking anyway and this is a larger endeavour than what we should do in a PR which fixes a bug

pkg/device/builder.go
ffromani: Forgot to mention that https://github.com/kubernetes-sigs/dra-driver-cpu/issues/311#issuecomment-5436497507 was a good enough answer for me. Internal consistency is important and save pain in the future, even though I still believe (and that was not challenged!) that the chance of actual race is tiny. Especially hotplug/hotunplug is likely to break a lot of assumptions anyway.

Nevertheless, better to have this fix and to remove the other `online CPUs` helper (or to reimplement it explicitly on top of the validated primitive)

Issue #305, opened 2026-08-21
Idea: Mixed CPU allocation via composable exclusive + shared devices


pravk03:
### **Related discussions:**
* https://github.com/kubernetes-sigs/dra-driver-cpu/issues/103 
* https://github.com/kubernetes-sigs/dra-driver-cpu/pull/149 

This would be another approach of achieving the same mixed allocation goal as with https://github.com/kubernetes-sigs/dra-driver-cpu/issues/103, but with an approach that differs from https://github.com/kubernetes-sigs/dra-driver-cpu/pull/149 .

### **Idea:**

With [KEP-5517](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md), every allocation is charged against node allocatable cpu. So the driver can publish the same CPUs in two forms: as exclusive CPUs, and as shared CPUs. Both can be published the full machine size. This looks like too much capacity published through resource slice, but it is safe because the exclusive and the shared parts represent the same underlying node capacity and the scheduler accounts for it correctly.

### Example: 

* Device class 

```yaml
# DeviceClass — exclusive CPUs
apiVersion: resource.k8s.io/v1
kind: DeviceClass
metadata:
  name: dra.cpu
spec:
  selectors:
  - cel:
      expression: >-
        device.driver == "dra.cpu" &&
        device.attributes["dra.cpu"].mode == "exclusive"
---
# DeviceClass — shared CPU time
apiVersion: resource.k8s.io/v1
kind: DeviceClass
metadata:
  name: dra.cpu.shared
spec:
  selectors:
  - cel:
      expression: >-
        device.driver == "dra.cpu" &&
        device.attributes["dra.cpu"].mode == "shared"
```

* ResourceSlice

```yaml
# ResourceSlice — exclusive CPUs (grouped mode, same as today)
devices:
- name: cpudevnuma0
  allowMultipleAllocations: true
  attributes:
    dra.cpu/mode:
      string: "exclusive"
  capacity:
    dra.cpu/cpu:
      value: "16"
  nodeAllocatableResources:
    cpu: {mapping: {capacityKey: dra.cpu/cpu, capacityMultiplier: "1"}}
- name: cpudevnuma1
  allowMultipleAllocations: true
  attributes:
    dra.cpu/mode:
      string: "exclusive"
  capacity:
    dra.cpu/cpu:
      value: "16"
  nodeAllocatableResources:
[...]

PR #313, opened 2026-08-31
docs: document preventing CFS quota throttling with exclusive CPU claims

pravk03:
#### What type of PR is this?
/kind documentation

Document preventing CFS quota throttling with exclusive CPU claims

#### What type of PR is this?
/kind documentation

Document preventing CFS quota throttling with exclusive CPU claims

#### Release note

```release-note

```

#### Documentation updates
- user and developer docs 

#### Release note

```release-note

```

#### Documentation updates
* docs/user/workload-requirements.md
* docs/dev/kubelet-cgroups-and-qos.md

**Note:** AI was used to generate this documentation and it was reviewed by the PR author.

### DIFF
--- a/docs/dev/kubelet-cgroups-and-qos.md
+++ b/docs/dev/kubelet-cgroups-and-qos.md
@@ -1,0 +1,200 @@
+# Deep Dive: Cgroup v2, QoS Classes, and CPU Enforcement in Kubernetes
+
+This document provides a technical deep-dive into how Kubernetes organizes Linux control groups under **cgroup v2**, calculates and applies Quality of Service (QoS) classes, manages cgroup hierarchies at node, QoS, pod, and container levels, and enforces CPU allocations. It also documents the architectural relationship between Kubelet's cgroup management and out-of-tree Dynamic Resource Allocation (DRA) drivers such as `dra-driver-cpu`.
+
+## 1. The Linux Cgroup v2 Hierarchy in Kubernetes
+
+Under cgroup v2 with the `systemd` cgroup driver, Kubernetes organizes container processes into a strictly hierarchical tree under `/sys/fs/cgroup/kubepods.slice`:
+
+```
+/sys/fs/cgroup/
+  └── kubepods.slice (Root Node Allocatable Cgroup)
+        ├── kubepods-burstable.slice (Burstable QoS Slice)
+        │     ├── kubepods-burstable-pod<BURSTABLE_POD_UID_1>.slice (Pod Cgroup)
+        │     │     └── <container-id-1> (Container Cgroup)
+        │     └── kubepods-burstable-pod<BURSTABLE_POD_UID_2>.slice
+        ├── kubepods-besteffort.slice (BestEffort QoS Slice)
+        │     └── kubepods-besteffort-pod<BESTEFFORT_POD_UID_1>.slice (Pod Cgroup)
+        │           └── <container-id-1> (Container Cgroup)
+        ├── kubepods-pod<GUARANTEED_POD_UID_1>.slice (Guaranteed Pod Cgroup)
+        │     └── <container-id-1> (Container Cgroup)
+        └── kubepods-pod<GUARANTEED_POD_UID_2>.slice (Guaranteed Pod Cgroup)
+              └── <container-id-1> (Container Cgroup)
+```
+
+- Top-level QoS cgroups are only created for `Burstable` and `BestEffort`. Guaranteed pods have 1:1 resource reservations directly against the root Node Allocatable cgroup (`kubepods.slice`) and do not share an overcommitted QoS pool. Only `Burstable` and `BestEffort` require intermediate QoS slices so Kubelet can dynamically adjust aggregate CPU shares (`cpu.weight`) and pin BestEffort to the floor of `MinShares = 2` (`cpu.weight = 1`).
+
+In Linux cgroups, resource constraints enforced on an ancestor cgroup are **strictly binding** on all nested descendants:
+
+- CPU execution by a thread in a leaf container charges against that container's cgroup, **and** walks up the tree charging against the parent `pod.slice`, the QoS tier slice (if applicable), and `kubepods.slice`.
+- If a parent cgroup reaches its CFS quota ceiling, the Linux kernel CFS scheduler throttles **all threads in all containers inside that parent cgroup**, regardless of whether the individual container cgroup has unlimited quota (`cpu.max = max`).
+
+## 2. QoS Class Evaluation and Cgroup Settings
+
+Kubernetes defines 3 QoS classes strictly evaluated from standard container specifications in `pod.spec`. DRA claims are ignored during QoS classification.
+
+### 2.1 QoS Classification Rules
+
+Kubernetes evaluates QoS classes strictly from standard resource requirements in `pod.spec` (DRA claims are ignored during QoS classification).
+
+#### Standard Container-Level Resources (without Pod Level Resources)
+
+When Pod Level Resources (`pod.spec.resources`) is not set, Kubelet evaluates requests and limits across all containers (including init containers):
+
+```
+                      +-----------------------------+
+                      |    Are any requests or      |
+                      |   limits set in pod spec?   |
+                      +-----------------------------+
+                                     |
+                         +-----------+-----------+
+                         | No                    | Yes
+                         v                       v
+               +------------------+    +--------------------------------+
+               |    BestEffort    |    |    Does EVERY container set    |
+               +------------------+    |      non-zero req == lim       |
+                                       |    for BOTH cpu and memory?    |
+                                       +--------------------------------+
+                                                     |
+                                         +-----------+-----------+
+                                         | Yes                   | No
+                                         v                       v
+                               +------------------+    +------------------+
+                               |    Guaranteed    |    |    Burstable     |
+                               +------------------+    +------------------+
+```
+
+#### With Pod Level Resources
+
+When the `PodLevelResources` feature gate is enabled, Kubelet checks whether resources are declared in `pod.spec.resources` before falling back to container-level evaluation:
+
+```
+                    +------------------------------------+
+                    |   Are Pod Level Resources set in   |
+                    |        pod.spec.resources?         |
+                    +------------------------------------+
+                                       |
+                          +------------+------------+
+                          | No                      | Yes
+                          v                         v
+           +-----------------------------+   +-----------------------------+
+           | Evaluate Standard Container-|   |   Does pod.spec.resources   |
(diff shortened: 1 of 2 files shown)


pravk03:
@ffromani https://github.com/kubernetes-sigs/dra-driver-cpu/issues/305 got me thinking about CFS quota behavior and our current workload recommendations, and I thought there was room for some improvement.

I also added `docs/dev/kubelet-cgroups-and-qos.md` to provide context on how Kubelet manages cgroups. Some of this touches general kubelet topics, so I'd appreciate thoughts on whether `dra-driver-cpu` is the right place for this deep dive or we could point to any other existing docs.

/assign @ffromani 
/hold

ffromani:
> I also added `docs/dev/kubelet-cgroups-and-qos.md` to provide context on how Kubelet manages cgroups. Some of this touches general kubelet topics, so I'd appreciate thoughts on whether `dra-driver-cpu` is the right place for this deep dive or we could point to any other existing docs.

I think it is. This driver sits at the border of sig-node and DRA and touches system topics. It totally makes sense to have deep dives about these topics, I see no concern here.

We totally may want to bring the findings or proposals stemming from these findings to bigger forums (e.g. wg-device-management, sig-node) later, or to consolidate this information somehow.
In that sense, this driver may not be the best final resting place for these docs; but IMO it is a very good and very legit place to have them. TL;DR: maybe not only there, but definitely also there.

Will review the PR shortly.

ffromani:
/lgtm

the content is sound and the deep dive is helpful. The docs are IMO a clear improvement even if we still have gaps: we solve something, we lose something (specifically, strongest eviction guarantees).

Some extra musings about the general topic.

Tying the QoS class to the exclusive allocation was probably questionable and is not a constraint we necessarily need to carry forward. But that gave the workload an extra guarantee which, from my experience, workload authors now take for granted and that we will need to reintroduce _somehow_: OOM score and eviction protection. I think there's a gap here. Fair, we're deep in research and development phase, but still.

Likewise, the general DPDK-style app requirement holds: a _container_ needs access to both a capped pool and an unthrottled exclusive pool. How we deliver that, is yet to be decided. Doesn't _have_ to be a continuation or an extension of the cpumanager direction. I'm probably stating the obvious, but worth reiterating the concept: cpumanager is _an_ implementation and the only concrete reference, but the goal is the feature set not how we reach the implementation; we may likely need to diverge from that path.

We probably need some minimal form of handoff between the kubelet and the driver to properly integrate with the current invariant:
- pod shape as isolation boundary managed and prepared by the kubelet
- container shape controlled by the driver (replacing the kubelet functionality)
- composability with pod-level-resources (PLR). Some concepts introduced in the KEP 5526 may help perhaps? We added the **internal** concept of isolation level. Perhaps it should bubble up somehow in some form to the API?

I was toying with the idea of using sub-groups in addition to the cgroup layout kubelet sets. Adding subgroups should be fair game and relatively uncontroversial, and can perhaps be a way to implement the mixed CPUs.
I sadly haven't had enough time to properly develop these ideas to form a coherent design, and I dearly hope (and trying to make arrangements) to have time to spend here in the 0.4.0 cycle.

docs/dev/kubelet-cgroups-and-qos.md
```diff
@@ -0,0 +1,140 @@
...
+                      |   limits set in pod spec?   |
+                      +-----------------------------+
+                                     |
+                         +-----------+-----------+
+                         | No                    | Yes
+                         v                       v
+               +------------------+    +--------------------------------+
+               |    BestEffort    |    |  Does EVERY container set BOTH |
+               +------------------+    |  cpu AND memory limits, with   |
+                                       |  requests == limits for each?  |
```
AutuSnow: The QoS decision tree only considers container resources. With PodLevelResources enabled, kubelet derives QoS from pod.spec.resources when those resources are set. Could we add that branch or scope this diagram to pods without pod-level resources?

docs/dev/kubelet-cgroups-and-qos.md
```diff
@@ -0,0 +1,140 @@
...
+| :------------------------------------- | :-------------------- | :--------------------------------------------------------------------------------------------------------------------------- | :---------------------------------------------------------------------------------------------------------- | :------------------------------- |
+| **Hard Ceiling (`cpu.max`)**           | **Container**         | • `unlimited` (static CPU policy and integer requests)<br>• `limits.cpu` (fractional requests or static CPU policy disabled) | • `unlimited` (if `limits.cpu` omitted)<br>• `limits.cpu` (if `limits.cpu` declared)                        | `unlimited`                      |
+|                                        | **Pod (`pod.slice`)** | • `unlimited` (if any container is assigned exclusive CPUs by kubelet)<br>• sum of limits (otherwise)                        | • `unlimited` (if any container omits `limits.cpu`)<br>• sum of limits (if all containers set `limits.cpu`) | `unlimited`                      |
```
AutuSnow: This pod-level cpu.max rule needs a PodLevelResources caveat. A non-zero pod.spec.resources.limits.cpu makes kubelet configure pod CPU quota even when a container omits limits.cpu. Without that caveat, a PLR user may still get the throttling this document is trying to avoid.

docs/dev/kubelet-cgroups-and-qos.md
pravk03: Good point. Added a diagram to cover PLR.

docs/dev/kubelet-cgroups-and-qos.md
pravk03: done.

pravk03:
Re https://github.com/kubernetes-sigs/dra-driver-cpu/pull/313#issuecomment-5491481874 

Big +1 like the sub-cgroups idea, definitely worth more exploration. That said, we’ll likely need to continue work around two limitations:
1. QoS class implicitly derived by Kubelet. Many behaviors in Kubelet are dictated by this, including the top-level cgroup hierarchy. I see this implicit derivation come up as a limitation in multiple contexts (though I’m still looking into all the historical context behind the original decision). Having an option to explicit set/influence QoS class in the Pod spec could open up a lot of improvements both within and outside this project.
2.  Pod-level cgroups are owned by Kubelet, and NRI has to play within whatever envelope is set at that level. I believe accepting it as a constraint is the right approach to work with.

docs/dev/kubelet-cgroups-and-qos.md
```diff
@@ -0,0 +1,199 @@
...
+                    +------------------------------------+
+                    |   Are Pod Level Resources set in   |
+                    |        pod.spec.resources?         |
+                    +------------------------------------+
+                                       |
+                          +------------+------------+
+                          | No                      | Yes
+                          v                         v
+           +-----------------------------+   +-----------------------------+
+           | Evaluate Standard Container-|   | Do BOTH cpu & mem have      |
```
AutuSnow: Could we mention that both CPU and memory request/limit pairs must be present, equal, and non-zero here?
﻿
`ComputePodQOS` treats an absent CPU pair as BestEffort for that resource, so the memory-only PLR example later is Burstable. `req == lim` alone could be read as treating `0 == 0` as Guaranteed.

docs/user/workload-requirements.md
```diff
@@ -98,32 +98,67 @@ The driver then publishes a `nodeAllocatableResources` mapping on every device,
...
-QoS classification logic is an explicit non-goal of the KEP). The assigned class is reported
-in `pod.status.qosClass`.
+When deploying workloads with exclusive CPU claims, the recommended pattern is:
 
-The classification rules below are the standard Pod QoS rules; the table only adds the
-claim-specific configuration on top of them. QoS can be set at either level: when
-**pod-level resources** (`pod.spec.resources`, beta since 1.34) are set, the kubelet
-computes QoS from the pod-level values only; otherwise it is computed from the containers.
+1. **Omit `cpu` requests and limits** on the container referencing the exclusive claim. This prevents the kubelet from setting a CFS bandwidth quota (`cpu.max`), avoiding throttling on dedicated cores.
+1. **Specify `memory` requests and/or limits** matching the container's memory needs. This assigns the pod to the **Burstable** QoS class, protecting it against eviction and out-of-memory (OOM) termination.
```
AutuSnow, line 115: Could we soften this to “better protected than BestEffort”?
﻿
A Burstable pod still participates in kubelet eviction ordering, and its `oom_score_adj` is request-based rather than the fixed value used for Guaranteed pods. It can also still be OOM-killed at its memory limit.
﻿
Calling it “protected against eviction and OOM termination” seems to hide the trade-off Francesco mentioned above. Should we also adjust the wording in the table below?

docs/user/workload-requirements.md
```diff
@@ -98,32 +98,67 @@ The driver then publishes a `nodeAllocatableResources` mapping on every device,
...
+| **Burstable**<br>*(Recommended)* | Set `memory` requests/limits;<br>**omit `cpu` requests and limits** on claim containers. | ✅ **No quota set**    | ✅ **Protected** (OOM score scales with the memory request) |
+| **Guaranteed**                   | Set `requests == limits` for both `cpu` and `memory` on every container.                 | ⚠️ **Quota set**       | ✅ **Most protected** (lowest OOM score, fixed)             |
+| **BestEffort**                   | Omit all `cpu` and `memory` requests and limits in all containers.                       | ✅ **No quota set**    | ❌ **Evicted and OOM-killed first**                         |
+
+- Non-claim containers (such as logging sidecars) can specify standard CPU requests and limits normally; they run on the driver's [shared pool](how-it-works.md).
+- When using **pod-level resources** (`pod.spec.resources`), set pod-level `memory` requests and omit pod-level `limits.cpu`.
+
+#### Preventing CFS Throttling on Exclusive CPUs
+
+A primary motivation for requesting exclusive CPUs is to achieve predictable, zero-jitter execution without CFS quota throttling.
```
AutuSnow: Could we scope this to CFS quota-induced latency?
﻿
Omitting CPU limits removes the `cpu.max` ceiling, but it does not make the workload zero-jitter; interrupts, softirqs, scheduling, and topology effects still remain.
﻿
Would “avoiding CFS quota-induced throttling and latency spikes” describe the guarantee more precisely?

docs/user/workload-requirements.md
pravk03: done

docs/user/workload-requirements.md
pravk03: rephrased

docs/dev/kubelet-cgroups-and-qos.md
pravk03: done

ffromani:
LGTM (intentionally deferring)

AutuSnow:
Thanks ！！ 
/lgtm

Issue #275, opened 2026-07-31
Add tests to measure NRI hook latency


pravk03:
**Why is this needed**:

All NRI hooks (Synchronize, CreateContainer, StopContainer, …) execute under containerd's NRI default timeout. 
containerd skips the plugin if it exceed the timeout. 
We currently have no tests that measure or bound NRI hook latency, so we have no way to tell how much latency a change to these hooks adds.

**What would you like to be added**:

It would help to have some microbenchmarks for NRI hooks with realistic values for number of pods, containers, claims. This gives authors and reviewers a baseline to compare against and an intuition for how a change scales.

ffromani:
another thing we can help is to publish prometheus metrics with histograms about the NRI callback execution time. This complements, not replaces, the NRI hook latency testing which we should have anyway. I can take care of adding these metrics.

PR #314, opened 2026-09-02
metrics: record NRI callbacks duration

ffromani:
#### What type of PR is this?

/kind feature

Spawned of after a concern emerged in https://github.com/kubernetes-sigs/dra-driver-cpu/pull/270#discussion_r3692378266 . We want to have a way to monitor performance in CI (or in general be able to do performance assessment), but this is yet to materialize and a non-trivial effort. A complementary and cheaper improvement is to export metrics about the NRI callback duration, so we can have visibility about how we are doing, and at least some data to work with.
We can probably build the performance measurement/tests on top of these metrics in the future, but this is yet to be decided.

N/A

AI usage disclosure: I use AI to accelerate the initial refactoring, but under close supervision.
I use AI to accelerate the writing of e2e test *utilities*. The e2e flow and expectations were written by a human

#### Release note

```release-note
Add prometheus metrics for NRI callbacks duration
```

#### Documentation updates

- user documentation

### DIFF
--- a/docs/user/metrics.md
+++ b/docs/user/metrics.md
@@ -13,15 +13,16 @@ dracpu introspect metrics
 
 The command prints JSON metadata for custom `dra_cpu_*` metrics only. It does not include default Go runtime, process, or Prometheus client metrics.
 
-| Metric                                   | Type      | Labels   | Description                                                                                |
-| ---------------------------------------- | --------- | -------- | ------------------------------------------------------------------------------------------ |
-| `dra_cpu_allocated_cpus`                 | Gauge     | none     | CPUs currently allocated to prepared resource claims.                                      |
-| `dra_cpu_available_cpus`                 | Gauge     | none     | CPUs still available for allocation after reserved and active claim CPUs are excluded.     |
-| `dra_cpu_reserved_cpus`                  | Gauge     | none     | CPUs excluded from DRA management by driver configuration.                                 |
-| `dra_cpu_resource_claims_active`         | Gauge     | none     | Resource claims currently recorded as active by the allocation store.                      |
-| `dra_cpu_prepare_claims_total`           | Counter   | `result` | Per-claim `PrepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.   |
-| `dra_cpu_unprepare_claims_total`         | Counter   | `result` | Per-claim `UnprepareResourceClaims` results. `result` is `success`, `error`, or `unknown`. |
-| `dra_cpu_prepare_claim_duration_seconds` | Histogram | none     | Per-claim prepare latency in seconds.                                                      |
-| `dra_cpu_claim_allocated_cpus`           | Histogram | none     | CPUs allocated for each newly successful claim allocation.                                 |
+| Metric                                     | Type      | Labels   | Description                                                                                |
+| ------------------------------------------ | --------- | -------- | ------------------------------------------------------------------------------------------ |
+| `dra_cpu_allocated_cpus`                   | Gauge     | none     | CPUs currently allocated to prepared resource claims.                                      |
+| `dra_cpu_available_cpus`                   | Gauge     | none     | CPUs still available for allocation after reserved and active claim CPUs are excluded.     |
+| `dra_cpu_reserved_cpus`                    | Gauge     | none     | CPUs excluded from DRA management by driver configuration.                                 |
+| `dra_cpu_resource_claims_active`           | Gauge     | none     | Resource claims currently recorded as active by the allocation store.                      |
+| `dra_cpu_prepare_claims_total`             | Counter   | `result` | Per-claim `PrepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.   |
+| `dra_cpu_unprepare_claims_total`           | Counter   | `result` | Per-claim `UnprepareResourceClaims` results. `result` is `success`, `error`, or `unknown`. |
+| `dra_cpu_prepare_claim_duration_seconds`   | Histogram | none     | Per-claim prepare latency in seconds.                                                      |
+| `dra_cpu_unprepare_claim_duration_seconds` | Histogram | none     | Per-claim unprepare latency in seconds.                                                    |
+| `dra_cpu_claim_allocated_cpus`             | Histogram | none     | CPUs allocated for each newly successful claim allocation.                                 |
 
 The custom metrics intentionally avoid labels for namespace, pod, claim, device, node, socket, NUMA node, group mode, and error reason. Those labels would either be high-cardinality or need more API design before becoming part of the driver's metric surface. Node identity should come from scrape target labels.
--- a/docs/user/metrics.md
+++ b/docs/user/metrics.md
@@ -13,16 +13,30 @@ dracpu introspect metrics
 
 The command prints JSON metadata for custom `dra_cpu_*` metrics only. It does not include default Go runtime, process, or Prometheus client metrics.
 
-| Metric                                     | Type      | Labels   | Description                                                                                |
-| ------------------------------------------ | --------- | -------- | ------------------------------------------------------------------------------------------ |
-| `dra_cpu_allocated_cpus`                   | Gauge     | none     | CPUs currently allocated to prepared resource claims.                                      |
-| `dra_cpu_available_cpus`                   | Gauge     | none     | CPUs still available for allocation after reserved and active claim CPUs are excluded.     |
-| `dra_cpu_reserved_cpus`                    | Gauge     | none     | CPUs excluded from DRA management by driver configuration.                                 |
-| `dra_cpu_resource_claims_active`           | Gauge     | none     | Resource claims currently recorded as active by the allocation store.                      |
-| `dra_cpu_prepare_claims_total`             | Counter   | `result` | Per-claim `PrepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.   |
-| `dra_cpu_unprepare_claims_total`           | Counter   | `result` | Per-claim `UnprepareResourceClaims` results. `result` is `success`, `error`, or `unknown`. |
-| `dra_cpu_prepare_claim_duration_seconds`   | Histogram | none     | Per-claim prepare latency in seconds.                                                      |
-| `dra_cpu_unprepare_claim_duration_seconds` | Histogram | none     | Per-claim unprepare latency in seconds.                                                    |
-| `dra_cpu_claim_allocated_cpus`             | Histogram | none     | CPUs allocated for each newly successful claim allocation.                                 |
+| Metric                                          | Type      | Labels                          | Description                                                                                                                                    |
+| ----------------------------------------------- | --------- | ------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| `dra_cpu_allocated_cpus`                        | Gauge     | none                            | CPUs currently allocated to prepared resource claims.                                                                                          |
+| `dra_cpu_available_cpus`                        | Gauge     | none                            | CPUs still available for allocation after reserved and active claim CPUs are excluded.                                                         |
+| `dra_cpu_reserved_cpus`                         | Gauge     | none                            | CPUs excluded from DRA management by driver configuration.                                                                                     |
+| `dra_cpu_resource_claims_active`                | Gauge     | none                            | Resource claims currently recorded as active by the allocation store.                                                                          |
+| `dra_cpu_prepare_claims_total`                  | Counter   | `result`                        | Per-claim `PrepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.                                                       |
+| `dra_cpu_unprepare_claims_total`                | Counter   | `result`                        | Per-claim `UnprepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.                                                     |
+| `dra_cpu_prepare_claim_duration_seconds`        | Histogram | none                            | Per-claim prepare latency in seconds.                                                                                                          |
+| `dra_cpu_unprepare_claim_duration_seconds`      | Histogram | none                            | Per-claim unprepare latency in seconds.                                                                                                        |
+| `dra_cpu_claim_allocated_cpus`                  | Histogram | none                            | CPUs allocated for each newly successful claim allocation.                                                                                     |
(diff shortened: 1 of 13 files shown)


docs/user/metrics.md
```diff
@@ -13,15 +13,25 @@ dracpu introspect metrics
...
+| `dra_cpu_resource_claims_active`                | Gauge     | none     | Resource claims currently recorded as active by the allocation store.                        |
+| `dra_cpu_prepare_claims_total`                  | Counter   | `result` | Per-claim `PrepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.     |
+| `dra_cpu_unprepare_claims_total`                | Counter   | `result` | Per-claim `UnprepareResourceClaims` results. `result` is `success`, `error`, or `unknown`.   |
+| `dra_cpu_prepare_claim_duration_seconds`        | Histogram | none     | Per-claim prepare latency in seconds.                                                        |
+| `dra_cpu_unprepare_claim_duration_seconds`      | Histogram | none     | Per-claim unprepare latency in seconds.                                                      |
+| `dra_cpu_claim_allocated_cpus`                  | Histogram | none     | CPUs allocated for each newly successful claim allocation.                                   |
+| `dra_cpu_nri_synchronize_duration_seconds`      | Histogram | `result` | Duration of the NRI `Synchronize` callback in seconds. `result` is `success` or `error`.     |
```
ffromani: this I strongly think it deserves a metric

docs/user/metrics.md
```diff
@@ -13,15 +13,25 @@ dracpu introspect metrics
...
+| `dra_cpu_nri_create_container_duration_seconds` | Histogram | `result` | Duration of the NRI `CreateContainer` callback in seconds. `result` is `success` or `error`. |
+| `dra_cpu_nri_stop_container_duration_seconds`   | Histogram | `result` | Duration of the NRI `StopContainer` callback in seconds. `result` is `success` or `error`.   |
+| `dra_cpu_nri_remove_container_duration_seconds` | Histogram | `result` | Duration of the NRI `RemoveContainer` callback in seconds. `result` is `success` or `error`. |
```
ffromani: but those? added for completeness sake. Unlike DRA, we can have a quite large set of hooks to implement. Worth a metric for each? perhaps a cumulative metric "just in case" for the hooks which we deem less critical?

pkg/driver/nri_hooks.go
```diff
@@ -30,13 +31,15 @@ import (
...
-func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, containers []*api.Container) ([]*api.ContainerUpdate, error) {
+func (cp *CPUDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, containers []*api.Container) (rupdates []*api.ContainerUpdate, rerr error) {
+	startTime := time.Now()
 	_, logger := ctxlog.WithValues(ctx, "opID", generateShortID(opIDLen))
-
 	// this happens once at startup and it's critical enough that we always want to see it.
 	logger.Info("begin: synchronize state with the runtime", "numPods", len(pods), "numContainers", len(containers))
 	defer logger.Info("end: synchronize state with the runtime", "numPods", len(pods), "numContainers", len(containers))
 
+	defer cp.metrics.RecordNRISynchronize(rerr, startTime)
```
AutuSnow: Could we defer this through a closure?
﻿
`rerr` is evaluated when the defer is registered, so it is still nil at this point. If the callback later returns an error, would this still record the sample as `result="success"`?
﻿
The same pattern appears in the other three NRI callbacks. Could we also add a failure-path test for the error label?

AutuSnow (commented):
Thanks for adding these metrics. The direction fits the latency concern raised in #270 and complements the benchmark work tracked in #275. The metric names, bounded result label, and NRI-specific buckets look reasonable.

I found one correctness issue in the deferred metric recording: the current implementation captures the named error return value too early, so failed callbacks are reported as successful. Please fix this in all four NRI callbacks and add at least one failure-path assertion.

The targeted unit and race tests pass. After this fix, I would be happy with the overall design.

ffromani:
thanks for the callout @AutuSnow . Let me invest more time on more and better tests

docs/user/metrics.md
pravk03: +1

docs/user/metrics.md
pravk03: I think create_container deserves a metric as the cost is proportional to number of pods, while the rest of the hooks have a fixed cost. Leaning yes on having separate metric for each of the other hooks.

pkg/driver/nri_hooks.go
```diff
@@ -209,11 +212,15 @@ func (cp *CPUDriver) getSharedContainerUpdates(logger logr.Logger, excludeID typ
...
 // CreateContainer handles container creation requests from the NRI.
-func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
+func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) (radjust *api.ContainerAdjustment, rupdates []*api.ContainerUpdate, rerr error) {
+	startTime := time.Now()
+
 	_, logger := ctxlog.WithValues(ctx, "opID", generateShortID(opIDLen), "pod", ctxlog.KObj(pod), "podUID", pod.Uid, "container", ctr.Name, "containerID", ctr.Id)
 	logger.V(2).Info("begin: CreateContainer")
 	defer logger.V(2).Info("end: CreateContainer")
 
+	defer cp.metrics.RecordNRICreateContainer(rerr, startTime)
```
pravk03: Would it be good to break this down by container type ?. Since guaranteed container's have a higher cost that comes with scanning through all the shared containers. Ok to defer this for later though.

pravk03 (commented):
Thanks a lot for this much-needed PR!

docs/user/metrics.md
ffromani: ok, let's start like this. We're pretty much all slightly (or more) in favor of granular metrics, so let's keep them this way. We declare ourselves and the metrics both alpha-level, so let's take and use some slack.

pkg/driver/nri_hooks.go
ffromani: I'd be fine adding a dimension per container type (exclusively assigned/shared). Leaning toward using a more abstract dimension and not, for example, use QoS Class directly. Would that work?

AutuSnow: That makes sense. Would a simple guaranteed/shared label work here? One small question: should a restarted guaranteed container still be counted as guaranteed, even though it skips the shared-container scan?

ffromani: that's a very good point. My best answer atm is: it should still counted. IOW, it should not lose qualifications. I'm trying to pin to attributes which are derived from these qualifications (= claim count!) and atm this is the best we can do IMO. I guess we will need to restart the convo about how we handle persistence in the next cycle. No shortage of roadmap items!

pkg/driver/nri_hooks.go
ffromani: yes, ended up using a closure. I didn't in the first upload because I grossly overestimated the cost of a closure, which should actually be negligible. Fixed now.

pkg/driver/nri_hooks.go
pravk03: > I'd be fine adding a dimension per container type (exclusively assigned/shared). Leaning toward using a more abstract dimension and not, for example, use QoS Class directly. Would that work?

Yes, that would be better.

Sorry if it adds more work on your plate. I think we may get most of the value by introducing the metric (without breakdown). More granular breakdown can easily be follow up work.

ffromani: it's fine, no worries. updating the PR soon!

pkg/driver/nri_hooks.go
```diff
@@ -285,27 +295,34 @@ func (cp *CPUDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, c
...
-func (cp *CPUDriver) RemoveContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) error {
+func (cp *CPUDriver) RemoveContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) (rerr error) {
+	startTime := time.Now()
+
 	_, logger := ctxlog.WithValues(ctx, "opID", generateShortID(opIDLen), "pod", ctxlog.KObj(pod), "podUID", pod.Uid, "container", ctr.Name, "containerID", ctr.Id)
 	logger.V(2).Info("begin: RemoveContainer")
 	defer logger.V(2).Info("end: RemoveContainer")
 
 	claimUIDs, removed := cp.podConfigStore.RemoveContainerState(types.UID(pod.GetUid()), ctr.GetName(), types.UID(ctr.GetId()))
+	defer func() { cp.metrics.RecordNRIRemoveContainer(rerr, len(claimUIDs), time.Since(startTime)) }()
```
pravk03: Since claim would already be removed from state in StopContainer(), all the metrics would be recorded as "shared" for RemoveContainer(). 

To avoid complexity, we could just the keep the mode breakdown only in CreateContainer()

ffromani: right, good point. Yours is certainly an option, let me think a bit if we can preserve the mode breakdown. 

Thinking aloud, what bothers me is losing signal here and introducing reasonnable and justified asymmetries, but still asymmetries.

Exploring alternatives, and leveraging the fact neither `StopContainer` or `RemoveContainer` cannot actually fail in our current design, would be to update metrics only if `RemoveContainerState` returns true (so actually does the work).

It would look like this:
```
func (cp *CPUDriver) StopContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) ([]*api.ContainerUpdate, error) {
	startTime := time.Now()

	_, logger := ctxlog.WithValues(ctx, "opID", generateShortID(opIDLen), "pod", ctxlog.KObj(pod), "podUID", pod.Uid, "container", ctr.Name, "containerID", ctr.Id)
	logger.V(2).Info("begin: StopContainer")
	defer logger.V(2).Info("end: StopContainer")

	updates := []*api.ContainerUpdate{}
        claimUIDs, removed := cp.podConfigStore.RemoveContainerState(types.UID(pod.GetUid()), ctr.GetName(), types.UID(ctr.GetId()))
	if !removed {
		logger.V(2).Info("ignoring stale or unknown StopContainer event")
		return updates, nil
	}
        cp.metrics.RecordNRIStopContainer(nil, len(claimUIDs), time.Since(startTime))
	return updates, nil
}
```

and

```
func (cp *CPUDriver) RemoveContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) (rerr error) {
	startTime := time.Now()

	_, logger := ctxlog.WithValues(ctx, "opID", generateShortID(opIDLen), "pod", ctxlog.KObj(pod), "podUID", pod.Uid, "container", ctr.Name, "containerID", ctr.Id)
	logger.V(2).Info("begin: RemoveContainer")
	defer logger.V(2).Info("end: RemoveContainer")

	claimUIDs, removed := cp.podConfigStore.RemoveContainerState(types.UID(pod.GetUid()), ctr.GetName(), types.UID(ctr.GetId()))

if !removed {
		logger.V(2).Info("ignoring stale or unknown RemoveContainer event")
		return nil
	}
	if len(claimUIDs) > 0 {
		// this serves only for debugging purposes. We should never get here
		updates, err := cp.getSharedContainerUpdates(logger, types.UID(ctr.GetId()))
		if err != nil {
			logger.Error(err, "unable to calculate shared container updates after RemoveContainer")
		} else {
			logger.Info("RemoveContainer spurious updates needed (unexpected, please file a bug)", "updates", updates)
		}
                cp.metrics.RecordNRIRemoveContainer(err, len(claimUIDs), time.Since(startTime))
	}
	return nil
}
```

this would preserve the signal and (aims to :) ) correctly represent when we do the real work.

WDYT?

pravk03: Thanks! I think that can work too. The metric would then be an error signal indicating StopContainer didn't clean up the state.  The only trade-off is that in the normal path we won't see any samples although we still do some work in checking state.
I am ok with either option

ffromani: ok, fine by me. Implemented a slight variation in RemoveContainer ro make the case `len(claimUIDs)==0` meaningful and not accidental (oversight in my previous sketch).

ffromani:
I went and added e2e tests. While we should prefer unit tests, I don't want us to neglect e2e tests coverage and we should, I believe, strive to keep _some_ e2e test coverage. From my side the PR is complete now.

docs/user/metrics.md
```diff
@@ -13,15 +13,25 @@ dracpu introspect metrics
...
+| `dra_cpu_nri_synchronize_duration_seconds`      | Histogram | `result`                        | Duration of the NRI `Synchronize` callback in seconds. `result` is `success` or `error`.                                                       |
+| `dra_cpu_nri_create_container_duration_seconds` | Histogram | `result`, `cpu_allocation_mode` | Duration of the NRI `CreateContainer` callback in seconds. `result` is `success` or `error`; `cpu_allocation_mode` is `shared` or `exclusive`. |
+| `dra_cpu_nri_stop_container_duration_seconds`   | Histogram | `result`, `cpu_allocation_mode` | Duration of the NRI `StopContainer` callback in seconds. `result` is `success` or `error`; `cpu_allocation_mode` is `shared` or `exclusive`.   |
+| `dra_cpu_nri_remove_container_duration_seconds` | Histogram | `result`, `cpu_allocation_mode` | Duration of the NRI `RemoveContainer` callback in seconds. `result` is `success` or `error`; `cpu_allocation_mode` is `shared` or `exclusive`. |
```
pravk03, line 30: nit: Can we document that this is fired only when StopContainer() failed to clear the state.

pravk03:
Thanks a lot for the changes. Just one minor nit from me the rest look great. 

/lgtm
/hold

ffromani:
/hold cancel

because LGTM was gone due to the update

docs/user/metrics.md
ffromani: sure thing, added

pravk03:
Thanks. 
(Since @AutuSnow's previous comments have been addressed, taking the liberty to LGTM and get this merged)

/lgtm

Issue #304, opened 2026-08-21
umbrella issue: external scheduler support in 0.3.0


ffromani:
Track the work needed to support the external schedulers using machine mode and grouped device mode.
We track the enhancements and the cleanups we do in the process

This issue will be closed once the linked PRs are either merged or postponed and the next steps are clear.
 
- #303 clarifies the allocation behavior with count > 1. It's a behavior we always had (correctly!) but was not explicitly documented nor tested. This PR rectifies so.
- #229 add support for external allocator explicitly, in all grouped mode and streamlines machine mode, which is no longer a big exception now
- #316 (originally in #229) reports CPUIDs of the grouped devices. Will incorporate portions of #217; also address most of machine mode parity; further gaps, if any, will be fixed during the 0.4.0 cycle.

ffromani:
/cc @pravk03 @killianmuldoon 

no new content here, just better tracking

pravk03:
Wondering if we can arrive at a decision on https://github.com/kubernetes-sigs/dra-driver-cpu/issues/164 (also related issues - https://github.com/kubernetes-sigs/dra-driver-cpu/issues/152, https://github.com/kubernetes-sigs/dra-driver-cpu/issues/112) with https://github.com/kubernetes-sigs/dra-driver-cpu/pull/229 validated and merged ?. If yes, we can mark the individual mode as deprecated (at least in docs) in `0.3.0`.

ffromani:
@pravk03 yes, that would be my main goal and my main driving force here. To the point that IMO is worth delaying 0.3.0 a couple weeks or so (but not much more: KEP 5517 compatibility is huge and needs to ship, plus the other truckload of quiet QoL improvements made) to reach consensus

pravk03:
Thanks @ffromani. Just want to confirm. 

> To the point that IMO is worth delaying 0.3.0 a couple weeks or so

For sure, and no rush. We would definitely want this in 0.3.0. Since we are also validating with the slurm usecase with @killianmuldoon's help, it should give us good confidence to make the decision on individual mode.

ffromani:
all PRs merged. Closing. 
Please reopen if there's any gap and we will plan accordingly

Issue #290, opened 2026-08-10
Recommended way to allocate more than 32 CPUs in individual mode?


JiaHanDong:
Hi maintainers,

We are using `dra-driver-cpu` in **individual mode** and have workloads that need to allocate more than 32 CPU cores to a single Pod, for example 64 or more CPUs.

In individual mode, each CPU is represented as an individual DRA device. Therefore, allocating 64 CPUs appears to produce 64 entries in `status.allocation.devices.results` .

However, Kubernetes currently defines the following limit:

```go
const AllocationResultsMaxSize = 32
```

This is defined in `pkg/apis/resource/types.go` and limits the number of entries in `allocation.devices.results` for one ResourceClaim.

We would like to understand the recommended way to support workloads requiring more than 32 CPUs in individual mode.

Would splitting the allocation across multiple ResourceClaim objects—for example, two claims requesting 32 CPUs each—and attaching both claims to the same Pod/container be the recommended approach?

Alternatively, are there any other recommended methods or mechanisms for allocating more than 32 CPUs when using `dra-driver-cpu` in individual mode?

Our current environment is:
Kubernetes: v1.35.4
dra-driver-cpu: v0.2.0
CPU allocation mode: individual
Required allocation size: typically 64 CPUs or more per Pod

We would appreciate guidance on the intended and supported approach for this scenario.

Thank you!

ffromani:
Hi, the only way I'm aware of is indeed split the allocation in multiple claims. While this (likely?) work, it adds nontrivial orchestration complexity, I reckon, to have a robust solution. I'm relatively sure the happy path is easy, but being robust in presence of allocation failures/partial allocation has a high chance to turn out complex.

Since you seem to depend on the individual mode, please be aware that for all the above reasons and more, we would like to gradually _deprecate_ individual mode and move towards grouped mode. We want to make more progress in the 0.3.0 cycle (https://github.com/kubernetes-sigs/dra-driver-cpu/pull/229).

Tagging @pravk03 to check if they have more insights on this matter.

pravk03:
[...]

PR #316, opened 2026-09-04
feat: extra attributes when using the external allocators

ffromani:
/kind feature

Exposes DRA resourceslice attributes that external allocators can consume to determine fine placement of workload on cpus. This closes the loop, adding on the existing opaque configuration feature that let the driver enforce the configuration built by an external allocator.
Highlights of this change
- prefactoring to reshape the device builder API both to streamline, make more regular and detect attribute too big or malformed; enabler for the rest of changes
- hard fail if a computed value (e.g. too many CPUs) exceeds the compiled-in DRA attribute size limit.
- introduced SMT sibling layout with a compact representation; versioned the attribute to enable upgrade

#304 
Depends on #229

Introduced a novel "versioned attribute" concept, documented in `docs/user/device-attributes.md`
AI was used, under strict human control and review, to update docs.
AI was used to translate the smt_test.go topologies from the kubernetes representation

#### Release note

```release-note
The driver will now refuse to start if run on a cpu with 3 or more thread siblings. Note the internal representation was already wrong, the failure is now evident.
Expose extra attributes to help external schedulers doing placement decisions
```

#### Documentation updates

user and developer documentation (TBD)


pkg/device/smt.go
```diff
@@ -0,0 +1,169 @@
...
+	"k8s.io/utils/cpuset"
+)
+
+const (
+	AtomSep   = ";"
+	StrideSep = ">"
+	NoSibling = -1
+)
+
+// Compact SMT sibling representation aka `SMTMap`.
```
ffromani, line 36: @killianmuldoon PTAL to this file and to the general concept. The idea here is to report a compact representation of SMT/HyperThread/Core sibling mapping, which should address the last (I hope) missing gap for precise allocation using external allocators (SLURM).

This approach is sustainable for the short term and I'm confortable supporting it, but as we chatted in https://github.com/kubernetes-sigs/dra-driver-cpu/issues/290#issuecomment-5253732695 and https://github.com/kubernetes-sigs/dra-driver-cpu/pull/286#issuecomment-5253994780 the elephant in the room is that **in its current shape** ResourceSlice is ill-suited to represent complex CPU topologies. We will need to either discuss that with DRA architects and possibly extend the API or add a complementary topology representation API.

Again: for the short term and to validate the integration side I'm comfortable shipping and supporting, but this is meant to be transitional until we - as community - tackle the representation issue.
IOW, the driver should be the enforcer, not the topology reporter. It can be _ad interim_ also the component which exposes the topology, but we need to make that role explicit and find a good, generic, extensible representation.

Open to further discussion here; I believe a prototype like this is _useful_ to actually gather experience and let us explore the option, so I intend this approach as necessary exploration, not a hack to unblock things.

cc @pravk03 and @AutuSnow for awareness

ffromani: please note this PR is **NOT** yet in reviewable shape and many tests are still missing. But the core ideas are (I hope) represented in a understandable enough way.

ffromani:
Undrafting because I'm now happy enough with the `device.Build` refactoring. Still quite some work to do, so reviewers be advised: not in reviewable shape until WIP is lifted. The ideas are there though: if ppl wants to have a design/high level review, that's welcome and the PR can sustain that.

ffromani:
also note that if we agree with the general direction, we should merge #229 first and then I can rebase this one and un-WIP

ffromani:
rebased only, still some pending items before I can lift the WIP.
No expected user-visible changes

pkg/device/attributes.go
```diff
@@ -48,6 +48,7 @@ const (
 	// ResourceSlice) that indicates how many CPUs were allocated to a specific
 	// claim from a grouped device's capacity.
 	AttributeAllocatedNumCPUs resourceapi.QualifiedName = "dra.cpu/allocatedNumCPUs"
+	AttributeCPUIDs           resourceapi.QualifiedName = "dra.cpu/cpuIDs"
```
ffromani, line 50: wrong rebase, should go in the following commit

pkg/device/builder.go
```diff
@@ -39,27 +38,130 @@ const (
...
-		nameToID[dev.name] = dev.cpu.CpuID
+// Inventory holds the machine data and resource availability
+// as configured. Reflects a subset of the physical resources
+type Inventory struct {
+	CPUTopology  *cpuinfo.CPUTopology
+	ReservedCPUs cpuset.CPUSet
+}
+
+// OnlineCPUs returns all the online managed CPUs
+func (inv Inventory) OnlineCPUs() cpuset.CPUSet {
```
ffromani: should be `OnlineCPUs` or `ManagedCPUs` (checking the call sites)

ffromani:
@killianmuldoon @pravk03 @AutuSnow the change is now reviewable

pkg/cpuinfo/cpuinfo.go
```diff
@@ -202,6 +203,11 @@ func (s *SystemCPUInfo) GetCPUTopology(logger logr.Logger) (*CPUTopology, error)
 		logger.Info("could not determine SMT status from sysfs, falling back to CPU/Core count", "err", err)
 		smtEnabled = len(cpuInfos) > cores.Len()
 	}
+	smtLevel := 1
+	if smtEnabled {
+		// hardcoded: the CPUInfo struct has only 1 sibling ID, hence max supported SMTLevel is 2
+		smtLevel = 2
```
AutuSnow, line 209: On a >2-way SMT node populateCpuSiblings leaves every SiblingCPUID at -1, so we publish smtLevel=2 alongside an smtMapV1 claiming no CPU has a sibling. Could we derive the level from thread_siblings_list, or hard fail here as we do for the other limits?

I ran a probe test on a topology configured with 2 physical cores and 4 threads. The output showed `smtMap on 4-way SMT = "0-7" (smtEnabled=true smtLevel=2)`, and `DecodeSMTMap` returned `map[0:-1 1:-1 ... 7:-1]`. This presents a contradiction—where `smtLevel=2` and `smtEnabled=true` coexist with an `smtMapV1` indicating that "no CPUs have siblings" on the same device—yet the process completed without any errors. A consumer cannot distinguish, based on the attributes alone, whether this represents the actual topology or a probe failure.

pkg/device/builder.go
```diff
@@ -304,16 +444,57 @@ func createCPUDeviceSlices(deviceInfos []cpuDeviceInfo, pcieRootMapper *store.PC
...
+	}
 	// Note: union semantics are correct because kernel cpulistaffinity currently collapses to NUMA granularity;
 	// grouped allocation at socket/NUMA level therefore covers all CPUs local to every reported root.
 	// See docs/dev/topology-linux-sysfs.md for in-depth exploration about the topic.
 	pcieRoots := pcieRootMapper.GetPCIeRootsForCPU(cpuIDs...)
 	if len(pcieRoots) == 0 {
-		return
+		return nil // nothing to do
+	}
+	if len(pcieRoots) > resourceapi.DeviceAttributeMaxValueLength {
```
AutuSnow, line 461: DeviceAttributeMaxValueLength is the per-string byte limit for StringValues entries, but this compares the number of roots against it, so an over-long root name still slips through while a machine-wide device with many short roots would fail Build.

pkg/cpuinfo/cpuinfo.go
ffromani: right, I need better handling here. We need to hard-fail if SMT level would be > 2: our code can't handle it properly anyway all across the stack.

ffromani: the fix is to drop `smtLevel`: it can't represent anything meaningful on existing real silicon (we have already a sample in the tests!) with mixed P/E core - another thing we support since day 1.
Likewise, `smtEnabled`  will be kept - even though it's easy to be misused
The `smtMapV1` will carry the precise sibling representation.

pkg/device/builder.go
ffromani: good point, checking

docs/dev/smt-mapping.md
```diff
@@ -0,0 +1,29 @@
+# SMT sibling-map encoding
+
+`dra.cpu/smtMap` is a compact representation of logical-CPU sibling pairs.
+
+The encoding is composed by entries separated by `;`.
+Each entry has the generic format `cpuset[>S]` where `cpuset` is the encoding
+of a linux `cpuset`, `>` is the literal `>` sign and `S` is a positive integer
+representing the "stride" aka the offset akacthe absolute difference between 2
```
pravk03: nit: typo

pkg/device/builder.go
```diff
@@ -39,27 +38,128 @@ const (
...
+
+// ManagedCPUs returns CPUs reports the validated online CPUs.
+func (inv Inventory) ManagedCPUs() cpuset.CPUSet {
+	return inv.CPUTopology.CPUDetails.CPUs()
+}
+
+// AllocatableCPUs returns all the machine-level CPUs available for allocation
+// NOTE this DOES NOT take into account currently allocated CPUs: it returns
+// the maximum available theoretical pool depending only on static configuration
+func (inv Inventory) AllocatableCPUs() cpuset.CPUSet {
```
pravk03: nit, optional: Just thinking if AvailableCPUs is a better name for this ? 

" DOES NOT take into account currently allocated CPUs" - this seems counter intuitive for a function called AllocatableCPUs :)

pkg/device/builder.go
```diff
@@ -39,27 +38,128 @@ const (
...
+// device layer selector.
+type Layout string
+
+const (
+	// LayoutIndividual exposes each CPU as a separate device.
+	LayoutIndividual Layout = "individual"
+	// LayoutSocket exposes one device per CPU socket.
+	LayoutSocket Layout = "socket"
+	// LayoutNUMA exposes one device per NUMA node.
+	LayoutNUMA Layout = "numa"
```
pravk03: nit:  We can use "numanode", keeping it consistent with GROUP_BY_NUMA_NODE ?

pkg/device/builder.go
```diff
@@ -39,27 +38,128 @@ const (
...
+	// NameToSocketID is populated for LayoutSocket.
+	NameToSocketID map[string]int
+	// NameToNUMANodeID is populated for LayoutNUMA.
+	NameToNUMANodeID map[string]int
+}
+
+// Layout controls how CPUs are exposed as DRA devices.
+// It abstracts the driver configuration grouping mode and criteria on a single
+// device layer selector.
+type Layout string
```
pravk03: Do we need a string here, since these are internal only ?

docs/dev/smt-layout.md
```diff
@@ -0,0 +1,35 @@
...
+The encoding is composed by entries separated by `;`.
+Each entry has the generic format `cpuset[>S]` where `cpuset` is the encoding
+of a linux `cpuset`, `>` is the literal `>` sign and `S` is a positive integer
+representing the "stride" aka the offset akacthe absolute difference between 2
+sibling CPU IDs.
+
+Examples:
+
+- `0-7>4` means CPUs `0` and `4`, `1` and `5`, `2` and `6`, and `3` and `7`
+  are sibling pairs.
```
pravk03, line 26: Mostly for my understanding, I am wondering how the attribute would like for the below machine shape (which is from a real machine I work on) 
numa node 0 = 0-31,64-95
numa node 1 = 32-63,96-127

Would the attribute look something like - 
numa 0 device: smtMapV1 - "0-31>64"
numa 1 device: smtMapV1 - "32-63>64"

Or, would it contian the whole machine's stride for each numa node 
"0-127>64"  (I think this maybe confusing to the reader since it contains cpu's not in the device's cpuset ?)

Either way, would it be worth capturing an example like this?

pravk03 (commented):
Thanks a lot for this change. Just started reviewing this and had one question about the the smtV1 attribute representation. I plan to spend more time reviewing this PR over the next couple of days.

pkg/device/builder.go
ffromani: sure thing. Updating.

pkg/device/builder.go
ffromani: good point. We probably don't and it would be another nudge this is not an external representation (despite sitting into `pkg`). Updating.

docs/dev/smt-layout.md
ffromani: Assuming grouping by NUMA node, both devices will expose 0-127>64 (assuming 64 is the stride in your example). The map represents the node-wide SMT sibling relationships, not a device-group-specific representation.
Consumers should first reconstruct the sibling map, then intersect it with the device-specific `cpuIDs` attribute to determine the precise topology of CPUs in that device.
This keeps the `smtMapV1` encoding minimal, while leveraging `cpuIDs`, which is already exposed for the external allocator.
In principle, `smtMapV1` could represent all in one go, but I’m hesitant (though open to discussion) to conflate those two meanings in one attribute.
I will improve the docs to better reflect the intended usage.
Finally, `topoDualXeonGold6320` leveraged in the `smt_test.go` tests is I believe almost the example you have. I'm very open to add more tests and add more real-world topologies.
And yes, I'll add examples in the docs!

killianmuldoon:
I've been running this with slurm-bridge for a number of days now and it all seems to be working as expected and is mostly a direct replacement for individual mode. Thanks again for working on this - it solves a number of functional and UX bugs for us.

killianmuldoon (commented):
looks good to me pending open comments

docs/dev/smt-mapping.md
ffromani: thanks, fixed

pkg/device/builder.go
ffromani: this prompted me a deeper review about how we do checking to preemptively check we don't violate the API limits. Updating.

pkg/cpuinfo/cpuinfo.go
```diff
@@ -214,6 +217,32 @@ func (s *SystemCPUInfo) GetCPUTopology(logger logr.Logger) (*CPUTopology, error)
 	}, nil
 }
 
+// validateSMTSiblings verifies that every managed CPU belongs to a core with
+// at most two hardware threads. CPUInfo can represent one sibling only, so
+// wider SMT topologies are unsupported.
+func (s *SystemCPUInfo) validateSMTSiblings(cpuInfos []CPUInfo) error {
+	for _, cpuInfo := range cpuInfos {
+		siblingsPath := path.Join("devices", "system", "cpu", fmt.Sprintf("cpu%d", cpuInfo.CpuID), "topology", "thread_siblings_list")
```
pravk03: I am thinking if we need this extra file read as this could be expensive on larger machines and would affect the driver startup time ?. Can we do this validation in populateCpuSiblings() where we already have sibling ID ? 

If we want this secondary check, one optimization could be to track and visited cpu and sibling and skip the check for the sibling (assume bidirectional uniformity)

pkg/device/builder.go
```diff
@@ -459,3 +467,25 @@ func addCPUIDsAttribute(attrs map[resourceapi.QualifiedName]resourceapi.DeviceAt
...
+		return fmt.Errorf("SMT map unexpectedly empty")
+	}
+	if len(smtMap) > resourceapi.DeviceAttributeMaxValueLength {
+		return fmt.Errorf("SMT map %q cannot be represented within the limit of DRA max value length=%d", smtMap, resourceapi.DeviceAttributeMaxValueLength)
+	}
+	attrs[AttributeSMTMap] = resourceapi.DeviceAttribute{StringValue: new(smtMap)}
+	return nil
+}
+
+func cpuSMTStride(info cpuinfo.CPUInfo) int {
```
pravk03: nit: should this function go in `smt.go` ?

pkg/device/smt.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, info := range topo.CPUDetails {
+		stride := cpuSMTStride(info)
+		cpus, ok := strides[stride]
+		if !ok {
+			cpus = []int{}
+		}
+		if stride == 0 { // prevent NoSibling to leak into the cpuid slice
+			strides[stride] = append(cpus, info.CpuID)
+		} else {
+			strides[stride] = append(cpus, info.CpuID, info.SiblingCPUID)
```
pravk03, line 87: We should also skip when CpuID is already visited (as a SiblingCPUID) before ?

docs/dev/smt-mapping.md
```diff
@@ -1,6 +1,12 @@
 # SMT sibling-map encoding
 
-`dra.cpu/smtMap` is a compact representation of logical-CPU sibling pairs.
+`dra.cpu/smtMapV1` is a compact representation of logical-CPU sibling pairs.
```
pravk03: nit, optional: If we get it right the first time, we would forever be stuck with the V1 name :)
Consider starting with an unversioned name, and we can introduce V2 suffix in the future if we need a breaking format change.

pravk03: nit, optional: Since this is not a key value pair, would smtTopology or smtLayout work better ?

docs/dev/smt-layout.md
pravk03: Thanks a lot  for adding the examples.

pkg/cpuinfo/cpuinfo.go
ffromani: good point, I was lazy :) no need to redo it again, we can check at construction time and fail explicitly if we're reading a topology we can't represent.

pkg/device/builder.go
ffromani: probably yes, moving.

docs/dev/smt-mapping.md
ffromani: good point. Let's start without V1, but I'll document the versioning escape hatch anyway. But the first name we publish will be unversioned.

pkg/device/smt.go
ffromani: IIUC, the concern is that both members of each SMT pair are appended to `strides[stride]`. That duplication is temporary and intentional: `cpuset.New(...)` is needed anyway to produce the compact CPUSet representation, and it deduplicates the IDs before we serialize the SMT map.
I’d prefer to keep the current simple construction rather than add a separate visited-CPU bookkeeping step to avoid these temporary duplicates.
I'm adding add a short comment here to make that intentional behavior clearer. If you think an early skip has a benefit beyond that, please let me know.

docs/dev/smt-mapping.md
ffromani: good point. Going for `smtLayout`.

pkg/device/smt.go
pravk03: SG. Thank you.

pravk03:
/lgtm 

Thanks for addressing the comments. This looks great :)

Issue #203, opened 2026-06-25
Audit atomic state replacement in NRI synchronization paths


AutuSnow:
Follow-up from #193.
﻿
During review of the NRI synchronize fail-closed changes, we noted that rebuilding runtime state into a completed temporary data structure and then replacing the live state atomically is the safer pattern. Incrementally updating live state during synchronization can leave partially rebuilt state visible if an error occurs halfway through.
﻿
This issue tracks auditing the NRI synchronization and related state rebuild paths to make sure this pattern is used consistently.
﻿
Scope:
- Review `Synchronize` and related NRI state reconstruction paths.
- Check whether `podConfigStore`, `claimTracker`, and other shared state are only replaced after successful reconstruction.
- Avoid exposing partially rebuilt state on parse, validation, or runtime errors.
- Add or adjust tests where useful to cover fail-closed behavior.
﻿
This is not a regression from #193; it is follow-up hardening for a future release.

AutuSnow:
I dug into the runtime model a bit more, and I think there is one important concurrency question we should settle before choosing between `atomic.Pointer` and a top-level lock.
﻿
What I verified locally:
﻿
- `kubeletplugin.Start(ctx, cp, ...)` starts the DRA gRPC server handling `PrepareResourceClaims` / `UnprepareResourceClaims`.
- `stub.New(cp, ...)` + `runNRIPluginWithRetry(...)` starts the NRI plugin handling `Synchronize` / `CreateContainer` / `StopContainer`.
- Those are separate serving loops / goroutines.
- `PrepareResourceClaims` and `UnprepareResourceClaims` do read-modify-write operations against `cpuAllocationStore` rather than pure reads.
﻿
That seems important because `atomic.Pointer[*state]` only solves atomic replacement of the state bundle itself: readers won't observe a torn `(podConfig, cpuAllocation, claimTracker)` combination.
﻿
However, it does not seem to solve cross-epoch lost updates if a DRA callback starts from the old state while `Synchronize` swaps in a newly rebuilt state. In that case, a claim allocation written into the old store could effectively disappear from the live state after the swap.
﻿
So the design question I want to put on the table is:
﻿
- Is issue #203 intended to address only fail-closed / atomic replacement during `Synchronize` itself?
- Or should we also protect against cross-subsystem state-epoch races between DRA gRPC callbacks and NRI callbacks?
﻿
[...]

PR #321, opened 2026-09-17
docs: explicitly document the singleton invariant

ffromani:
#### What type of PR is this?

/kind documentation

Another curse-of-knowledge gap: we expect a singleton driver instance per node, but we never mention explictly in the docs.

N/A

Realized thinking about the ownership model while commenting on #203 

#### Release note

```release-note
The driver expects to be a singleton per node and to own all the non-reserved CPUs.
```

#### Documentation updates

- README updates
- upgrade instructions
- install instructions


docs/user/how-it-works.md
```diff
@@ -1,6 +1,14 @@
 # How it Works
 
-The driver is deployed as a DaemonSet running a single executable which implements two core components. The lifecycle of a claim, from pod creation to a pinned container:
+The driver is deployed as a DaemonSet. A single executable implements the DRA
+and NRI interfaces, and it owns the node-local CPU allocation state.
+Exactly one active driver instance is expected to manage each node.
+The driver manages all the available CPUs except those configured as reserved;
+partitioning those CPUs is not supported.
```
AutuSnow: In this sentence the second "those CPUs" reads as pointing back to the reserved CPUs, which is the opposite of the managed pool you mean. Could we spell it out as "the driver-managed CPUs" so the scope stays unambiguous?

docs/user/installation.md
```diff
@@ -23,6 +23,11 @@ The driver is Linux-only and needs node-level privileges (host networking, and N
 CDI directory, and kubelet plugin directory access) — see
 [Security considerations](#security-considerations).
 
+Exactly one active `dra-driver-cpu` instance may manage a node.
```
AutuSnow: This paragraph is repeated in three docs with different modality ("may manage" here, "is expected to manage" in how-it-works.md). Would it work better to state the invariant once and link to it, so the three copies cannot drift apart later?

docs/user/how-it-works.md
ffromani: right, let me rephrase. Good point.

docs/user/installation.md
ffromani: yes, this was bothering me as well. But I'm struggling to find a single document which we can be reasonably sure is evident enough and users will read. Perhaps it's better to start from "how it works" and link back as needed.

pravk03:
/lgtm 

Thanks!

Issue #152, opened 2026-05-20
[BLOCKER][GATED] 0.3.0: switch defaults, deprecate grouped mode


ffromani:
As agreed and planned for 0.2.0 (#57 #112) we need to switch the default to individual mode and to deprecate the grouped mode, hinting to the cpumem driver, before to cut the 0.2.0 release.

Ideally we will have the `cpumem` repo up and running to point users to, but we can use #112 as anchor as well.

himsngh:
@ffromani can I work on this?

ffromani:
postponed to v0.3.0. Decision to implement #112 is now gated by #164 . If #164 enables us to deprecate and remove the individual mode, the split becomes unnecessary.

Issue #112, opened 2026-04-08
Split the driver and retain only the individual mode


ffromani:
We discussed on April 7, 2026 in both the "native resources in DRA" and then in SIG-Node weekly ([slides](https://docs.google.com/presentation/d/18db1pLS4oIZfc10ZbKqMMP3-16IP9HRMOpikDuFaGdA/edit?slide=id.g396ee90418d_0_15#slide=id.g396ee90418d_0_15)).

# The idea

We plan to split the grouped mode support and the memory integration and all the other enhancements we planned into a grouped-mode-only driver, and repurpose and consolidate this driver as individual-mode-only.

Essentially:
keep the existing DRA CPU driver codebase and target individual mode only
- remove the grouped mode
- featurewise, we are already done: polish, optimize, stabilize
- while this is *not* a commitment or a promise, the release 1.0 GA of such a refocused driver is realistic by end of 2026

create a new dra-driver-cpumem (final name TBD) grouped mode only
- remove the individual mode
- merge the memory support
- keep evolving with the ecosystem

# Rationale

We currently have two very different resource representations packed in one driver.
We didn't set a goal about canonical resource representation, and different resource drivers (e.g. vendor-specific CPU drivers) can expose different resource representations. This variety brings flexibility and is almost universally seen as plus. The question is however if a single driver should expose very different resource representations.

Currently, the driver represents resources in “individual mode” and “grouped mode”, which are very different and incompatible representations. These two different modes are very different operational mode stitched together because the initial explorationatory and evolutionary nature of the project. This is unavoidable and actually fine, but we can now stop and review.

Two very different modes bundled together suggest unclear scope, and create unnecessary UX friction: "why should a driver expose so different resource models?" "which one is the best for me?" are very legit and very natural questions.

There's also a timeline consideration: featurewise, the individual mode is “done” bar stabilization.
If we keep the two modes together, it's harder to call GA or to distill clear GA criterias, because the grouped mode dilutes the GA potential; we can call a subset of features GA but this sidesteps the core tension.
In addition, the memory support composes badly with the individual mode, because memory is inherently pooled and anonymous; but the main argument stand even without considering the memory support.

# Timeline

The initial idea is to split *before* 0.2.0 if consensus is reached, to minimize the disruption for early adopters. If we split, the sooner the better

ffromani:
# Naming

the individual mode driver should at least initially retain the current name.

[...]

PR #322, opened 2026-09-18
docs: config: deprecate individual mode

ffromani:
#### What type of PR is this?

/kind documentation

Now that #229 and #316 landed with good early feedback, we can deprecate individual mode at least for the next release (0.4.0) and, if everything goes as planned, remove it in the next-next release (0.5.0).
This effectively replaces https://github.com/kubernetes-sigs/dra-driver-cpu/issues/152 and obsoletes #112

Related to 152 and 112

While I think this is the next logical step after merging #229 and #316 , let's agree and acknowledge the change of course wrt 152 (we now need to do the opposite) and #112 (which will become counterproductive at this stage - at least for the original reasons)

I used AI for the first draft but I reviewed all the hunks and rewrote a big part of them

#### Release note

```release-note
The individual mode is now deprecated; deployments which require precise CPU placement should use the external allocator support instead.
```

#### Documentation updates

All the docs


ffromani:
/approve cancel

because of the implications on #152, #112 and the roadmap in general, this require a wide review and consensus

docs/user/device-attributes.md
```diff
@@ -33,7 +34,7 @@ CPU group, `individual` one device per CPU.
 
 #### External allocator attributes
 
-With the external allocator in grouped mode, the driver also publishes:
+With the external allocator in grouped mode, the driver also publishes these attributes for NUMA-node, socket, and machine groups.
```
pravk03: nit: "additional attributes" ?

docs/user/configuration.md
```diff
@@ -67,20 +66,28 @@ on - is configured through other Helm values, not through this file.
...
 - `grouped`: exposes a single device representing a group of CPUs. This mode treats CPUs
   as a [consumable capacity](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5075-dra-consumable-capacity/README.md)
   within the group, improving scalability by reducing the number of API objects.
 
 `groupBy` (string, default: `numanode`)
 
 - Grouping strategy used when `cpuDeviceMode` is `grouped`.
-- `numanode`: groups CPUs by NUMA node.
-- `socket`: groups CPUs by socket.
+- `numanode`: groups CPUs by NUMA node. Supported with an external allocator (see: `machine`).
```
pravk03, line 79: nit: This reads as if `numanode` is only supported with the `external` allocator ?

We can probably just drop that extra sentence here, since `machine` below already explicitly notes that it requires the external allocator

README.md
```diff
@@ -170,7 +170,7 @@ when filing an issue — it collects the CPU topology and driver configuration n
 - [Feature Support](docs/user/feature-support.md) - supported/unsupported features.
 - [Matching Kubelet CPU Manager Options](docs/user/feature-support.md#matching-cpu-manager-options) - kubelet cpumanager policy options and their driver equivalents.
 - [Workload Configuration Requirements](docs/user/workload-requirements.md) - how to set pod/container CPU requests alongside DRA claims.
-- [Custom Opaque CPUSet Allocation Overrides](docs/user/opaque-cpuset-overrides.md) - explicit core assignment for `groupBy: machine` mode.
+- [Custom Opaque CPUSet Allocation Overrides](docs/user/opaque-cpuset-overrides.md) - implement explicit core assignment for `groupBy: machine` mode; recommended upgrade path for users of `individual` mode.
```
pravk03, line 173: This should be updated ?

pravk03 (commented):
Thanks again!. 
A couple of minot nits, but looks great overall. 
/lgtm

docs/user/configuration.md
ffromani: good points. Let me check.

README.md
ffromani: indeed. Done!

AutuSnow:
Thanks！！
/lgtm

Issue #320, opened 2026-09-16
Use upstream deviceattribute helpers and support the list form of the standard numaNode attribute


alanhuangch:
## Summary

The driver publishes the scalar `resource.kubernetes.io/numaNode` attribute (added in #219) by building the value by hand, and only in the scalar form. Now that the repo is rebased on Kubernetes 1.37, `k8s.io/dynamic-resource-allocation v0.37.0` provides `deviceattribute.GetNUMANodeAttribute(numaNode, form, ...)` together with the list form.

#219 explicitly anticipated this follow-up ("A later PR can switch to upstream `deviceattribute` constants/helpers once available in this repo's Kubernetes dependency version"), so filing this to track and scope it.

/kind feature

## Motivation

- Stop hand-rolling the standard attribute and let the upstream helper own the validation (negative NUMA nodes must not be published).
- Publish the list form (`[physical NUMA node, same-socket NUMA nodes at minimum ACPI SLIT distance]`) so that `matchAttribute` can pair CPUs with devices that are equally local, instead of requiring an exact node match.
- Reuse the existing sysfs abstraction: the helper accepts `deviceattribute.WithFS(fs.ReadLinkFS)`, which `pkg/sysfs.SysFS` already embeds, so the existing test overlays keep working.

## Proposed change

1. Use `deviceattribute.GetNUMANodeAttribute` in `pkg/device/builder.go` for individual CPU devices (and for `group-by=numanode` if grouped mode is still supported in the release this lands in).
2. Add a configuration switch that mirrors the cluster's `DRAListTypeAttributes` state, following the outcome of #302; until #302 is decided, an option in the style of the existing `--expose-pcie-roots` would be acceptable.
3. Make `Config.DevicesPerResourceSlice()` account for list-type attributes regardless of which switch enabled them (the limit drops from 128 to 64).
4. Treat a dropped list value as a fatal publish error rather than adapting silently: when the gate is off the apiserver drops the value, and the `resourceslice` controller already reports this as a `DroppedFieldsError` naming `DRAListTypeAttributes`.

## Semantics and migration notes

- With `DRAListTypeAttributes` enabled the allocator compares attributes as sets and converts scalars to single-element sets, so a CPU publishing `[0,1]` still matches a peer publishing the scalar `0`; enabling this on the CPU side alone does not break existing pairings.
[...]

PR #323, opened 2026-09-19
device: use the upstream deviceattribute helper for numaNode

alanhuangch:
Part of #320. Switch the internal machinery to the kube libs and keep publishing a scalar value.

## What changes

The standard resource.kubernetes.io/numaNode attribute was built by hand. Both places that publish it (individual CPUs and group-by=numanode) now take the value from deviceattribute.GetNUMANodeAttribute(id, ScalarAttribute).

The attribute keeps the same name, the same scalar int type and the same value. The user docs already describe it as an int, so they are unchanged.

## Validation now comes from upstream

The helper refuses a NUMA node outside the valid range rather than publishing it. That path is unreachable in production: topology discovery drops CPUs with incomplete topology before devices are built, so a build reaching it now fails loudly instead of publishing a bogus value.

## Tests

TestBuildPublishesScalarNUMANodeAttribute pins the published name, scalar type and value for the individual and numanode layouts. It is a characterization test: it also passes on the pre-change code, which is the point.

TestBuildRejectsNegativeNUMANodeAttribute (linux-only) covers the new validation path. It was compile-checked locally (GOOS=linux go test -c) but not executed, because no container runtime was available on my machine; CI will run it.

## Verification

gofmt and golangci-lint clean; GOOS=linux go build ./... and go vet ./... pass; go test ./pkg/... passes. Two failures on darwin/arm64 are pre-existing and reproduce on clean main: TestBuildValidatesDeviceAttributeValueCount needs an amd64 PCIe scanner, and internal/gatherinfo hangs.


pkg/device/builder.go
```diff
@@ -360,13 +360,15 @@ func createGroupedCPUDeviceSlices(layout Layout, deviceInfos []groupedCPUDeviceI
...
+				// Driver-specific/non-standard attributes
 				AttributeSocketID:   {IntValue: new(int64(deviceInfo.socketID))},
 				AttributeSMTEnabled: {BoolValue: new(topo.SMTEnabled)},
 				AttributeNumCPUs:    {IntValue: new(availableCPUs)},
 			}
+			// DRA standard attributes, published through the upstream helpers
+			if err := addNUMANodeAttribute(deviceAttrs, deviceInfo.numaNodeID); err != nil {
+				return nil, err
+			}
 			addCompatibilityAttributes(deviceAttrs, int64(deviceInfo.numaNodeID))
```
ffromani: nit: if you resubmit, please let's make `addCompatibilityAttributes` the first function to be called after we create `deviceAttrs`:
```
deviceAttrs :=map[resourceapi.QualifiedName]resourceapi.DeviceAttribute{
    ...
}
addCompatibilityAttributes(deviceAttrs, int64(deviceInfo.numaNodeID))
if err := addNUMANodeAttribute(deviceAttrs, deviceInfo.numaNodeID); err != nil {
    return nil, err
}
...
```

pkg/device/builder.go
```diff
@@ -436,6 +436,10 @@ func createCPUDeviceSlices(deviceInfos []cpuDeviceInfo, pcieRootMapper *store.PC
 			AttributeCPUID:      {IntValue: new(int64(cpu.CpuID))},
 		}
 
+		// DRA standard attributes, published through the upstream helpers
+		if err := addNUMANodeAttribute(deviceAttrs, cpu.NUMANodeID); err != nil {
+			return nil, err
+		}
```
ffromani, line 441: ditto

pkg/device/builder.go
```diff
@@ -452,6 +456,19 @@ func createCPUDeviceSlices(deviceInfos []cpuDeviceInfo, pcieRootMapper *store.PC
 	return allDevices, nil
 }
 
+// addNUMANodeAttribute publishes the standard "numaNode" device attribute, in
+// scalar form, for a device bound to a single NUMA node. The value is built by
+// the upstream deviceattribute helpers (see numaattribute_linux.go), so the
+// attribute name, its type and the validation are owned by the DRA library.
+func addNUMANodeAttribute(attrs map[resourceapi.QualifiedName]resourceapi.DeviceAttribute, numaNodeID int) error {
+	value, err := numaNodeAttributeValue(numaNodeID)
```
ffromani: can we remove this intermediate layer? I'd just implement `addNUMANodeAttribute` in `builder_linux.go` and add a one-line stub in `builder_other.go`

ffromani (commented):
thanks, looks pretty good. Let's simplify a bit and we can merge

pkg/device/builder.go
alanhuangch: Done in f8dc94c2c84c487fde689bf7567b7004c8978656

pkg/device/builder.go
alanhuangch: Done in f8dc94c2c84c487fde689bf7567b7004c8978656

pkg/device/builder.go
alanhuangch: Done in f8dc94c2c84c487fde689bf7567b7004c8978656

alanhuangch:
@ffromani  Thanks! I pushed the requested simplification in f8dc94c, including the platform-specific helper split and the attribute ordering adjustment. The Linux build and targeted tests pass. Ready for another look.

pkg/device/builder_other.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	resourceapi "k8s.io/api/resource/v1"
+	"k8s.io/dynamic-resource-allocation/deviceattribute"
+)
+
+func addNUMANodeAttribute(attrs map[resourceapi.QualifiedName]resourceapi.DeviceAttribute, numaNodeID int) error {
+	attrs[deviceattribute.StandardDeviceAttributeNUMANode] = resourceapi.DeviceAttribute{IntValue: new(int64(numaNodeID))}
+	return nil
+}
```
ffromani, line 29: thanks, this is is what the old code did. I wonder if in this path we should just not report anything.
But is not too big of a deal, our code kinda assumes we're running on linux anyway.
This assumption is half enforced and half assumed inconsistentyl anyway, so better to do a sweeping pass later (0.4.0 meterial).

ffromani (commented):
/approve
/lgtm

thanks!

ffromani:
@alanhuangch Thanks, the two commits are fine. For future PRs, once review is complete and before final approval, feel free to squash a small follow-up commit that addresses review feedback into the commit it amends. I generally prefer the cleaner final history whenever possible.

alanhuangch:
@ffromani  Thanks! Got it, will do that in future PRs.

PR #325, opened 2026-09-20
fix(helm): register under the node object name via downward API

matchseller:
#### What type of PR is this?

/kind bug

The helm chart deploys the driver without any hostnameOverride, so the driver registers under the kernel hostname. On hosts where the kernel hostname differs from the Node object name (commonly when the machine hostname is an IP address), the ResourceSlice controller cannot retrieve the node object used as slice owner and ResourceSlice publishing fails repeatedly:

    processing ResourceSlice objects err="retrieve node "10-0-130-70": nodes "10-0-130-70" not found" driver="dra.cpu"

This PR injects `NODE_NAME` via the downward API (`spec.nodeName`) and passes `--hostname-override=$(NODE_NAME)` whenever `hostnameOverride` is unset in both `args` and `driverConfig`. Each node now registers under its Node object name with no per-node configuration.

Explicitly configured values keep the precedence: when `hostnameOverride` is set in `args` or `driverConfig`, the flag is not
passed, since a flag would shadow the config file.

N/A

- Helm-chart-only change: `templates/daemonset.yaml` plus the generated docs (`values.yaml` comment, `values.schema.json`, `README.md`).No Go code or default behavior of the driver binary changes.
- Behavior change only for users who set neither `args.hostnameOverride` nor `driverConfig.hostnameOverride`: those deployments now pass the new flag. For all such users `spec.nodeName` matches the Node object name by definition, so this should be strictly a fix; deployments relying on the kernel hostname being identical to the Node object name are unaffected in practice.
- The env var expansion `$(NODE_NAME)` is processed by the kubelet, so no image or entrypoint changes are needed.
- Docs were hand-edited to match `values.yaml`; happy to rerun helm-docs if the maintainers prefer generated output committed as-is.

#### Release note

```release-note
Fix the helm chart so the driver registers under the Node object name(spec.nodeName via the downward API) instead of the kernel hostname,when no explicit hostnameOverride is configured.
```

#### Documentation updates

- deployment/helm/dra-driver-cpu/README.md (args.hostnameOverride description)
- deployment/helm/dra-driver-cpu/values.yaml (helm-docs comment)
- deployment/helm/dra-driver-cpu/values.schema.json (matching description)


ffromani (commented):
/lgtm

thanks for your contribution! the issue you identified is legit and the fix LGTM.

in #282 we introduced more validation for helm-rendered charts.
To add coverage, we can add validation in `test/render/validator_test.go` and testcases in `hack/ci/helm-render-check.sh`

It would be great to add some coverage, any chance you can try this? we would probably need to extend the `accepts` and `render` shell function and to add some golang checks like `checkNodeNameOverride` or something.

If you have bandwidth, I’d prefer to add this coverage before merging; otherwise, we can track it as a follow-up. Please let us know which you prefer.

matchseller:
> /lgtm
> 
> thanks for your contribution! the issue you identified is legit and the fix LGTM.
> 
> in #282 we introduced more validation for helm-rendered charts. To add coverage, we can add validation in `test/render/validator_test.go` and testcases in `hack/ci/helm-render-check.sh`
> 
> It would be great to add some coverage, any chance you can try this? we would probably need to extend the `accepts` and `render` shell function and to add some golang checks like `checkNodeNameOverride` or something.
> 
> If you have bandwidth, I’d prefer to add this coverage before merging; otherwise, we can track it as a follow-up. Please let us know which you prefer.

Done — added the coverage you suggested:

- test/render/validator_test.go: a new checkNodeNameOverride verifies the
  fallback flag (`--hostname-override=$(NODE_NAME)`) appears exactly once
  when nothing is configured, and never appears at an explicit
  configuration (a flag would shadow the config file); a companion check
  pins the NODE_NAME env to its downward API source (spec.nodeName).
- hack/ci/helm-render-check.sh: every existing case now expects the
  fallback, and two new cases cover the explicit overrides via
  args.hostnameOverride and driverConfig.hostnameOverride, including the
  fold into the rendered config.yaml.

`make helm-render-check` is green (11 accepts + 5 refuses).

test/render/validator_test.go
```diff
@@ -243,3 +250,58 @@ func checkFlag(c *corev1.Container, root string, want bool) error {
...
+		return fmt.Errorf("%d env vars named %q, want one", len(found), nodeNameEnv)
+	}
+	e := found[0]
+	if e.ValueFrom == nil || e.ValueFrom.FieldRef == nil {
+		return fmt.Errorf("env %q is not a fieldRef", nodeNameEnv)
+	}
+	if want := "spec.nodeName"; e.ValueFrom.FieldRef.FieldPath != want {
+		return fmt.Errorf("env %q comes from %q, want %q", nodeNameEnv, e.ValueFrom.FieldRef.FieldPath, want)
+	}
+	return nil
```
ffromani: this works, but I'd prefer to make `spec.nodeName` a module-level constant. It's one of the things we care most in the validation after all.

test/render/validator_test.go
```diff
@@ -243,3 +250,58 @@ func checkFlag(c *corev1.Container, root string, want bool) error {
...
+	if want == "" {
+		if len(found) != 0 {
+			return fmt.Errorf("the chart passes %v, want no hostname-override at an explicit config", found)
+		}
+		return nil
+	}
+	if len(found) != 1 {
+		return fmt.Errorf("%d hostname-override arguments, want one", len(found))
+	}
+	if got := found[0]; got != "--hostname-override="+want {
```
ffromani: `--hostname-override` should probably be a top-level constant

hack/ci/helm-render-check.sh
```diff
@@ -70,20 +77,51 @@ refuses() {
...
 # Unset: the chart leaves the root to the driver, so there is no flag to check.
-accepts default '' "${default_root}" false
-accepts relocated /var/lib/custom-kubelet /var/lib/custom-kubelet true
+# With no hostnameOverride configured either, every case below falls back to the
+# Node object name expanded from the downward API.
+accepts default '' "${default_root}" false "${nodename_expandable}"
+accepts relocated /var/lib/custom-kubelet /var/lib/custom-kubelet true "${nodename_expandable}"
 # Cleaned by the template, so the mounts and the flag cannot differ by a slash.
-accepts uncleaned /var/lib/custom-kubelet/ /var/lib/custom-kubelet true
+accepts uncleaned /var/lib/custom-kubelet/ /var/lib/custom-kubelet true "${nodename_expandable}"
```
ffromani: I *think* we can change `accepts` in such a way that we don't require to add this argument?
meaning, what we should try is to make the extra tests you are adding additive: the old relocated-root directory tests should be unchanged, and then we can add the ones for hostname override.
I *think* this should be doable but I didn't spent too much effort on this yet. I'd like to avoid to waste effort so what's your assessment? if you think we can pivot cheaply enough let's please do it, otherwise let's re-evaluate if we can keep the current shape or if there is a low-effort alternative improvement.

ffromani (commented):
thanks for quickly adding the tests, appreciated! a couple of cleanups and a question inside, then I think we're good to go: the main (likely opnly) remaining blocker I see is the inline question

matchseller:
> thanks for quickly adding the tests, appreciated! a couple of cleanups and a question inside, then I think we're good to go: the main (likely opnly) remaining blocker I see is the inline question

Hi,Both points are addressed:

- `spec.nodeName` and `--hostname-override` are now module-level
  constants.
- The old relocated-root cases are untouched: the fallback
  expectation is pinned once inside `accepts`, and the validator
  checks the fallback flag and the NODE_NAME downward API source.

ffromani:
Thanks @matchseller we are almost there. I'd like a bit more isolation, as follows.
First, we can add a conditional to the `validator_test.go` like this:
```
 var checkHostnameOverride = flag.Bool("check-hostname-override", false, "whether to validate the hostname-override contract")
```
and use it later like this
```
if *checkHostnameOverride {
       if err := checkNodeNameOverride(ds, *expectedOverride); err != nil {
               t.Fatal(err)
       }
}
```
now we can fully isolate the neww checks in `hack/ci/helm-render-check.sh`, something like this:
```
checkHostnameOverride() {
       local name=$1 want_config=$2
       shift 2

       echo "hostname override render check: ${name}"
       "${helm}" "${template_args[@]}" "$@" \
               >"${tmp}/${name}-daemonset.yaml"
       local test_args=(
               -manifest "${tmp}/${name}-daemonset.yaml"
               -expected-root "${default_root}"
               -expect-root-flag=false
               -check-hostname-override
       )
       if [[ -n ${want_config} ]]; then
               test_args+=(-expected-override='')
       fi
       go test -count=1 ./test/render/ -args "${test_args[@]}"

       if [[ -n ${want_config} ]]; then
               "${helm}" template test "${chart}" --kube-version "${kube_version}" \
                       --show-only templates/configmap.yaml "$@" \
                       >"${tmp}/${name}-configmap.yaml"
               if ! grep -Fq "hostnameOverride: ${want_config}" "${tmp}/${name}-configmap.yaml"; then
                       echo "the rendered config does not contain hostnameOverride: ${want_config}" >&2
                       return 1
               fi
       fi
}

checkHostnameOverride fallback ''
checkHostnameOverride args-override mynode --set-string 'args.hostnameOverride=mynode'
checkHostnameOverride config-override mynode --set-string 'driverConfig.hostnameOverride=mynode'
```
(note this intentionally duplicates and then changes the existing `accepts`)

finally we can now revert _all_ the modifications to the `accepts`, making the extra tests fully additive, which is the real goal and my last concern.

Once all the test additions are purely additive I think we can merge

matchseller:
> Thanks @matchseller we are almost there. I'd like a bit more isolation, as follows. First, we can add a conditional to the `validator_test.go` like this:
> 
> ```
>  var checkHostnameOverride = flag.Bool("check-hostname-override", false, "whether to validate the hostname-override contract")
> ```
> 
> and use it later like this
> 
> ```
> if *checkHostnameOverride {
>        if err := checkNodeNameOverride(ds, *expectedOverride); err != nil {
>                t.Fatal(err)
>        }
> }
> ```
> 
> now we can fully isolate the neww checks in `hack/ci/helm-render-check.sh`, something like this:
> 
> ```
> checkHostnameOverride() {
>        local name=$1 want_config=$2
>        shift 2
> 
>        echo "hostname override render check: ${name}"
>        "${helm}" "${template_args[@]}" "$@" \
>                >"${tmp}/${name}-daemonset.yaml"
>        local test_args=(
>                -manifest "${tmp}/${name}-daemonset.yaml"
>                -expected-root "${default_root}"
>                -expect-root-flag=false
>                -check-hostname-override
>        )
>        if [[ -n ${want_config} ]]; then
>                test_args+=(-expected-override='')
>        fi
>        go test -count=1 ./test/render/ -args "${test_args[@]}"
> 
>        if [[ -n ${want_config} ]]; then
>                "${helm}" template test "${chart}" --kube-version "${kube_version}" \
>                        --show-only templates/configmap.yaml "$@" \
>                        >"${tmp}/${name}-configmap.yaml"
>                if ! grep -Fq "hostnameOverride: ${want_config}" "${tmp}/${name}-configmap.yaml"; then
>                        echo "the rendered config does not contain hostnameOverride: ${want_config}" >&2
>                        return 1
>                fi
>        fi
> }
> 
> checkHostnameOverride fallback ''
> checkHostnameOverride args-override mynode --set-string 'args.hostnameOverride=mynode'
> checkHostnameOverride config-override mynode --set-string 'driverConfig.hostnameOverride=mynode'
> ```
> 
> (note this intentionally duplicates and then changes the existing `accepts`)
> 
> finally we can now revert _all_ the modifications to the `accepts`, making the extra tests fully additive, which is the real goal and my last concern.
> 
> Once all the test additions are purely additive I think we can merge

Done — the checks now follow your sketch exactly:

- validator: the node name checks are gated behind
  -check-hostname-override, defaults untouched when unset
- harness: standalone checkHostnameOverride(name, wantConfig, ...)
  runs the three cases (fallback, args-override, config-override),
  pinning both the flag absence and the config.yaml fold for the
  explicit cases;

ffromani:
/approve
/lgtm

thanks. As followup, we would need to avoid the deprecate `--hostname-override` flag. Perhaps we just need to revisit the priorities for node name resolution. I'd prefer to track as followup.

